PR #41, opened 2013-09-20
Command line option to show VRRP status

jonasj76:
This commit adds a new command line option, -s, --vrrp-status, which will show the status of VRRP instances.

Example output:

<pre>
  VRRP Instance             : VI_1
     Interface              : eth0
     Virtual Router ID      : 1
     State                  : BACKUP
     Virtual IP address     : 192.168.2.99
     Advertisement interval : 1.00 sec
     Preemption             : Disabled
     Multicast group        : 192.168.2.2
     Priority               : 100
     Effective Priority     : 100
     Master router          : 192.168.2.1 priority 150
     Master down interval   : 3.0
</pre>

### DIFF
--- a/keepalived/check/check_daemon.c
+++ b/keepalived/check/check_daemon.c
@@ -162,6 +162,15 @@ sigend_check(void *v, int sig)
 		thread_add_terminate_event(master);
 }
 
+/* Dump handler */
+void
+sigdump_check(void *v, int sig)
+{
+	log_message(LOG_INFO, "Dumping Healthchecker config on signal");
+	dump_global_data(global_data);
+	dump_check_data(check_data);
+}
+
 /* CHECK Child signal handling */
 void
 check_signal_init(void)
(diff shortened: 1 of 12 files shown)


hangingman:
So far, this feature is not implemented.
I'm from this thread, [keepalived vrrp status check from cmdline](https://keepalived-devel.narkive.com/E8uJ9HiN/keepalived-vrrp-status-check-from-cmdline)

pqarmitage:
The majority of this has been implemented. Sending SIGUSR1 to the keepalived parent process will trigger the writing of `/tmp/keepalived_parent.data`, `/tmp/keepalived.data` (VRRP), `/tmp/keepalived_check.data` and `/tmp/keepalived_bfd.data` (the configure option --with-tmp-dir can change the location to be other than /tmp, e.g. /var/lib/misc).

A script to trigger this could be along the lines of:
```
#!/bin/bash

kill -USR1 $(cat /run/keepalived.pid)
```
so I am not sure that it needs a command line option for keepalived to be able to trigger it.

There are now also JSON and B-Bus options as well.

hangingman:
@pqarmitage I don't understand design concept that why I need to send signal to keepalived just I want to know VRRP status.
But, this is because maybe I'm layman. Anyway thank you !

PR #42, opened 2013-09-25
Add VRRPv3 support (RFC 5798)

jonasj76:
Adds a VRRP version configuration option:
<pre>version &lt;VERSION&gt;</pre>
where version can be "2" or "3".
Using version 2, which is the default, will send packets in VRRPv2 format (RFC 3768), i.e. unchanged behavior from previously. By using version 3, the packets will be send in VRRPv3 format instead (RFC 5798).

Adds support for centiseconds advertisement interval for VRRPv3. To configure an interval in centiseconds, a value between 0.01 and 40.95 can be used.
<pre>advert_int &lt;FLOAT&gt;</pre>
Note that an interval in centiseconds is only available for VRRPv3.

### DIFF
--- a/keepalived/include/vrrp.h
+++ b/keepalived/include/vrrp.h
@@ -38,23 +38,41 @@
 #include "vector.h"
 #include "list.h"
 
-typedef struct _vrrphdr {			/* rfc2338.5.1 */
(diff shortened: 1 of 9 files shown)


acassen:
patch set has been merged ! thanks for your job jonas !

PR #111, opened 2014-12-03
add arp_sleep to throttle gratuitous arp

dupondje:
Hi,

I created this small patch to resolve an issue we are having with keepalived.
On servers with alot of virtual IP's, we have issues that our switches throttle the gratuitous arp packages.

This can't be changed on switch level (hardcoded), so we had to find some 'workaround'/'fix' for this.
So I implemented the following patch.

The patch makes you able to configure an 'arp_sleep' variable.
This will cause a small usleep between the arp's, which causes to resolve the throttling.

If you don't want the feature, you can just not set it (default = 0), and it won't affect anything.

The change also contains a seconds patch which is a rewrite of the code to a threaded version based on the comments in https://github.com/acassen/keepalived/pull/108.

Thanks!
Jean-Louis

### DIFF
--- a/doc/keepalived.conf.SYNOPSIS
+++ b/doc/keepalived.conf.SYNOPSIS
@@ -32,6 +32,7 @@ global_defs {				# Block identification
     smtp_connect_timeout <INTEGER>	   # Number of seconds timeout connect
  					   #  remote SMTP server
     router_id <STRING>			   # String identifying router
+    arp_sleep <INTEGER>                    # Sleep between Gratuitous ARP (in ms)
     vrrp_mcast_group4 <IPv4 ADDRESS>	   # optional, default 224.0.0.18
     vrrp_mcast_group6 <IPv6 ADDRESS>	   # optional, default ff02::12
 }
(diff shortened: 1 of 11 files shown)


dupondje:
Any update on this? Could this be added? Or are there more changes needed?

dupondje:
Any reason why this hasn't been merged yet? Are you waiting for something specific?

pqarmitage:
Jean-Louis,

Many apologies for the long time taken to respond to your pull request. You should see that there is now a further pull request that extends your idea and makes if more generic, so I hope that that will meet your needs.

The configuration is slightly different from what you proposed, in order to make it more generic. In order to configure to to work as your last version, adding the following before any vrrp instances in the config file will achieve the same result as you had before:
garp_group {
    garp_interval 0.1         # This time is specified in seconds, with a granularity of 0.000001s
    interfaces {
        eth0
        eth1
    }
}
where all the interfaces over which keepalived is running are specified in the interfaces secion (for any VMACs, the interface to specify is the underlying interface, and not the VMAC interface).

PR #135, opened 2015-03-12
This patch is a demonstration of how to add counters to the keepalived u...

akshat-sri:
...sing the "ZeroMQ" IPC.

Types of counters:
1. Invalid packets			-No. of VRRP advertisements received with invalid type.
2. Too short				-No. of VRRP advertisements received with packet length less than length of VRRP header.
3. Transitions to Master		-No. of times the virtual router transits to "Master" state.
4. Total received			-No. of valid VRRP advertisements received by the virtual router in "Backup" state
5. Bad TTL				-No. of VRRP advertisements received with invalid "time to live" field.
6. Failed authentication		-No. of VRRP advertisements received which failed in authentication check due to invalid password.
7. Unknown authentication		-No. of VRRP advertisements received with unknown authentication type (invalid).
8. Conflicting authentication		-No. of VRRP advertisements received with conflicting authentication type (mismatch).
9. Conflicting Advertise time		-No. of VRRP advertisements received with conflicting advertisement interval (mismatch).
10.Conflicting Addresses		-No. of VRRP advertisements received with conflicting secondary and primary virtual IP address.
11.Received with zero priority		-No. of VRRP advertisements received with zero priority.
12.Sent with zero priority		-No. of VRRP advertisements sent with zero priority.

Assumptions:
1. ZeroMQ is the IPC mechanism is been used in this patch, hence the zmq library path shall be provided in order to compile the code. This has been added as an optional feature in "configure.in" file.
2. VRRP counter is an optional feature which has been placed under conditional compilation flag. User can enable/disable based on the requirement.
3. The ordering of VRRP counter in IPC data field is based on the sequence mentioned in "Types of counters" section. The counters shall be send/receive in the string format.
4. The namespace file used to bind the socket shall be user's choice.
5. Currently we use the "ipc:///tmp/zmq_file" i.e. "ipc://" is the transport mechanism (zmq provided) and "/tmp/zmq_file" is any file created in the kernelspace. This has been done in order the two daemons on the same device can interact to each other.
6. The maximum no. of tokens that can be received by zmq is 259 (258+null character).
7. The magic figure of 258 is (255 (max vrrp instance) + 1 (vrrp count) + 1 (msg identifier) + 1 (all or specific inst.)).
8. Each IPC token can never exceed the length of 21 in character format as the largest data type of token is "unsigned long"
9. New thread (linux pthread type) has been implemented for the communication of IPC between daemons. This is for send/receive of the IPC data which consists of counters information.
10. Support for clearing the VRRP counters has also been provided along with the patch.
11. Support for virtual router details which is in "Master" state to other virtual routers which are in "Backup" & "Init" state.


acassen:
unfortunately this is completely broken now. I must close it.

PR #140, opened 2015-04-10
ICMPv6

RomainLaniece:
Hi,

I use Keepalived 1.2.16, with native_ipv6 flag.

This week, I noticed that when my master turned off, my backup became master almost instantly (;D), but my IPv6 virtual addresses were not "pingable" until about 30-45 seconds. 

I did it once again with wireshark, in order to check the ICMPv6 packets sent by Keepalived, and saw that the packet was not considered as ICMPv6, but was unknown : the "next header" field was not NEXTHDR_ICMP (58), but was an unknwon value. It means that nobody around was able to understand the notification sent from the new master.

After having debugged a little into the file keepalived/vrrp/vrrp_ndisc.c, I found out that my problem was while copying the source address into the IPv6 header. The copy of this address was overwriting the two previous bytes ("next header" and "hop limit").

To resolve my problem, I replaced the copy with memcpy (and did the same into the ICMPv6 checsum calculation), and the ICMPv6 packets are now recognized by all the neighbours.

I don't know if the problem exists for everybody, or if it's just because of my cross-compilation toolchain (?). By the way, if it can helps in any way, here is my patch.

Romain Laniece.

### DIFF
--- a/keepalived/vrrp/vrrp_ndisc.c
+++ b/keepalived/vrrp/vrrp_ndisc.c
@@ -85,8 +85,8 @@ ndisc_icmp6_cksum(const struct ip6hdr *ip6, const struct icmp6hdr *icp, uint32_t
 
 	/* pseudo-header */
 	memset(&phu, 0, sizeof(phu));
-	phu.ph.ph_src = ip6->saddr;
-	phu.ph.ph_dst = ip6->daddr;
+	memcpy(&phu.ph.ph_src, &ip6->saddr, sizeof(struct in6_addr));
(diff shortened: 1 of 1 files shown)


acassen:
Hi Romain,

maybe using struct ip6_hdr instead must be considered... memcpy sound overkill
will have a deeper look at this. Thanks for pointing. Your arch is ARM ? sound like memory alignment is in the corner...


regs,
Alexandre

RomainLaniece:
Hi Alexandre,

Yes, I'm working on ARM boards.

Regards,
Romain

PR #144, opened 2015-05-08
Upstreaming efforts from vyatta

coldoch:
A number of proposed keepalived changes from LATEST vyatta. Also including a number of other contributors including <Jonas Johansson>.


acassen:
Hi Colin,

thanks for the good job. The Accept-Mode part need to be rewritten since using execv to launch system iptables is just way to quick to address it...

Regs,
Alexandre

PR #146, opened 2015-05-09
snmp: restore use of net-snmp-config to build SNMP support

vincentbernat:
With a lazy linker, `libnetsnmpmibs` may require some additional
libraries to be linked (like `libsensors`). Therefore, only rely on
`net-snmp-config` to get the appropriate flags.
    
Also add some additional tests:
     - check that we can build a simple executable (NetSNMP can be quite
       broken and in this case, the error during compilation is not crystal
       clear, checking that in configure is more informative)
     - check if we subagent support is compiled in (This is optional and
       again, the error is not crystal clear during compilation).
     - check that net-snmp/agent/util_funcs.h is present (Due to a flaw in
       NetSNMP build process, this header was not installed for quite a long
       time, notably on RHEL derivatives; code to handle its absence was
       already present in Keepalived).

This code is derived from the one I am using in lldpd. It is quite "battle-tested" and I am able to compile SNMP support with distributions as old as RHEL 4.

Also, I didn't regenerate `configure`. Do you want me to?

### DIFF
--- a/configure.in
+++ b/configure.in
@@ -219,7 +219,7 @@ AC_SUBST(VRRP_VMAC)
 
 dnl ----[ Checks for SNMP support ]----
 SNMP_SUPPORT="_WITHOUT_SNMP_"
-if test "$enable_snmp" != "no"; then
+if test "$enable_snmp" = "yes"; then
 
   AC_PATH_TOOL([NETSNMP_CONFIG], [net-snmp-config], [no])
   if test x"$NETSNMP_CONFIG" = x"no"; then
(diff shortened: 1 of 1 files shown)


vincentbernat:
Please, wait a bit, I pushed too fast, I didn't test. ;-)

vincentbernat:
There is a conflict due to use of libperl by NetSNMP. I suppose this was why I did disable CFLAGS in the past... Let me look if I can find a proper fix.

acassen:
outch, didn't read your msg...

PR #148, opened 2015-05-14
Set global data default values after parsing config file

rohara:
This patch will defer setting the global data default values until
after the config file has been parsed. This will potentially avoid two
calls to getaddrinfo. For example, if the router_id and/or email_from
parameters are set in the config file, there is no need to call
getaddrinfo twice in order to set a default value. Instead, this patch
will check to see if they values are unset after parsing the config
file. Note that email_from and smtp_connection_to are only set to a
default value if they are unitialized and smtp_server is specified.

### DIFF
--- a/keepalived/check/check_daemon.c
+++ b/keepalived/check/check_daemon.c
@@ -113,6 +113,7 @@ start_check(void)
 		stop_check();
 		return;
 	}
+	init_global_data(global_data);
 
 	/* Post initializations */
 	log_message(LOG_INFO, "Configuration is using : %lu Bytes", mem_allocated);
--- a/keepalived/core/global_data.c
+++ b/keepalived/core/global_data.c
@@ -84,18 +84,6 @@ set_default_mcast_group(data_t * data)
 	inet_stosockaddr("ff02::12", 0, &data->vrrp_mcast_group6);
 }
 
-static void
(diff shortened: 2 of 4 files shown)


rohara:
This is needed for OpenStack Neutron, which is using keepalived for the HA L3 agent. In this case, keepalived is running in a network namespace that does not have access to DNS. As a result, the getaddrinfo calls will block waiting for a response which causes from some master/backup flapping when keepalived is started or restarted. The OpenStack community would like to avoid this, and I believe this will solve the problem.

beekhof:
Hi, just wondering if there are any plans to review this patch?
We're quite keen to get it accepted.

assafmuller:
I wrote a lot of the code for virtual router high availability in OpenStack Neutron. Ryan is right that router namespaces may not have DNS connectivity in some deployments and this is killing the viability of this feature in those cases. I actually had two different operators talk to me about this in the last OpenStack summit. The OpenStack community would really appreciate your help here.

acassen:
Well, merged this, but you MUST consider in such an isolation use-case tweaking system conf accordingly! this is the reason nsswitch.conf is here. If you dont have any resolv access then it seems a better way to use proper settings.

regs,
Alexandre

mangelajo:
Thank you for working on this & merging.

PR #150, opened 2015-05-21
VRRP incorrectly sets receive interface to vmac after reload

DStape:
When VRRP is reloaded after initially being configured with
the use_vmac option, VRRP sets its receiving interface pointer
to point to the vmac interface which is incorrect.

For transmitting, VRRP should be using the vmac interface that
results in frames being sent onto the wire with the vMAC address.
For receiving, VRRP should continue to use the base/real
interface.

This fix removes the line that incorrectly sets the receiving
interface pointer. This means VRRP will continue to use the
real interface for receiving traffic as the pointer was correctly
set before reload and this persists after reload.


DStape:
This fix only makes sense in the context of a patch that has not yet been submitted here. The patch will be pushed up soon.

PR #151, opened 2015-05-31
Add upstart job

CameronNemo:


acassen:
Upstart is up to sysadmin. I dont see any generic needs to make it installed by default

PR #153, opened 2015-06-11
Use /etc/iproute2/rt_tables for lookups to allow better readable table names in VRRP virtual routes

rolffokkens:
Currently keepalived only support numerical representations of virtual routes in config files. This patch (based on iproute2 code snippets) allows to specify character representations of tables by reading /etc/iproute2/rt_tables.

### DIFF
--- a/keepalived/include/vrrp_iproute.h
+++ b/keepalived/include/vrrp_iproute.h
@@ -67,5 +67,7 @@ extern void dump_iproute(void *);
 extern void alloc_route(list, vector_t *);
 extern void clear_diff_routes(list, list);
 extern void clear_diff_sroutes(void);
+extern void load_rt_tables(void);
+extern void free_rt_tables(void);
 
 #endif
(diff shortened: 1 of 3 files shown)


acassen:
What happens if your are changing /etc/iproute2/rt_tables and then perform a reload ? it will skip this update. I would also prefer not to keep struct in memory while not necessary. Please also consider using length controlled system call (strncpy, strncat, ...).

Regs,
Alexandre

rolffokkens:
Thanks for your feedback. Given your genuine interrest, I'll improve the change to not only fulfil my need but to create broader support.

rolffokkens:
@acassen, I processed your feedback.

huihongxiao:
From my perspective, the functionality here is very needed. We use keepalived to maintain virtual routes. We want to keep route table name to be consistent across hosts, so that we can use some existing data model to map route table. But now keepalived only support route table ID in virtual route. We have to do extra effort to maintain a consistent route table ID in existing data model.

pqarmitage:
Pull request #281, commit bf0a6a7 updates the patches in this pull request to include support for table names in iprules, allow table numbers greater than 255, to only load the rt_tables file if there are named tables in the keepalived configuration, and to discard the table names after the keepalived configuration file has been read, with them being reread if needed if the configuration is reload.

PR #154, opened 2015-06-12
Re-enable VRRPv2 checksum on inbound pkts

andempsey:
In the VRRPv3 code enhancement the ability to check VRRPv2 checksums
was mistakenly removed. This commit adds the functionality back into the
code base.


andempsey:
Cheers Alexandre

PR #155, opened 2015-06-15
Fix vrrp removes incorrect IPv4 address when VIPs are removed

DStape:
When vrrp has an IPv4 VIP that matches the primary interface
address, when the VIP is removed from the interface, the original
address ends up getting removed instead of the VIP.

The kernel receives a netlink message instructing it to remove
address x from a particular interface. For IPv4, address x can
be configured multiple times providing the prefix lengths differ.
If the IFA_ADDRESS attribute is not specified in a RTM_DELADDR
message, the kernel will delete the first address it finds a
match on, prefix length is not taken into account.

This fix therefore adds the IFA_ADDRESS attribute when deleting
IPv4 addresses so that the address removed is actually the VIP.

### DIFF
--- a/keepalived/vrrp/vrrp_ipaddress.c
+++ b/keepalived/vrrp/vrrp_ipaddress.c
@@ -86,6 +86,9 @@ netlink_ipaddress(ip_address_t *ipaddress, int cmd)
 	} else {
 		addattr_l(&req.n, sizeof(req), IFA_LOCAL,
 			  &ipaddress->u.sin.sin_addr, sizeof(ipaddress->u.sin.sin_addr));
+		if (cmd == IPADDRESS_DEL)
+			addattr_l(&req.n, sizeof(req), IFA_ADDRESS,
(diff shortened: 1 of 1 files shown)


DStape:
Repro steps to produce this bug:

- Add an IPv4 address to interface vrrp runs over, e.g. sudo ip addr add 192.168.122.7/24 dev eth0
- Make sure the IPv4 address you've added gets added to keepalived.conf as one of the virtual_ipaddress members
- Start or reload keepalived
- Verify that the VIP has indeed been added to the interface with a /32 (assuming the instance is the MASTER): ip addr
- Stop keepalived: sudo service keepalived stop (OR you can cause the vrrp instance to transition to BACKUP state since this will also remove the VIPs)
- Look at the output of "ip addr" - the /32 will remain (the VIP) but the original /24 will have been removed.

DStape:
Thanks Alexandre.

PR #184, opened 2015-08-20
make 'smtp_server' config to support domain name

flygoast:
Added support for domain name in config `smtp_server`, such as `smtp_server mail.example.com`


flygoast:
Changed it to use `getaddrinfo()`.

flygoast:
Foggoted it. Added.

PR #201, opened 2015-12-04
Dynamic addition of interfaces from netlink msg

andempsey:
When a tracked interface is deleted then recreated with the same config
VRRP groups tracking this interface will remain down. This is due to
tracking of stale information.

This patch listens for netlink messages for the creation of interfaces
and does one of two things.
i) If the interface doesn't exist in the vrrp interface list a new
interface structure is created and the information from the message is
used to fill the structure. This new interface is then added to the
interface queue.

ii) If the interface already exists in the queue we zero it and then
use the information in the message to fill the structure.

Signed-off-by: Anthony Dempsey <adempsey@brocade.com>


andempsey:
Looks like I've messed up the history trying to keep in track with the other pull requests. Sorry.

PR #202, opened 2015-12-04
check: segfault when there is no real server for a virtual server.

ppenzo:
It may happen that a VS is configured without any RS causing the check thread to segfault when accessing the RS data.
This patches fixes this issue.


ppenzo:
Return code fixed.
Sorry.

PR #217, opened 2016-01-07
add advert_count option

xiezhenye:
add advert_count option to configure how many vrrp advertisement will cause role switch instead of hard coded 3

### DIFF
--- a/doc/man/man5/keepalived.conf.5
+++ b/doc/man/man5/keepalived.conf.5
@@ -229,6 +229,7 @@ which will transition together on any state change.
 
     # VRRP Advert interval, secs (use default)
     advert_int 1
+    advert_count 3
     authentication {     # Authentication block
         # PASS||AH
         # PASS - Simple password (suggested)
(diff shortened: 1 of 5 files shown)


DStape:
I agree that the 3 does seem arbitrary, however this is the value stated in the RFC (https://tools.ietf.org/html/rfc5798) for the master down interval:

Advertisement interval contained in ADVERTISEMENTS received from the Master (centiseconds).  This value is saved by virtual routers in the Backup state and used to compute Skew_Time and Master_Down_Interval.  The initial value is the same as Advertisement_Interval.

xiezhenye:
3 can be the default value to match the RFC, but is too sensitive for some cases.

acassen:
please modify pull request to make it default to 3 to match RFC.

DStape:
In the cases that it is too sensitive then you should bump up the vrrp advertisement interval. Based on the RFC, I don't feel that this value should be configurable. However, 3 multiplies the advertisement interval anyway, so there is already a means to adjust this value.

Issue #14, opened 2012-11-01
Is it possible that no GARP was sent from a master that received a lower prio advert?


ashak:
Hi,

I've spent most of the morning trying to debug an issue with our keepalived setup and the only conclusion that i've been able to come up with is the following (unfortunately there's mention of something that sounds just like this in the changelog from 2002... so hrm!):

We have two instances of keepalived running on virtual machines.

This morning at 09:39:29 the slave instance decided that it was going to take over as master (fair enough, that's what it's supposed to do if it thinks the other end is dead right?).

The slave logged a whole bunch of Transition to MASTER STATE entries, then began logging Entering MASTER STATE. There's 20 VRRP Instances and just before the last VRRP instance entered master state, it logged that this instance had received a higher prio advert (This is at the point where the existing master wondered what on earth was going on and began logging: Received lower prio advert, forcing new election). At 09:39:30 all of the instances on the slave returned to BACKUP state, the master logged nothing more as presumably all its instances stayed in MASTER state.

At this point everything should have been fine, but it was not. Our alerting started complaining that almost everything was broken (this specific VMs are routers between a bunch of networks).

Some time later (and some tcpdumping later), I worked out that the reason why nothing was working was that some packets were travelling via the master and others were attempting to travel via the slave (and therefore being rejected).

An example is, I could see a SYN from a request to one of our services traverse the master and hit our load balancers. The SYN being sent on from the load balancers (we also run keepalived and use ipvs on our load balancers) came back through the router for the packet to hit the real server. The SYN ACK from the real server was nowhere to be seen on the master, however it did appear on the slave (and therefore was rejected).

Now I can't prove this....

(As shortly afterwards I shutdown keepalived on the master (allowing failover to the slave to happen properly), confirmed everything worked when the slave was now the master, then started keepalived on the original master and let everything fail back. Confirming that now everything worked as expected again.)

But all I can think is that the machines (ie my real servers, not that this is important here), had the incorrect MAC address of their default gateway in their ARP cache. As the packets were hitting my slave machine, I assume that during the short period where the slave also had all of the vrrp instances in MASTER state, some of my machines had arped for their default gateways and received the MAC addresses from the slave. When the master forced an election causing the slave to relinquish control again, I have a feeling that no GARP was sent out by the master meaning that some of my machines had the wrong MAC address in their cache.

Like I said, I can't prove that as in restating keepalived (and therefore everything being fixed) I no longer had the ability to go see what was in various arp caches.

[...]

PR #235, opened 2016-01-27
No auth - request #14

pqarmitage:


acassen:
Wow.... Goog job Quentin ! thanks...

PR #257, opened 2016-02-25
Implement RFC2787 SNMP (VRRPv2)

pqarmitage:
The first 8 commits are a series of minor fixes to the existing SNMP code.

The 9th commit adds a full, read-only, implementation of RFC2787, SNMP for VRRPv2.

The remaining commits allow different combinations of SNMP support to be built, and allow the different components of SNMP to be enabled via the configuration file.


acassen:
Hi quentin,

line 232/parser.c

there is a sleep(2)

acassen:
would you prefer me to merge your current pull request and then you can change active in install_keyword_root and submit a new pull request ?

pqarmitage:
On Mon, 2016-02-29 at 04:25 -0800, Alexandre Cassen wrote:

> would you prefer me to merge your current pull request and then you
> can change active in install_keyword_root and submit a new pull
> request ?
> 

Alexandre,

Yes, I think it would be easiest if you merge the current request and
I'll submit another one shortly. I'll also remove the sleep; it appears
that I must have inadvertently left it in after some testing.

With regards,
Quentin

PR #263, opened 2016-02-29
sctp ad persistent engine support

vrit:


vrit:
I've started from scratch, this should work now... hopefully :)

PR #264, opened 2016-03-02
Add RFC6527 SNMP for VRRPv3 read-only and trap support

pqarmitage:
The final commit adds the RFC6527 SNMP support. The previous patches are tidying up, mainly of the SNMP code.

Commit 4b72e7a is perhaps the only commit of any significance, other than the final commit. It makes the VRRP instances with priority 255 behave as thought "state MASTER" had been specified in the config file. This is in accordance with the VRRP RFCs which state that a VRRP instance will transition directly to master if it is the address owner, i.e. priority is 255.


acassen:
I have a HUGE warning in my test tool in that vrrp_scheduler. I certainly dont fully remind the deep reason but if I put this Warning, it is certainly because I kick the bucket here somehow already :)

The other warning I had was the changes in vrrp_leave_fault, but ready your update, I cant see good reason why this update is wrong.

Ok I close your req and w8 your next one.

I must also ask to Alexey, but do you think I can push the current Master branch as new stable release ? (IMHO, I can) let me know please

pqarmitage:
On Wed, 2016-03-02 at 10:19 -0800, Alexandre Cassen wrote:

> I have a HUGE warning in my test tool in that vrrp_scheduler. I
> certainly dont fully remind the deep reason but if I put this Warning,
> it is certainly because I kick the bucket here somehow already :)
> 
> The other warning I had was the changes in vrrp_leave_fault, but ready
> your update, I cant see good reason why this update is wrong.
> 

The alternative approach that has occurred to me is to do the same as
the config "state MASTER" does, if priority == 255 and nopreempt is not
set.

Since this is the same as having:
state MASTER
priority 255
in the VRRP_instance config, it can't break anything that works now,
since that can already be configured.

My reason for doing this is that the RFCs state that if the priority is
255, then the instance should transition to master as it does if state
MASTER is specified, so unless the configuration explicitly stops that
(i.e.nopreempt), then it should do so.

The patch I therefore now propose is:


> +       else if (vrrp->base_priority == VRRP_PRIO_OWNER && !vrrp->nopreempt) {
> +               vrrp->init_state = VRRP_STATE_MAST;
> +               vrrp->wantstate = VRRP_STATE_MAST;
> +       }

in function vrrp_complete_instance in vrrp.c. Does that seem reasonable
to you? If so, I can do some final testing and submit a pull request
later this evening.

> Ok I close your req and w8 your next one.
> 
> I must also ask to Alexey, but do you think I can push the current
> Master branch as new stable release ? (IMHO, I can) let me know please
> 

From my point of view if you can include the latest VRRPv3 SNMP code,
once I have submitted another pull request, then I would be more than
happy for a new stable release.

I have also just created a patch that removes all extra spaces/tabs and
the ends of lines, spaces before tabs anywhere in lines, and replaces
occurrences of 8 spaces with tabs; it might be worth including that in
the new stable release, since it just makes the source code a bit
tidier. I have compiled the code after these changes and the executable
files before and after the changes are identical, so it doesn't change
the executable at all. Would you be happy to include this patch as well?

Just for your information, I have compiled the latest code on a variety
of systems from Centos 6.7, through to the latest Fedora 23, with
kernels from 2.6.32 to 4.4.3, on both 32 bit and 64 bit systems, and
with various combinations of libraries installed, to try and make sure I
haven't created any build problems like I did a couple of weeks ago, and
everything has build fine. Last night I also went though full configure
and compile cycles for all distinct combinations of all the configure
--disable-... and --enable-... options, 2048 builds in total, and they
all built without any warnings (it took 3 hours to complete).

With best wishes.
Quentin

acassen:
Ok Quentin, I am not in a hurry. I take that mutex on you to publish a new release on website.

(for VRRP: in RFC there is sync_group, and this feature introduce some tricky side effect in FSM if not done with a lot of attention while coding)

acassen:
s/there is/there is no/g

andriyanov:
> I must also ask to Alexey, but do you think I can push the current Master branch as new stable release ? (IMHO, I can) let me know please

I don't have any objections, but I don't use either the latest master, vrrp or snmp subsystems in production, so can't tell how stable the new master is.

pqarmitage:
On Wed, 2016-03-02 at 11:00 -0800, Alexandre Cassen wrote:

> Ok Quentin, I am not in a hurry. I take that mutex on you to publish a
> new release on website.
> 
> (for VRRP: in RFC there is sync_group, and this feature introduce some
> tricky side effect in FSM if not done with a lot of attention while
> coding)
> 

Alexandre,

Many apologies for the delay in completing this. I found a problem with
ip addresses being left on interfaces if keepalived abnormally
terminated, and it has taken a while to get that fix sorted.

From my perspective this would be a good point to publish a new release.

With many thanks for your help and guidance,

Quentin

acassen:
NO PROBLEMS ! you are very welcome.

Will merge this today. Will just fix a small logic issue.

Thanks for all this work !

Issue #199, opened 2015-11-23
vrrp dns dependency


tbsky:
hi:
    I am testing keepalived vrrp on openwrt box. I notice sometimes vrrp will take about 3-5 minutes to startup. after debugging, I found when keepalived starts, it will do a name resolve of hostname. if the name resolve system is trying hard to do the work and can not give out result, then keepalived will just wait there for minutes,until the resolve failed.
   
  if I put  my hostname to /etc/hosts then the situation is solved. but I am curious why keepalived try to resolve the hostname when start, is it necessary?

pqarmitage:
keepalived does a name resolve of hostname in get_local_name(). It gets the local name from the nodename returned by uname(). In case this name is not a FQDN, or in case it is an alias, it calls getaddrinfo() to resolve this into a non-alias FQDN for the host.

keepalived called get_local_name() several times at startup if certain defaults are being used. For each of the vrrp process and checker process it would call get_local_name() once to get the router_id, and once to set the default email_from name for sending SMTP messages. It would also call get_local_name() every time it sent an email, to get the name to use in the HELO message.

Since this clearly calling get_local_name() more often than necessary, and so commit 9d028ac reduces the total number of calls to one per thread (saving the name for subsequent use in HELO messages). The commit also introduces an addition global configuration parameter, smtp_helo_name, and so if that, router_id and notification_email_from are all set in the global configuration, and if an smtp server is specified by IP address rather than host name, then keepalived shouldn't make and DNS queries at start up.

PR #270, opened 2016-03-09
Reduce number of calls to getaddrinfo() reducing DNS lookups

pqarmitage:
Issue #199 identified that keepalived could be very slow to start up
if DNS queries resolving the local host name were very slow.

keepalived could call get_local_name() twice at startup for each of the
vrrp and checker threads, in order to get the router_id and also the
set the default email_from address, unless these were manually configured.
It also called getaddrinfo() for the local host name every time it sent an
email, in order to set the HELO string.

This commit reduces the total number of calls per thread to one, using
the returned name to set the router_id, email_from address and the HELO
name (which is now saved for reuse every time an email is sent).

In order to allow no DNS lookups to be made, an extra global configuration
parameter is introduced, smtp_helo_name, so that if router_id, smtp_email_from
and smtp_helo_name are all set in the configurationi (and also the smtp_server
is specified by IP address rather than by name), then no DNS queries will
be made.

Signed-off-by: Quentin Armitage <quentin@armitage.org.uk>

### DIFF
--- a/doc/keepalived.conf.SYNOPSIS
+++ b/doc/keepalived.conf.SYNOPSIS
@@ -28,7 +28,10 @@ global_defs {				# Block identification
        ...
     }
     notification_email_from <EMAIL ADDRESS> # Email From dealing with SMTP proto
+					   #   defaults to keepalived@<local host name>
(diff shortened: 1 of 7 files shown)


acassen:
I can revert if more useful for your dev task ?

PR #316, opened 2016-05-08
More verbose logging on (effective) priorities

rolffokkens:
While using the weight feature and analysing its effect it is helpful to have additional log info on the effective priority of a keepalived instance and received priorities of other instances.

Furthermore there's a oddity regarding a potentially shaded if clause for prority == 0 detection. It seems like this clause is shaded by the if clause right above it. If so the if clause is obsolete, or (which is in this PR) it should be "taken out of the shadow".


pqarmitage:
These commits were applied in pull request #330. Many thanks for the patches.

PR #319, opened 2016-05-10
vrrp: update struct msghdr

stintel:
The vrrp netlink code assumes an order for the members of struct msghdr.
This breaks recvmsg and sendmsg with musl libc on mips64. Fix this by
using designated initializers instead.


pqarmitage:
This commit was incorporated in pull request #330, which has now been applied. Many thanks for the patch.

PR #323, opened 2016-05-13
Bind read sockets to particular interface

DStape:
Otherwise, since we use RAW sockets, we will receive IPPROTO_VRRP
packets that come in on any interface.


DStape:
https://github.com/acassen/keepalived/pull/323

Issue #296, opened 2016-04-12
"preempt_delay " set, but doesn't  work ？？


TigererFeng:
Maybe ms_down_timer expire,the device (it's fault state) transition to master state ,but “preempt_delay 60“ configuration doesn't work ,Does Keepalived1.2.19 support preemption delay function ??

wtaps:
Same question.

wtaps:
You can see the changelog, i think Keepalived 1.2.19 was supported.

TigererFeng:
it not support !!
   I think it's necessary to wait some secends to transition to backup state (the initial state in configure file) from fault state (when network dropped); 
    Because it's not steady when network resume  ** result in**  that  vrrp stance being backup state can not receive vrrp packet from master device , so it send Gratuitous ARP to gatway , then it (Device A)received the vrrp packet that higher prio advert from master device (Device B), so  transition to backup state , But** gatway keep the Mac address of Device A **other than the Mac address of master Device B !!!!!

Cugar15:
I'm also having difficulties to get 'preemt_delay' or 'nopreempt' to work (latest revision).
Neither setting shows any effect.

jonasj76:
Maybe this commit fixes the problem: fdd37e4.
https://github.com/jonasj76/keepalived/tree/respect-preempt-from-fault

pqarmitage:
I think I can see where the problem is (in vrrp_state_backup(), but I need to think this through a bit further to double check I have got this right. I'm busy until after 18:00 GMT, but I'll investigate it further then.

PR #346, opened 2016-06-03
Fix preempt issue from fault state

pqarmitage:
The main commit is the penultimate one (fix preempt delay), which resolves the issue of preempt_delay and nopreempt being ignored when transitioning out of fault state.

While investigating the issue, a number of other issues were found relating to state changes following receiving adverts, and included are commits relating to those issues.

This pull request should resolve issue #296.

### DIFF
--- a/doc/KEEPALIVED-MIB
+++ b/doc/KEEPALIVED-MIB
@@ -22,13 +22,15 @@ IMPORTS
         FROM SNMPv2-TC;
 
 keepalived MODULE-IDENTITY
-     LAST-UPDATED "201605221540Z"
+     LAST-UPDATED "201606030000Z"
      ORGANIZATION "Keepalived"
      CONTACT-INFO "http://www.keepalived.org"
      DESCRIPTION
         "This MIB describes objects used by keepalived, both
          for VRRP and health checker."
-     LAST-UPDATED "201605221540Z"
+     REVISION-UPDATED "201606030000Z"
+     DESCRIPTION "update comment for vrrpInstancePreemptDelay"
+     REVISION-UPDATED "201605221540Z"
      DESCRIPTION "smtpServerPort added"
      REVISION "201510270000Z"
      DESCRIPTION "routerId added to traps variables"
@@ -496,7 +498,7 @@ vrrpInstancePreemptDelay OBJECT-TYPE
     MAX-ACCESS read-only
     STATUS current
     DESCRIPTION
-        "Delay after startup until preemption can happen. 0 means that there is no delay."
+        "Delay after startup or lower priority advert received until preemption can happen. 0 means that there is no delay."
     ::= { vrrpInstanceEntry 14 }
 
 vrrpInstanceAuthType OBJECT-TYPE
--- a/doc/keepalived.conf.SYNOPSIS
+++ b/doc/keepalived.conf.SYNOPSIS
@@ -318,8 +318,8 @@ vrrp_instance <STRING> {		# VRRP instance declaration
     }
 
     nopreempt					# Override VRRP RFC preemption default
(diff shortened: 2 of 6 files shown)


Cugar15:
HI, I tested this commit and still can't get this feature (nopreempt) to work.

Here is what I do - maybe, I have  wrong understanding..

1) Added nopreempt to the BACKUP vrrp_instance (not to the master instance)
2) Once I either stop keepalived on the master machine - or stop the vrrp_script monitored process -
    the backup machine transitions into MASTER state.
3) Once I either start keepalived on the master machine - or star the vrrp_script monitored process -
    the backup machine transitions into BACKUP state.

My expectition is, that the backup machine stays in MASTER state.

I need to add, that only the backup machine has this new revision of keepalived installed.

Thanks!

pqarmitage:
If you don't want the master machine to preempt the backup machine, you need to configure nopreempt on the master machine, to tell it not to preempt the backup machine.

It is the master machine that needs the configuration (and potentially the new code) since it is that machine that receives a lower priority advert, and it has to be told (via the nopreempt configuration) not to preempt the backup machine.

I hope that helps.

acassen:
Hi Quentin,

I must read deeply those patches... Each time I am changing something into the VRRP FSM it breaks somethings else. VRRP FSM is really tricky :/ there is plenty of corner cases specially when using sync_group. This feature is not present in VRRP RFC and can bring some complicate use cases.

regs,

Cugar15:
Hi Quentin,

   what you are writing makes perfect sense - however still no success when modifying the master machine.

My confusion is create because of the following message when setting nopreempt in the master configuration:

(VI_1): Warning - nopreempt will not work with initial state MASTER

Looking at the code, this is a warning only...
Is it a correct understand, that this warning is supposed to mean, that once keepalieved is started, it will listen to lower priority adverts??

Sorry to aks this maybe stupid questions..

pqarmitage:
@Cugar15 can you try setting state to be BACKUP rather than MASTER on the vrrp instance, since that is what it is warning about. There is no problem if all machines specify state BACKUP; specifying state MASTER just allows that system to become master more quickly after startup.

Cugar15:
@Quentin, OK, configuring BACKUP state seems to work. Can you confirm the following behaviour - which I think makes sense - to double check:

Scenario A:
1) Start keepalived on master (higher priority) machine. (Machine will be MASTER)
2) Then start keepalived on backup (lower priority) machine. (Machine will be BACKUP)
3) Stop keepalived on master machine -> backup machine will be MASTER
4) Start keepalived on master machine 
           -> master machine will enter BACKUP
           -> backup machine will stay MASTER

Scenario B: Turn startup order around
1) Start keepalived on backup (lower priority) machine. (Machine will be MASTER)
2) Then start keepalived on master (higher priority) machine. (Machine will be BACKUP)

Cheers!

pqarmitage:
@Cugar15 The behaviour you have described above is correct when nopreempt is specified on the higher priority machine.

Cugar15:
@Quentin - thanks for confirming.

FYI, out of curiosity, I went back to version 1.2.21. I can confirm, that
1.2.21 is also working as described above once getting the configuration
clean.

Cheers and thanks!


2016-06-06 8:41 GMT+02:00 pqarmitage <notifications@github.com>:

quentin:
Hi, I don't know this project, I don't understand why I am part of this discussion.

pqarmitage:
@Cugar15 mistakenly typed @Quentin instead of @pqarmitage (I am called Quentin too).

Cugar15:
Sorry, did not know that this system is taking the 'at' sign and associates an email address with it...

PR #353, opened 2016-06-07
Fix IPv4 vIP removal when addr matches pre-existing interface addr

DStape:
For IPv4 vIPs keepalived adds a /32 to the underlying interface. If
this address matches an address already configured, e.g. a /24, when
this vIP is eventually removed due to a configuration change or
keepalived shutdown, the original address matching the vIP, outside
of keepalived's control, is removed instead. This behaviour is
incorrect. The /32 added by keepalived should be the address being
removed. Keepalived should not be touching any addresses it does not
create.

This commit re-instates d80c171 which was removed in ef73279. The
IFA_ADDRESS attribute is included in the netlink delete address
message so that the /32 keepalived added is removed instead of the
pre-existing interface address.

### DIFF
--- a/keepalived/vrrp/vrrp_ipaddress.c
+++ b/keepalived/vrrp/vrrp_ipaddress.c
@@ -110,10 +110,16 @@ netlink_ipaddress(ip_address_t *ipaddress, int cmd)
 		addattr_l(&req.n, sizeof(req), IFA_LOCAL,
 			  &ipaddress->u.sin.sin_addr, sizeof(ipaddress->u.sin.sin_addr));
 
-		if (cmd == IPADDRESS_ADD)
+		if (cmd == IPADDRESS_ADD) {
 			if (ipaddress->u.sin.sin_brd.s_addr)
 				addattr_l(&req.n, sizeof(req), IFA_BROADCAST,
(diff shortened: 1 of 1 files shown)


acassen:
hmm, well, my very first consideration for keepalived code would be to not try to be smarter than admin using it :)

=> If admin want to use as VRRP VIP an already used ip address on interface... well, it is his concern, he should at least know what he is doing. No ?

DStape:
That is a fair point, however I'd argue that this is a bug. Keepalived should never be touching any interfaces or addresses that it does not create/add. So in this case, this bug in keepalived means that the wrong address is being removed. If I had acquired an interface address using dhcp and then set my vip to match, once the associated group goes into backup or keepalived is shutdown, that dhcp address would be removed - which is clearly incorrect. The address being removed should be the one keepalived added.

Why would an admin want to set the vip equal to the primary address? Typically, this will happen when there are very few available ipv4 addresses to use. Whilst this scenario may not be very common, it's still a scenario some admins will face in their networks.

pqarmitage:
+1 for this patch. I hadn't realised this somewhat unexpected behaviour when I deleted those lines in commit ef73279. There were other attributes being set that weren't relevant on deletion, and unfortunately I thought that this was one of those too.

DStape:
No problem @pqarmitage.

PR #356, opened 2016-06-07
Fix typo in function call

DStape:
Second argument delimited with full stop intead of a command, so
compilation fails.

"." -> ","


DStape:
My bad, apologies...

PR #369, opened 2016-06-25
Add temporary workaround re segfault when reloading with vmacs

pqarmitage:
Issue #365 identified that the vrrp child process intermittently
segfaults when reloading configuration. Investigation has shown
that vrrp->ifp->ifindex (and vrrp->ifp->ifname) are being overwritten,
and hence in netlink_link_del_vmac() function if_get_by_ifindex() is
returning null (since it is passed an invalid ifindex), and this return
value is passed unchecked to reset_interface_parameters(), which
then causes the segfault.

This commit is a temporary workaround to test the return value of
if_get_by_ifindex() and if it is NULL, then reset_interface_parameters()
isn't called, but an error message is logged.

Further investigation is needed to ascertain why the fields in the
interface struct are being overwritten.

Signed-off-by: Quentin Armitage <quentin@armitage.org.uk>


pqarmitage:
The cause of the segfault has been found, and this commit is no longer needed.

PR #379, opened 2016-07-06
Fix unused global vrrp_garp_master_refresh

lennardk:
global_data->vrrp_garp_refresh was not propagating into vrrp->garp_refresh


pqarmitage:
Acked-by: Quentin Armitage <quentin@armitage.org.uk>

PR #397, opened 2016-08-10
Convert build system to automake

pqarmitage:
The main commit is to move to using automake. The main benefit of this is that all include dependencies are tracked by automake/autoconfigure, rather than having to be manually maintained. A further benefit is that make tarball no longer kills the build environment.

On any system which doesn't have automake/autoconfigure installed, running ./configure continues to work as before, with the same syntax.

All make targets are preserved from prior to this commit.

The only real external change is that the VERSION file no longer exists (although it could be created by the build if desired). The version is now defined in the configure.ac in the AC_INIT statement. When new versions of keepalived are released, it will now be configure.ac that needs to be edited for the new version.

Tests in configure.ac or the code based on versions of Linux/glibc now generally state which versions are effected, to make it easier to remove tests supporting older versions of Linux/glibc when those version are no longer supported by keepalived.
 
This has been tested on Fedora 16, 18, 23 and 24, Debian Wheezy and Jessie and Centos 6.8, with kernels from 2.6.32 to 4.7.

The other commits fix building on 32 bit systems, fix building without one of LVS and VRRP, and adds an include needed on some systems.


acassen:
Wow well done Quentin...

Issue #392, opened 2016-08-08
Netlink: Received message overrun (No buffer space available)


jslocinski:
Opening new ticket for clearance, mentioned about this issue here [#390](https://github.com/acassen/keepalived/issues/390)
How to tune the buffer size? Tried to increase buffers system-wide via sysctl - didn't help.
Message appears when playing with dozen or hundred of instances on different interfaces with or w-out VMAC option enabled. Starts at level of about 15 instances (with vmac setup) or ~200 instances (no-vmac setup) or at. I start seeing that error at the time of becoming MASTER and sending GARPs.
How can I tune it up properly?

PS.
Tested on 1.2.22 and 1.2.23. Used dozen kB of rmem_max/default buffer size up to few or even hundred of Megs.. same result.

Best regards!

pqarmitage:
Could you please post both your with-VMAC and without-VMAC configs (or scripts that generate them), so that we can try to reproduce the problem. If you want to anonimise IP addresses etc, please just replace the first octet with 10; otherwise addresses are replaced with XXXs, we can't see the relationship between them.

It is interesting that you only seem to need a relatively small number of VMACs to generate the problem, compared to the number of non-VMAC interfaces.

I have a number of thoughts about what might be worth trying, but it would be really helpful to see your configs first to make sure that what I suggest makes sense.

jslocinski:
Sure,
simple scripts:

test_keepalived.sh
```
#!/bin/bash
if [ $# -ne 2 ]; then
  echo "USAGE: $0 (up|down) NBR_INSTANCES"
  exit 1
fi

N=$2

if [ "$1" == "up" ]; then
  cat <<EOF >/run/conf/keepalived.conf
global_defs {
}
EOF
  ip link add bondZ type dummy
  ip link set up dev bondZ
elif [ "$1" == "down" ]; then
  service keepalived stop
  unlink /run/conf/keepalived.conf
fi

for i in $(seq 1 $N); do
  ifi=bondZ.$i
  if [ "$1" == "up" ]; then
    ip link add link bondZ name $ifi type vlan id $i
    ip link set up dev $ifi
    cat /root/keepalived.template.test | sed s"/__ID__/$i/g" >> /run/conf/keepalived.conf
  fi
done

if [ "$1" == "up" ]; then
  sleep 1
  service keepalived start
elif [ "$1" == "down" ]; then
  ip link del bondZ
fi
```

/root/keepalived.template.test:
```
[...]

PR #401, opened 2016-08-13
Stop "Netlink: Received message overrun (No buffer space available)" messages

pqarmitage:
keepalived was only polling for netlink messages on the kernel reflection channel on a timed basis. If a large number of vmacs were added at startup, the netlink socket would run out of receive buffers, and an ENOBUFS error would occur when the netlink socket was read.

Polling for netlink messages after adding each interface resolves the problem.

This should resolve issue #392.

### DIFF
--- a/keepalived/vrrp/vrrp.c
+++ b/keepalived/vrrp/vrrp.c
@@ -53,6 +53,7 @@
 #include "utils.h"
 #include "notify.h"
 #include "bitops.h"
+#include "vrrp_netlink.h"
 #if !HAVE_DECL_SOCK_CLOEXEC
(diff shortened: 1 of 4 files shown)


keepalived/vrrp/vrrp.c
```diff
@@ -2234,8 +2235,6 @@ vrrp_complete_instance(vrrp_t * vrrp)
 		vrrp->vipset = true;	/* Set to force address removal */
 		vrrp_restore_interface(vrrp, false, true);
 	}
-
-	return 1;
```
acassen: hmm, this return is needed.

PR #415, opened 2016-08-31
Adding libipvs flags support

nirshatsmellanox:
Support passing libipvs flags by configuration file.
keyword for flags is lb_flags
for example:
lb_flags flag-1 flag-2 hashed

### DIFF
--- a/keepalived/check/check_parser.c
+++ b/keepalived/check/check_parser.c
@@ -124,31 +124,26 @@ static void
 lbflags_handler(vector_t *strvec)
 {
 	virtual_server_t *vs = LIST_TAIL_DATA(check_data->vs);
-	int i;
-	char* str;
-	vector_foreach_slot_from(strvec,str,i,1)
-	{
-		if (!strcmp(str, "persistant"))
-			vs->flags |= IP_VS_SVC_F_PERSISTENT;
-		if (!strcmp(str, "hashed"))
-			vs->flags |= IP_VS_SVC_F_HASHED;
-		if (!strcmp(str, "one-packet"))
-			vs->flags |= IP_VS_SVC_F_ONEPACKET;
-		if (!strcmp(str, "flag-1"))
-			vs->flags |= IP_VS_SVC_F_SCHED1;
-		if (!strcmp(str, "flag-2"))
-			vs->flags |= IP_VS_SVC_F_SCHED2;
-		if (!strcmp(str, "flag-3"))
-			vs->flags |= IP_VS_SVC_F_SCHED3;
-		if (!strcmp(vs->sched , "sh") )
-		{
-			if (!strcmp(str, "sh-port")  )
-				vs->flags |= IP_VS_SVC_F_SCHED_SH_PORT;
-			if (!strcmp(str, "sh-fallback"))
-				vs->flags |= IP_VS_SVC_F_SCHED_SH_FALLBACK;
-		}
-	}
+	char* str=vector_slot(strvec,0);
 
+	if (!strcmp(str, "hashed"))
+		vs->flags |= IP_VS_SVC_F_HASHED;
+	if (!strcmp(str, "flag-1"))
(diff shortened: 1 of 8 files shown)


keepalived/check/check_parser.c
```diff
@@ -115,12 +115,42 @@ lbalgo_handler(vector_t *strvec)
...
+lbflags_handler(vector_t *strvec)
+{
+	virtual_server_t *vs = LIST_TAIL_DATA(check_data->vs);
+	int i;
+	char* str;
+	vector_foreach_slot_from(strvec,str,i,1)
+	{
+		if (!strcmp(str, "persistant"))
+			vs->flags |= IP_VS_SVC_F_PERSISTENT;
+		if (!strcmp(str, "hashed"))
```
pqarmitage: Persistence can already be set using keywords `persistence_timeout` and/or `persistence_granularity`

keepalived/check/check_parser.c
```diff
@@ -115,12 +115,42 @@ lbalgo_handler(vector_t *strvec)
...
+			vs->flags |= IP_VS_SVC_F_HASHED;
+		if (!strcmp(str, "one-packet"))
+			vs->flags |= IP_VS_SVC_F_ONEPACKET;
```
pqarmitage: We already have an option for one-packet scheduling - keyword `ops`.

keepalived/check/check_parser.c
```diff
@@ -115,12 +115,42 @@ lbalgo_handler(vector_t *strvec)
...
+static void
+lbflags_handler(vector_t *strvec)
+{
+	virtual_server_t *vs = LIST_TAIL_DATA(check_data->vs);
+	int i;
+	char* str;
+	vector_foreach_slot_from(strvec,str,i,1)
+	{
+		if (!strcmp(str, "persistant"))
+			vs->flags |= IP_VS_SVC_F_PERSISTENT;
```
pqarmitage: Persistence can already be set using keywords `persistence_timeout` and/or `persistence_granularity`

keepalived/check/check_parser.c
```diff
@@ -115,12 +115,42 @@ lbalgo_handler(vector_t *strvec)
...
+		if (!strcmp(str, "flag-3"))
+			vs->flags |= IP_VS_SVC_F_SCHED3;
+		if (!strcmp(vs->sched , "sh") )
+		{
+			if (!strcmp(str, "sh-port")  )
+				vs->flags |= IP_VS_SVC_F_SCHED_SH_PORT;
+			if (!strcmp(str, "sh-fallback"))
+				vs->flags |= IP_VS_SVC_F_SCHED_SH_FALLBACK;
+		}
+	}
```
pqarmitage: It would be more consistent with the current way of configuring virtual servers to add new keywords `hashed`, `flag-1`, `flag-2`, 'flag-3`, `sh-port` and `sh-fallback` at the same level as `ops`, so each one has their own handler function, and not add the lb_flags keyword.

ops_handler() could then be modified to set the relevant bit in vs->flags, and the vs->ops field could be removed from the _virtual_server struct.

pqarmitage:
The documentation files doc/keepalived.SYNOPSIS and doc/man/man5/keepalived.conf should be updated to add the new options (we want users to know they can use them!).

The function dump_vs() should be updated to report the new settings.

For updating the documentation and dump_vs() it's easiest to search for ops, and copy that.

The snmp MIB and supporting code should be updated to report these new options. I'm quite happy to provide a further patch with that update if you'd rather not do that yourself.

nirshatsmellanox:
make another commit with new implementation + added the information to documentation.

I handled the sh-port and sh-fallback validity check in check_parser() and not in ipvswrapper() since these flags values are identical to flag-1,flag2 which are valid for all schedulars.
ops validity check remains at ipvswrapper

PR #420, opened 2016-09-12
Updates queued for 1.2.24

pqarmitage:
Add libipvs flags support - from @nirshatsmellanox 
Don't run a vrrp_update_priority thread if we are the address owner
Clarify namespace instructions - from @lzmartinico
Add DBus support - from @lzmartinico
Increase level of compiler warnings and resolve warnings
Fix testing and report of problems with netlink library installation
Bring generation of keepalived.spec up to date, and make it match configure options


ajf:
I'm assuming this will be in 1.1.25? Totally looking forward for b531076 for sh-port.

vincentbernat:
The shipped dbus policy seems to be overly broad:

```xml
<busconfig>
	<policy user="root">
		<allow own="org.keepalived.Vrrp1"/>
		<allow send_destination="org.keepalived.Vrrp1"/>
	</policy>
	<policy context="default">
		<allow send_interface="org.freedesktop.DBus.Introspectable" />
		<allow send_interface="org.freedesktop.DBus.Peer" />
		<allow send_interface="org.freedesktop.DBus.Properties" />
	</policy>
</busconfig>
```

I wonder what is the goal of the second block. It seems to allow anyone to set any properties on any object.

PR #430, opened 2016-09-26
Add check for PE selection support

rik78:
Without this patch, pe selection doesn't work with kernel 3.10 on centos 7.


pqarmitage:
@rik78 Good catch!

The problem isn't just with kernel 3.10 on centos 7; it is on any kernel on any distro that supports PE_NAME.

The problem is that IPVS_SVC_ATTR_PE_NAME is an enumeration constant rather than a defined name, but the test #ifdef IPVS_SVC_ATTR_PE_NAME just checks for the defined name. Basically it's my fault for not adding the right type of test.

Since you've highlighted this, I've had a look and there are a number of other cases with the same issue, so I'll follow your patch up with a further patch to resolve those.

rik78:
Thanks!

PR #429, opened 2016-09-21
Add DNS health checker

pandax381:
Hi,

I implemented DNS helath checker.
It has been running stable for several years in our system.

I want merge it to the built-in health checker.

There is also a way to use MISC_CHECK, but we avoided it for the following reasons.

- MISC_CHECK requires /bin/sh

 MISC_CHECK is using the system(3). It is equivalent to the "/bin/sh -c". Therefore, we can't use MISC_CHECK in environment does not have /bin/sh. It is, e.g. not uncommon in the application container environment.

- Risk of the process increase in MISC_CHECK

 If misc_path of the program has blocked, the process will increase. You can check the symptoms with the following settings.

 ```
MISC_CHECK {
      misc_path "/bin/sleep 3600"
      misc_timeout 10 
}
 ```

 It's results.

 ```
  UID   PID  PPID  PGID   SID COMMAND
    0 41010     1 41010 41010 /sbin/keepalived
    0 41013 41010 41010 41010  \_ /sbin/keepalived
    0 41361 41013 41010 41010  |   \_ /sbin/keepalived
    0 41362 41361 41010 41010  |   |   \_ sh -c /bin/sleep 3600
    0 41363 41362 41010 41010  |   |       \_ /bin/sleep 3600
    0 41364 41013 41010 41010  |   \_ /sbin/keepalived
    0 41365 41364 41010 41010  |   |   \_ sh -c /bin/sleep 3600
    0 41366 41365 41010 41010  |   |       \_ /bin/sleep 3600
    0 41367 41013 41010 41010  |   \_ /sbin/keepalived
    0 41368 41367 41010 41010  |       \_ sh -c /bin/sleep 3600
    0 41369 41368 41010 41010  |           \_ /bin/sleep 3600
    0 41014 41010 41010 41010  \_ /sbin/keepalived
    0 41019     1 41010 41010 sh -c /bin/sleep 3600
    0 41020 41019 41010 41010  \_ /bin/sleep 3600
    0 41025     1 41010 41010 sh -c /bin/sleep 3600
    0 41026 41025 41010 41010  \_ /bin/sleep 3600
    0 41031     1 41010 41010 sh -c /bin/sleep 3600
    0 41032 41031 41010 41010  \_ /bin/sleep 3600
 ```
 MISC_CHECK is calling system(3) after fork(2). In check_misc.c, it's sending SIGTERM and SIGKILL on timeout. However, the signal is received only in the process that called system(3), the subsequent process is not propagated.

[...]

PR #432, opened 2016-09-27
[WIP] Fix process increase

pandax381:
Hi,

This is the proposal to the matter of mentioned in #429.
- Script process does not respond to SIGTERM (fe9638b)
  
  ```
  Sep 26 16:31:03 lb01 Keepalived_healthcheckers[28302]: Process [28304] didn't respond to SIGTERM
  Sep 26 16:31:03 lb01 Keepalived_healthcheckers[28302]: pid 28304 exited due to signal 9
  Sep 26 16:31:06 lb01 Keepalived_healthcheckers[28302]: Process [28307] didn't respond to SIGTERM
  Sep 26 16:31:06 lb01 Keepalived_healthcheckers[28302]: pid 28307 exited due to signal 9
  Sep 26 16:31:09 lb01 Keepalived_healthcheckers[28302]: Process [28366] didn't respond to SIGTERM
  Sep 26 16:31:09 lb01 Keepalived_healthcheckers[28302]: pid 28366 exited due to signal 9
  ```
  
  I think there is a problem in the signal_handler_init().
  In signal_handler_init(), it remember the original disposition of signals.
  It's the first time will function normally. But, the second time it seems to unexpected behavior.
  Because, in the first round all of the signal is set to SIG_IGN. For this reason, dose not be restored in the signal_handler_script(). SIGTERM remains ignored.
  
  It has been modified to be stored only the first time.
- When the script is blocked, process will increase (4c471b2)
  
  I wrote...
  
  > MISC_CHECK is calling system(3) after fork(2). In check_misc.c, it's sending SIGTERM and SIGKILL on timeout. However, the signal is received only in the process that called system(3), the subsequent process is not propagated.
  
  Fixed to separate the process group. And it was to send a signal to the process group.
- Signal propagation for the script process groups (f0c5783)
  
  Since the separation of the process group of script-related process, the signal has become not propagate. So, we propagate important signals. It's SIGTERM/SIGINT and SIGHUP. 

Initially, I was thinking only of MISC_CHECK. Then, vrrp_script was also noticed that it is the same.

In this proposal, process was no longer increase.

What do you think about this proposal?
I want to listen to your opinion.

Best regards,
Masaya Yamamoto

### DIFF
--- a/keepalived/check/check_daemon.c
+++ b/keepalived/check/check_daemon.c
@@ -199,7 +199,7 @@ sigend_check(__attribute__((unused)) void *v, __attribute__((unused)) int sig)
 static void
 check_signal_init(void)
 {
-	signal_handler_init();
+	signal_handler_init(0);
 	signal_set(SIGHUP, sighup_check, NULL);
 	signal_set(SIGINT, sigend_check, NULL);
 	signal_set(SIGTERM, sigend_check, NULL);
--- a/keepalived/check/check_daemon.c
+++ b/keepalived/check/check_daemon.c
@@ -55,9 +55,11 @@ static char *check_syslog_ident;
 static void
 stop_check(int status)
 {
+	/* Terminate all script process */
+	script_killall(master, SIGTERM);
+
 	/* Destroy master thread */
 	signal_handler_destroy();
-	script_killall(master, SIGTERM);
 	thread_destroy_master(master);
 	free_checkers_queue();
 	free_ssl();
@@ -215,11 +217,13 @@ reload_check_thread(__attribute__((unused)) thread_t * thread)
 
 	log_message(LOG_INFO, "Got SIGHUP, reloading checker configuration");
 
+	/* Terminate all script process */
+	script_killall(master, SIGTERM);
+
 	/* Destroy master thread */
 #ifdef _WITH_VRRP_
 	kernel_netlink_close();
 #endif
-	script_killall(master, SIGHUP);
 	thread_cleanup_master(master);
 	free_global_data(global_data);
 
--- a/keepalived/check/check_daemon.c
+++ b/keepalived/check/check_daemon.c
@@ -35,6 +35,7 @@
 #include "pidfile.h"
 #include "daemon.h"
 #include "signals.h"
+#include "notify.h"
 #include "process.h"
 #include "logger.h"
 #include "list.h"
@@ -56,6 +57,7 @@ stop_check(int status)
 {
 	/* Destroy master thread */
 	signal_handler_destroy();
+	script_killall(master, SIGTERM);
 	thread_destroy_master(master);
 	free_checkers_queue();
 	free_ssl();
@@ -217,6 +219,7 @@ reload_check_thread(__attribute__((unused)) thread_t * thread)
 #ifdef _WITH_VRRP_
 	kernel_netlink_close();
 #endif
+	script_killall(master, SIGHUP);
 	thread_cleanup_master(master);
 	free_global_data(global_data);
 
--- a/keepalived/vrrp/vrrp_scheduler.c
+++ b/keepalived/vrrp/vrrp_scheduler.c
@@ -1094,7 +1094,7 @@ vrrp_script_child_thread(thread_t * thread)
 				log_message(LOG_INFO, "VRRP_Script(%s) timed out", vscript->sname);
 			vscript->result = 0;
 		}
-		kill(pid, SIGTERM);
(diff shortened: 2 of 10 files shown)


keepalived/check/check_daemon.c
```diff
@@ -217,6 +219,7 @@ reload_check_thread(__attribute__((unused)) thread_t * thread)
 #ifdef _WITH_VRRP_
 	kernel_netlink_close();
 #endif
+	script_killall(master, SIGHUP);
```
pqarmitage: This sends a SIGTERM to the child process if keepalived is stopping, and a SIGHUP to the child process if the config is being reloaded. Whichever of these two situations is occurring, does the child process need to know the difference. All we want is that it stops, although it might be possible that a child process wants to do something different depending on whether keepalived is stopping or reloading (but I can't think of a reason why), but then the problem is that if the child receives a SIGTERM, it can't distinguish between the keepalived process terminating and the child has been running for too long.

The same applies further down in `misc_check_child_thread()` and `stop_vrrp()`.

lib/notify.c
```diff
@@ -119,3 +120,22 @@ system_call_script(thread_master_t *m, int (*func) (thread_t *), void * arg, lon
...
+script_killall(thread_master_t *m, int signo)
+{
+	thread_t *thread;
+	pid_t p_pgid, c_pgid;
+
+	thread = m->child.head;
+
+	p_pgid = getpgid(0);
+
+	while (thread) {
```
pqarmitage: I think this loop needs to be called with SIGCHLD blocked, otherwise the `thread_child_handler()` signal handler may run and modify the child queue while we are modifying it.

lib/signals.c
```diff
@@ -185,6 +185,7 @@ signal_handler_init(void)
 	int sig;
 	struct sigaction act, oact;
 	int n;
+	static int remember = 1;
```
pandax381: I agree with you! (I was trying to suppress to minimal changes)
I'll fix so that it does not use a static variable.

lib/notify.c
```diff
@@ -105,6 +105,7 @@ system_call_script(thread_master_t *m, int (*func) (thread_t *), void * arg, lon
 
 	/* In case of this is parent process */
 	if (pid) {
+		setpgid(pid, pid);
```
pandax381: Oops... I'm sorry!
It's completely my mistake.

I agree with you. It should be setpgid() is child process.

lib/notify.c
pandax381: Sorry, I forgot to block SIGCHLD.
Use sigprocmask().

keepalived/check/check_daemon.c
pandax381: I chose a SIGHUP when reload just in case.
However, we want that it terminate. So, it looks good in SIGTERM.

I thought as a child process is there is no need to understand the cause of SIGTERM. And I think so child process (as the script program) would be required to terminate with SIGTERM.

keepalived/vrrp/vrrp_scheduler.c
```diff
@@ -1094,7 +1094,7 @@ vrrp_script_child_thread(thread_t * thread)
 				log_message(LOG_INFO, "VRRP_Script(%s) timed out", vscript->sname);
 			vscript->result = 0;
 		}
-		kill(pid, SIGTERM);
+		kill(-pid, SIGTERM);
 		thread_add_child(thread->master, vrrp_script_child_timeout_thread,
```
pandax381: Yes, We do not want immediate child process terminate ahead.

This is one of the solutions. 

- Immediate child process to handle the SIGTERM.
- And send it to the PID rather than PGID.
- Immediate child process re-sent a signal to the process group
- Finaly, SIGKILL send to the process group

I also looks like there is a risk of race conditions around SIGCHLD.
Also, it is trivial, but I think also should not do a lot of things in the signal handler.
In report_child_status(), it looks like has been calls is not asynchronous-signal-safety function.

I think that it is best to run in a loop of thread_fetch.
Let me think carefully.

keepalived/check/check_daemon.c
pqarmitage: Agreed.

pandax381:
Thank you for your review and comments.

I'll enhance the code.

pandax381:
Hi,

I was fixed issue that has been pointed out to you with the exception of this one.

https://github.com/acassen/keepalived/pull/432#discussion_r80675305

Regards,

PR #437, opened 2016-09-30
Implement not accepting packets to VIPs for VRRPv2

pqarmitage:
Alexandre (@acassen),

Could you have a look at this and make sure you are happy with this. It alters the default behaviour of keepalived so that packets sent to VIPs are discarded (which is what the RFC says should happen, and is already implemented for VRRPv3).

### DIFF
--- a/doc/keepalived.conf.SYNOPSIS
+++ b/doc/keepalived.conf.SYNOPSIS
@@ -276,6 +276,9 @@ vrrp_instance <STRING> {		# VRRP instance declaration
 					#  when using mixed IPv4&IPv6 conf
     state MASTER|BACKUP			# Start-up default state
     interface <STRING>			# Binding interface
+    accept				# Allow a non address-owner to process packets
+					# destined to VIPs and eVIPs. For VRRPv3, but
+					# also VRRPv2 if not strict mode.
     track_interface {			# Interfaces state we monitor
       <STRING>
       <STRING>
--- a/doc/man/man5/keepalived.conf.5
+++ b/doc/man/man5/keepalived.conf.5
@@ -460,7 +460,11 @@ which will transition together on any state change.
         to 192.168.2.0/24 table 1
     }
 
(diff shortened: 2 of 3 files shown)


acassen:
Hi Quentin,

sure if not in MASTER state having netfilter rules discarding traffic to instance (e)vips is good and aligned with RFC.

pqarmitage:
Alexandre,

After I'd submitted the pull request I realised that there would be a problem with IPVS, since real servers need to be able to receive traffic directed to the virtual address, and that is why I closed the pull request.

My reading of RFC3768 (and it is the same in RFC2338), and section 6.4.3, is that when MASTER, _- MUST NOT accept packets addressed to the IP address(es) associated with the virtual router if it is not the IP address owner._

My thought had been to add to the code so that if a VIP/eVIP is being used as the address of a real server, then the traffic to the address won't be blocked, but otherwise the traffic will be blocked (in accordance with the RFC) unless accept mode is set (which is an enhancement on stardard VRRPv2 functionality, but is in accordance with VRRPv3).

Does this seem reasonable to you?

Quentin

acassen:
sure for IPVS, haproxy and mainly any setup running services on top of VRRP VIPs need to be able to handle traffic. For security considerations, we can also imagine that if MAC address is getting stuck in some cam tables in upper side router, or an attaquer gain access or way to inject into router CAM table any MAC address, then it could potentially send traffic to VRRP VIP especially if those VIP are running on dummy interfaces...

Well, I think we can leave the code the way it is right now (I mean no changes or just keyword to activate this feature, but leaving it working the way it is running right now by default).

reaching security concerns, last time I read somewhere (sorry I just dont reming exactly where) people exposing theories about securing VRRP. IMHO, one of the best practice would be encourage people to prefer the use of e-vip since it offers the confort to hide virtual services ! especially if users are running VRRP in unicast fashion... in unicast using IPSEC-AH would also MUST be considered instead of no-auth or password to fight against injection and replay.

PR #440, opened 2016-10-04
Add no_accept mode for VRRPv2 and standardise VRRPv3 with it

pqarmitage:
Alexandre, @acassen 

This is an update of my previous pull request for making VRRPv2 not accept packets sent to the VIP/eVIPs unless the vrrp instance is the address owner. The default now, for the reason you pointed out that IPVS and other services running on the VIPs need accept mode as default, is that accept is the default unless strict mode is specified for an instance. Accordingly a `no_accept` keyword has been added to allow vrrp instances to behave in accordance with the RFCs.

I have changed the mode of operation for VRRPv3 so that it is now the same as VRRP2; in other words anyone who was using VRRPv3 and wanted packets addressed to the VIPs to be discarded will now have to add `no_accept` to their configuration.

If you're happy with the above approach, could you please merge it.


acassen:
great thanks.

PR #447, opened 2016-10-26
Fix up the code to compile without lvs

andempsey:
A few areas were accessing lvs members of the global_data that had been
#ifdefed out. Corrected this. (This is a quick fix up and I'm not so
sure if I've made sensible changes, in particular in core/snmp.c).
Updated the definitions of the KEEPALIVED_OID, and SNMPTRAP_OID. These
were causing some errors in compiling.
Reintroduced VRRP_OID it'd been copied to vrrp_snmp.c by accident.

Signed-off-by: Anthony Dempsey adempsey@brocade.com

### DIFF
--- a/keepalived/core/global_data.c
+++ b/keepalived/core/global_data.c
@@ -163,7 +163,9 @@ alloc_global_data(void)
 		new->enable_snmp_checker = true;
 #endif
 	}
+#ifdef _WITH_LVS_
 	new->lvs_syncd.syncid = PARAMETER_UNSET;
+#endif
 #ifdef _HAVE_IPVS_SYNCD_ATTRIBUTES_
 	new->lvs_syncd.mcast_group.ss_family = AF_UNSPEC;
 #endif
--- a/keepalived/core/snmp.c
+++ b/keepalived/core/snmp.c
@@ -151,7 +151,11 @@ snmp_scalar(struct variable *vp, oid *name, size_t *length,
 		long_ret = global_data->linkbeat_use_polling?2:1;
 		return (u_char *)&long_ret;
 	case SNMP_LVSFLUSH:
+#ifdef _WITH_LVS_
 		long_ret = global_data->lvs_flush?1:2;
+#else
+                long_ret = 2;
+#endif
 		return (u_char *)&long_ret;
 	case SNMP_IPVS_64BIT_STATS:
 #ifdef _WITH_LVS_64BIT_STATS_
--- a/keepalived/include/snmp.h
+++ b/keepalived/include/snmp.h
@@ -46,9 +46,10 @@ int header_generic(struct variable *, oid *, size_t *, int,
 
 #define SNMP_DEFAULT_NETWORK_SOCKET	"udp:localhost:705"
 
-#define KEEPALIVED_OID 1, 3, 6, 1, 4, 1, 9586, 100, 5
-#define SNMPTRAP_OID 1, 3, 6, 1, 6, 3, 1, 1, 4, 1, 0
+#define KEEPALIVED_OID {1, 3, 6, 1, 4, 1, 9586, 100, 5}
+#define SNMPTRAP_OID {1, 3, 6, 1, 6, 3, 1, 1, 4, 1, 0}
 #define GLOBAL_OID {KEEPALIVED_OID, 1}
+#define VRRP_OID {KEEPALIVED_OID, 2}
 
 typedef union _long_ret {
 	unsigned long u;
--- a/keepalived/vrrp/vrrp_snmp.c
+++ b/keepalived/vrrp/vrrp_snmp.c
@@ -107,7 +107,6 @@
 #ifdef _WITH_SNMP_KEEPALIVED_
 /* VRRP SNMP defines */
 enum snmp_vrrp_magic {
-	VRRP_OID KEEPALIVED_OID = 2,
 	VRRP_SNMP_SCRIPT_NAME,
 	VRRP_SNMP_SCRIPT_COMMAND,
 	VRRP_SNMP_SCRIPT_INTERVAL,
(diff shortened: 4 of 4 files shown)


andempsey:
@pqarmitage a few changes here to make --disable-lvs work for me, I'd like your opinion on the core/snmp.c change to make sure it make sense.

Cheers,
Anthony

pqarmitage:
Anthony,

Apologies for the delay in responding.

Many thanks for finding these issues; they only seem to occur when compiling with options --disable-lvs --enable-snmp-keepalived, although there is also a bug when compiled without --enable-snmp-keepalived regardless of the --disable-lvs option.

I've made some comments against your individual patches, and attached to this update is a file containing a revised patch set.

If you could try this patch set and let me know if it works for you. If so, either I can submit it, or it might be more appropriate for you to submit it, and I'll ack it.

[snmp_no_lvs.patch.txt](https://github.com/acassen/keepalived/files/564373/snmp_no_lvs.patch.txt)

keepalived/core/global_data.c
```diff
@@ -163,7 +163,9 @@ alloc_global_data(void)
 		new->enable_snmp_checker = true;
 #endif
 	}
+#ifdef _WITH_LVS_
```
pqarmitage: This is inside a conditional compilation block `#ifdef _WITH_SNMP_`, but it is needed regardless of whether it is being build with SNMP. In my updated patch, I have moved it outside the _WITH_SNMP_ block.

keepalived/core/snmp.c
```diff
@@ -151,7 +151,11 @@ snmp_scalar(struct variable *vp, oid *name, size_t *length,
 		long_ret = global_data->linkbeat_use_polling?2:1;
 		return (u_char *)&long_ret;
 	case SNMP_LVSFLUSH:
+#ifdef _WITH_LVS_
```
pqarmitage: The general approach with keepalived appears to be that if a feature isn't enabled, then no response is given for OIDs that aren't supported. That means the block should look more like:
```
#ifdef _WITH_LVS_
      case SNMP_LVSFLUSH:
      .
      .
      return ...
#endif
```
My updated patch incorporates this change.

keepalived/core/snmp.c
```diff
@@ -151,7 +151,11 @@ snmp_scalar(struct variable *vp, oid *name, size_t *length,
...
 		long_ret = global_data->lvs_flush?1:2;
+#else
```
pqarmitage: See above re no SNMP response if option not compiled in

keepalived/include/snmp.h
```diff
@@ -46,9 +46,10 @@ int header_generic(struct variable *, oid *, size_t *, int,
 
 #define SNMP_DEFAULT_NETWORK_SOCKET	"udp:localhost:705"
 
-#define KEEPALIVED_OID 1, 3, 6, 1, 4, 1, 9586, 100, 5
-#define SNMPTRAP_OID 1, 3, 6, 1, 6, 3, 1, 1, 4, 1, 0
+#define KEEPALIVED_OID {1, 3, 6, 1, 4, 1, 9586, 100, 5}
```
pqarmitage: In principle I'm not happy with this, because it result in a definition of `GLOBAL_OID` as `{{1, 3, 6, 1, 4, 1, 9586, 100, 5}, 1}` which doesn't seem right.

Could you provide details of the compiler error you are getting, and the version of the compiler, so I can see if I can reproduce it.

I haven't included this change in my updated patch.

keepalived/include/snmp.h
```diff
@@ -46,9 +46,10 @@ int header_generic(struct variable *, oid *, size_t *, int,
...
+#define SNMPTRAP_OID {1, 3, 6, 1, 6, 3, 1, 1, 4, 1, 0}
 #define GLOBAL_OID {KEEPALIVED_OID, 1}
+#define VRRP_OID {KEEPALIVED_OID, 2}
```
pqarmitage: This shouln't be here. VRRP_OID disappeared to a bad edit on my part - see below.

keepalived/vrrp/vrrp_snmp.c
```diff
@@ -107,7 +107,6 @@
 #ifdef _WITH_SNMP_KEEPALIVED_
 /* VRRP SNMP defines */
 enum snmp_vrrp_magic {
-	VRRP_OID KEEPALIVED_OID = 2,
```
pqarmitage: This is clearly nonsense, and was a bad edit by me in commit c4510e60.

My patch reinstates the definition of VRRP_OID, keeping it in this file.

keepalived/vrrp/vrrp_snmp.c
```diff
@@ -1821,13 +1820,18 @@ vrrp_snmp_instance(struct variable *vp, oid *name, size_t *length,
 		return (u_char *)&long_ret;
 
 	case VRRP_SNMP_INSTANCE_USELVSSYNCDAEMON:
+                long_ret.u = 0;
+#ifdef _WITH_LVS_
 		long_ret.u = (global_data->lvs_syncd.vrrp == rt)?1:2;
+#endif
 		return (u_char *)&long_ret;
 	case VRRP_SNMP_INSTANCE_LVSSYNCINTERFACE:
+#ifdef _WITH_LVS_
```
pqarmitage: See earlier comments re how conditional compilation is normally handled in keepalived for SNMP objects not defined.

pqarmitage:
It appears that one small part was missing from my previous patch. Attached is an updated version.

The only aspect this doesn't deal with is the `{` and `}` around the OID definitions, which I will have a look at further once you can let me know the compiler details and error messages.

andempsey:
Hey Quentin,
The {} problems has gone away now, don't know why it was being a problem.
The rest of your changes make sense and work for me. I'm happy to close this off and use your patch if you want.

Cheers,
Anthony

pqarmitage:
Anthony,

Yes, since you are happy with the changes, lets go with what I submitted, and it would be good if you could close this off.

Many thanks,
Quentin

PR #448, opened 2016-11-02
Resolve segfaults during configuration parsing and add 4 uses of defines

pqarmitage:
The main commit is designed to stop keepalived segfaulting while parsing the configuration file when a keyword requires a parameter but the parameter is missing.

If the parameter should have been the last word on a configuration line, keepalived would generally segfault if the parameter was missing. This commit instead reports an error to the log, and terminates gracefully.

The extra #define usage is for TIMER_HZ and FMT_STR_VLOT.

Merge a couple of pull requests the either needed conflict resolution or minor tweaking.


rohara:
This looks fine. My only question is why do we need the register/unregister functions? Do you have a use case in mind where we would register a different handler? Thanks for submitting this.

pqarmitage:
The reason for the register function is that if, for example,  genhash wanted to use strvec_slot, it may do so for a different purpose, and so the error messages would need to be different. This could even apply somewhere else in keepalived at some time in the future.

Having implemented the register function, it seemed appropriate to add the unregister function, in order to restore the previous state, and then I though it would be worth returning the previous address, so that a function could temporarily modify the error handling function, and then return it to the previous setting.

PR #468, opened 2016-11-26
Init_fail setting for vrrp_script

rolffokkens:
For some scripts (like checking if a daemon is running) I want keepalived not to jump to fast conclusions wether the script succeeded or failed after startup (which it normally does). For these scripts I want keepalived to just assume the script is failed initially and use the rise option to establish that the script succeeded.

This init_fail option allows to specify this behaviour.


rolffokkens:
The same reasoning may apply for an init_success option, but I don't see a clear case for that - what does it mean to initially make it hard for a script that it's failing?

If an init_success option seems useful though,  a better approach might be an "init_state" option with an argument "failed" or "success".

pqarmitage:
Rolf,

I would rather not merge this, since it will shortly superseded by a series of patches that I am currently finalising, and it would seem a shame to introduce a keyword that then becomes redundant.

The changes I am working on speed up transition to master, and also handles interface state changes and script state changes when they happen rather than waiting for the next receive timeout or receive event. As a consequence of these changes scripts start in the fault state and transition to succeed state after one successful return by the script, which is exactly what you want, and also what someone else asked for a few weeks ago.

I think probably the best solution is for you to work with your patch until what I believe will be v2.0 of keepalived is released. Are you happy with that?

If you would like to see the version of keepalived that I am working on, it is at github.com/pqarmitage/keepalived/ branch fixes.quick.

rolffokkens:
Thanks for your response Quentin. There may be some misunderstanding. In the current implementation vrrp_init_script () actually transition to succeed happens  when "one success is enough" after startup (result == VRRP_SCRIPT_STATUS_INIT). But for some scripts I consider this to be too fast, I want to delay this transition to succeed until rise attempts have proven the script to be "reliable" after startup of keepalived. So what I want is a kind of the opposite of "scripts start in the fault state and transition to succeed state after one successful return by the script".

I will look at the version you're working on though, and I won't mind sending a pull request later when still appropriate. No problem for me to keep using my patch until then.

pqarmitage:
Rolf,

I should have read the patch properly! I now see that what you are doing is more than is incorporated in the changes I have made, so I'll merge it.

rolffokkens:
Thx!

PR #478, opened 2016-12-08
When VRRP_Instance is MASTE state ,if vip was removed, it will be bou…

ZhaoJing1:
…nd to linux kernel interface again

### DIFF
--- a/keepalived/vrrp/vrrp.c
+++ b/keepalived/vrrp/vrrp.c
@@ -69,6 +69,8 @@
 #include <netinet/ip.h>
 #include <netinet/ip6.h>
 
+extern int vip_netlink_address_lookup(vrrp_t * vrrp);
+
 /* add/remove Virtual IP addresses */
 static bool
 vrrp_handle_ipaddress(vrrp_t * vrrp, int cmd, int type)
@@ -1555,6 +1557,9 @@ vrrp_state_master_tx(vrrp_t * vrrp, const int prio)
 	vrrp_send_adv(vrrp,
 		      (prio == VRRP_PRIO_OWNER) ? VRRP_PRIO_OWNER :
 						  vrrp->effective_priority);
+						  
+        /*When VRRP_Instance is MASTE state ,if vip was removed, it will be bound to linux kernel interface again.*/						  
+	vip_netlink_address_lookup(vrrp);						  
 	return ret;
 }
 
--- a/keepalived/vrrp/vrrp_netlink.c
+++ b/keepalived/vrrp/vrrp_netlink.c
@@ -1084,3 +1084,237 @@ kernel_netlink_close(void)
 	netlink_close(&nl_kernel);
 	netlink_close(&nl_cmd);
 }
+
+
+static int
+vip_netlink_if_address_filter(__attribute__((unused))struct sockaddr_nl *snl, struct nlmsghdr *h,ip_address_t *ipaddress,int *vip_flag)
+{
+	struct ifaddrmsg *ifa;
+	struct rtattr *tb[IFA_MAX + 1];
+	int len;
+    interface_t *ifp;
+	void *addr;
+	void *vip_addr;
+        int vip_set =0;
+
+	ifa = NLMSG_DATA(h);
+
+	/* Only IPV4 are valid us */
+	if (ifa->ifa_family != AF_INET && ifa->ifa_family != AF_INET6)
+		return -1;
+
+	len = h->nlmsg_len - NLMSG_LENGTH(sizeof (struct ifaddrmsg));
+	if (len < 0)
+		return -1;
+
+	memset(tb, 0, sizeof (tb));
+	parse_rtattr(tb, IFA_MAX, IFA_RTA(ifa), len);
+
+	/* Fetch interface_t */
+	ifp = if_get_by_ifindex(ifa->ifa_index);
+
+	if (!ifp)
+		return 0;
+	if (tb[IFA_LOCAL] == NULL)
+    {
+        tb[IFA_LOCAL] = tb[IFA_ADDRESS];    
+    }
+	if (tb[IFA_ADDRESS] == NULL)
+    {
+        tb[IFA_ADDRESS] = tb[IFA_LOCAL];
+    }
+
+
+	/* local interface address */
+	addr = (tb[IFA_LOCAL] ? RTA_DATA(tb[IFA_LOCAL]) : NULL);
+
+	if (addr == NULL)
+    {
+        return -1;
+    }
+
+    if(ifa->ifa_family == AF_INET)
+    {
+        vip_addr = (void *)&ipaddress->u.sin.sin_addr;
+    }
+    else
+    {
+        vip_addr = (void *)&ipaddress->u.sin6_addr;    
+    }
+	
+
+	/* If no address is set on interface then set the first time */
+	if ((ifa->ifa_family == ipaddress->ifa.ifa_family)&& (ifa->ifa_index == ipaddress->ifa.ifa_index ))
+    {
+    
+        if( inaddr_equal(ifa->ifa_family,vip_addr,addr))
+        {
+            vip_set = 1; 
+        }
+        else
+        {
+            vip_set = 2; /*标示这次未找到vip地址*/
+        }
+        * vip_flag = vip_set;		
+    }
+	return 0;
+}
+
+
+static int
+vip_netlink_parse_info(int (*filter) (struct sockaddr_nl *, struct nlmsghdr *,ip_address_t *,int *),
+		   nl_handle_t *nl, list vip_list)
+{
+	int status;
+	int status_bak;
+	int ret = 0;
+	int error;
+	ip_address_t *ipaddr;
+	element e;
+        int vip_flag = 0;
+	
+	if (LIST_ISEMPTY(vip_list))
+		return -1;
+
+	while (1) {
+		char buf[4096];
+		struct iovec iov = { buf, sizeof buf };
+		struct sockaddr_nl snl;
+		struct msghdr msg =
+		    { (void *) &snl, sizeof snl, &iov, 1, NULL, 0, 0 };
+		struct nlmsghdr *h;
+
+		status = recvmsg(nl->fd, &msg, 0);
+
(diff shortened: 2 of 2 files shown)


ZhaoJing1:
When VRRP_Instance is MASTE state ,if vip was removed, it will be bound to linux kernel interface again

pqarmitage:
Could you explain the circumstances that cause this patch to be needed; what are the circumstances under which a vip address gets removed?

ZhaoJing1:
When Exception in deleting vip by "ip -f inet addr delete "cmmand or restart network service, there was a error :
vip was removed, but VRRP_Instance is MASTER STATE and is sendding VRRP advertisement.

ZhaoJing1:
When we use keepalived, deleting vip by "ip -f inet addr delete "command or restart network service, sometimes there has been an error:
vip was removed, but VRRP_Instance is MASTER STATE and is sendding VRRP advertisement.

pqarmitage:
I think the first thing that needs to be resolved is what is the correct course of action to take if a vip address is removed. In the first case you quote, where someone (or some script) removes an IP address, then it must be assumed that that was done for a good reason, and so it would be wrong to simply add the address back. In the second case, of a network restart, any VMAC interfaces would presumably need to be recreated too.

My view is that if a VIP address is removed, then the VRRP instance should go into backup state, and if it subsequently returns to master state then the VIP address will be added back. Likewise, if there is a network restart, then the safest thing to do is to make all the instances fall back to backup state and let keepalived recover in the normal way.

I'm not sure I understand the logic of the design of the patch. Every time an advert is sent, it opens a netlink socket and retrieves all the IPv4 and IPv6 addresses on the system, to check if the addresses belonging to the VRRP instance for which the advert is being sent are still configured. If there is a large number of VRRP instances, or the advert interval is low (it can be as low as 10ms), this will cause a significant overhead on the system. Keepalived already receives notification when an address is removed (this is handled in netlink_if_address_filter()). It would be far more efficient to check in that function if an address that is reported as being removed is a VIP address, and in which case take the appropriate action.

There appear to be a couple of problems in vip_netlink_parse_info():
1. The outer loop starting at line 1179 reads a block of messages, and then checks all addresses to see if they are in the block, and if not, then adds the address. The problem is that if an address is not in the first block of messages, the function will attempt to add that address, even though information about that address might be in a subsequent block of messages. Indeed, if there are multiple blocks of messages (i.e. multiple successful calls of recvmsg() at line 1187), then for each block except the block that contains the address being checked, where will be an attempt to add the address.
2. In the for loop starting at line 1210, if the last message is read (nlmsg_type == NLMSG_DONE), the function returns. It will therefore not add a missing address (unless it happened to be added due to the problem at 1 above).

In function vip_netlink_if_address_filter() it uses the address family of the address returned by netlink to determine the type of address in the vip list (line 1136). This is actually handled by line 1147, but the sequence of the code is wrong.

In vip_netlink_address_lookup(), it should only call netlink_request() and vip_netlink_parse_info() for the appropriate address family (all addresses in vrrp->vip are of the same address family).
Also, this function should be handling the vrrp->evip list, which can contain a mix of IPv4 and IPv6 addresses.

There are lots of problems with the alignment of the code (it looks as though you are using 4 spaces rather than a tab for indentation).

What is the comment at line 1156?

In summary, I agree there is an issue if an address is removed, but I think the correct action for the circumstances you give is to go to backup state. This will then mean that the other issues above are superseded. What I do think needs handling, though, is if an interface disappears and is subsequently recreated (e.g. a hot swap device is removed and reinstated); in this case any VMACs would need to be recreated.

acassen:
what happen if you "ip addr del" the IP where your apache web server is listening on ? just nothing... you dont even have a short log about it.

IMHO, having just a short log saying kind of "hey man, something else than VRRP is deleting/managing VRRP VIP %u.%u.%u.%u, WTF?!" sound enough ?

We can find other strategies like transiting to FAULT state (maybe better than BACKUP ?), forcing adding/deleting back VIP when caught via netlink reflection channel, ...

Personally having just the short log is enough since operating a box you MUST know behaviour of what is running on, and adding logic that think on behalf of admin box is certainly not the best way. We should notify to admin that something weird is happing then admin MUST make his job by troubleshooting and fixing, since this is an issue.

Maybe there is other use-cases we are simply missing ?

ZhaoJing1:
I agree that the code modified by me was flawed and the strategy for the bug was not better .
     If vip deleted by "ip -f inet addr delete "command or restart network service, no event triggered netlink reflector to reports the vip removed to VRRP_Instance, and VRRP_Instance did not that the vip was removed.
      how do VRRP_Instance know this VIP was removed wiithout checking vip  periodically? Now,It is  important how to know VIP is deleted for VRRP instance. So, I think it is reasonable to Check if VIP is removed periodically for VRRP. I agree that the method of checking VIP provided by me was not very good.
    I think it is reasonable that VRRP instance should go into backup or FAULT, also.

flerxu:
In another scenario，if vip is missing on MASTER node weirdly（not being manually removed）, what should keepalived do?

pqarmitage:
The challenge is knowing what is the right thing to do if the address has "disappeared".

Perhaps the most common circumstance for this to happen is if the interface that the address has been configured on goes down, since in that circumstance all addresses configured on the interface are automatically removed. That is straightforward if the interface is also the interface that the vrrp instance is running over, since the vrrp instance will transition into fault state. However, if the address is configured on an interface that isn't the same as the one that the vrrp instance is configured on, then at the moment by default nothing will happen. This is actually an easy one to solve since the vrrp instance can be configured to track the interface without a weight, in which case if the interface goes down the vrrp instance will go to fault state. We could add a keyword `track_vip_ifs` for example, which would make the vrrp instance track the interfaces that vips are configured on (we probably ought to have a `track_vroute_ifs` since routes will be removed too if they are configured to use specific interfaces [note that ip rules are not removed if relevant interfaces go down or are deleted]).

The issue is what if the ip address (or route) is removed by some other process, perhaps legitimately. Should we add it back, should we transition to backup (or fault) state, should we do nothing, or is there some other option? If we add the address back, we might end up with the process that removed the address removing it again and we just end up in a loop with the address being added and deleted (and why was the address removed in the first place?). The problem with transitioning to fault state is what event would trigger returning from fault state. Transitioning to backup seems a "nice" thing to do. If we become master again, the address will be added again, but there will be a time delay before that happens (minimum 30ms depending on the advert interval). This solution would work especially well if nopreempt is configured. What other option is there? There could be notify_vip_deleted and notify_route_deleted notify scripts, but other than that I can't think what else could be done.

What about static_ipaddresses and static_routes? They may be essential to the operation of a virtual router.

So here's a proposal:

1. Add keywords `track_vip_ifs` and `track_vroute_ifs` as described above
2. Add keyword `track` against vips and vroutes so that if a vip or vroute is deleted, the instance will transition to backup state
3. Add keywords `track_vips` and `track_vroutes` to track all a vrrp instances vips or vroutes (same effect as configuring `track` against all vips or vroutes). We could also add these as global keywords too to make it the default for all vrrp instances.
4. Modify keywords `static_ipaddress` and `static_routes` to add an optional track name, for example:
`static_ipaddress track_group1 {`. Then add keywords `track_sip` and `track_sroute` for vrrp instances, so that the static ip addresses and static routes for the vrrp instance to track can be specified. Again, `track_sips` and `track_sroutes` could be global keywords to make all instances track all static ips and static routes.
5. If a static ip address or static route is deleted due to an interface going down, then when the interface returns, re-add the static ip address/routes.
6. Add global keyword `track_ip_route` which is the same as specifing `track_vips`, `track_vroutes`, `track_sips` and `track_sroutes` at the global level.
7. Add keyword `restore` for static ip addresses and static routes so that if they are deleted, we automatically add them back.
8. Add `notify_vip_deleted` and `notify_vroute_deleted` notify scripts. Can we also add `notify_sip_deleted` and `notify_sroute_deleted` scritps?
9. If a vmac interface is removed, transition to backup state (or fault state if the underlying interface is down) and re-add the vmac interface immediately if possible, or when the underlying interface comes up again (the vmac interfaces are ours, so no-one else should delete them).

The above has the advantage that the current default behaviour is not altered (other than reinstating vmacs and static ip address and routes after an interface is restored), but also allows keepalived to be configured to take whatever action is needed in the event of address or route deletion. The default behaviour can be changed by configuring `track_vips`, `track_vroutes`, `track_sips` and `track_sroutes` at the global level.

The above changes would also make keepalived fully functional with hot-swap NICs.

Do you have any thoughts on the above? Would it meet your needs?

This is something that I had planned in principle to implement, but I hadn't previously worked it out to the level of details specified above.

pqarmitage:
@flerxu Version 2.0.0 of keepalived was released yesterday. It includes support for monitoring virtual ip addresses, and if one is deleted the vrrp instance will transition to backup. It it becomes master again, and deleted VIPs will be re-added.

I'm not closing this pull request yet due to my comment https://github.com/acassen/keepalived/pull/478#issuecomment-284150985 which sets out handling for ip routes and ip rules if they are deleted too. I am currently implementing support for this, although slightly differently from the details set out in my earlier comment.

pqarmitage:
@ghost, @flerxu I have now implemented my planned changes for handling the deletion of VIPs, eVIPs, virtual routes and virtual rules. In addition I have implemented restoring static ip addresses, routes and rules when they are deleted.

I have added a no_track option to each of static and virtual ip addresses, routes and rules. Unless no_track is specified against an address/route/rule, action will be taken to restore it if it deleted. If it is a virtual address/route/rule, the vrrp instance will drop back to backup, and if it then subsequently will re-add any virtual addresses/routes/rules when it next transitions to master; if it cannot re-add the item (this will occur if the interface of an address or route is down), then the vrrp instance will go to fault state until the interface is restored. If it is a static address/route/rule keepalived will attempt to re-add the deleted item.

For static address/routes/rules there is an additional option - named track_group entries. A track group has a list of vrrp instances. If a static address/route(/rule) cannot be added (due to an interface being down), those vrrp instances will be put into fault state until the interface is restored.

In summary, by default any address, route or rule will be re-added if it is deleted, unless the address/route/rule is configured with the no_track option. In addition, for any static address/route(/rule) vrrp instances can be configured to go to fault state if the address/route/rule cannot be configured.

Issue #484, opened 2016-12-17
Vrrp script weight default value is 0 but documentation defines it as 2


joncastro:
The [keepalived.conf.SYNOPSIS](https://github.com/acassen/keepalived/blob/master/doc/keepalived.conf.SYNOPSIS) file defines vrrp script **weight** default value as 2 when **weight** is not provided.

```
If unspecified, the weight equals 2, which means that a success will add +2
to the priority of all VRRP instances which monitor it. On the opposite, a
negative weight will be subtracted from the initial priority in case of
failure.
```

But the file [vrrp_track.h](https://github.com/acassen/keepalived/blob/master/keepalived/include/vrrp_track.h) sets the constant `VRRP_SCRIPT_DW` to `0` instead of `2`.

```
#define VRRP_SCRIPT_DW 0	/* external script default weight */
```

The fix only requires to either `VRRP_SCRIPT_DW` to `2` or update the documentation.

This behavior can be easily reproducible with a simple configuration like following which weight is not defined.

```
vrrp_script chk_sshd {
       script "killall -0 sshd"
       interval 1
}

vrrp_instance VI_1 {
        interface eth0
        state MASTER
        virtual_router_id 51
        priority 100
        track_script {
          chk_sshd
        }
}
```

joncastro:
https://github.com/acassen/keepalived/pull/485

joncastro:
The recommended fix is to update the documentation because changing the code will produce change current deployments. Weight `0` is a special case which the vrrp goes to FAULT state in case of script failure. A weight different from 0 will increase/decrease the priority.

joncastro:
Pull request: https://github.com/acassen/keepalived/pull/487

joncastro:
documentation fixed and merged

PR #485, opened 2016-12-17
Issue: #484 Set vrrp script default value to 2 as defined by the documentation in…

joncastro:
… keepalived.conf.SYNOPSIS

### DIFF
--- a/keepalived/include/vrrp_track.h
+++ b/keepalived/include/vrrp_track.h
@@ -43,7 +43,7 @@
 /* VRRP script tracking defaults */
 #define VRRP_SCRIPT_DI 1	/* external script track interval (in sec) */
 #define VRRP_SCRIPT_DT 0	/* external script track timeout (in sec) */
-#define VRRP_SCRIPT_DW 0	/* external script default weight */
+#define VRRP_SCRIPT_DW 2	/* external script default weight */
 
 /* VRRP script tracking results.
  * The result is an integer between 0 and rise-1 to indicate a DOWN state,


pqarmitage:
I think the difficulty with this approach is that it is changing the default behaviour of keepalived, and so this patch could break some existing configurations.

Probably what should be done is the documentation should be updated to match the way keepalived currently works. Also doc/man/man5/keepalived.conf.5 needs to be changed to keep it in step with keepalived.SYNOPSIS.

joncastro:
Thanks @pqarmitage  for your comment.

I agree that changing the documentation might be a valid solution in most cases but in this particular case keeping `weight` as default value `0` does not seem the right approach because defining track script with `weight` `0` does not influence in the priority and consequently it is like not defining the track script.

The user would expect that if a track script is defined it will influence the priority in order to keep MASTER the one that contains the most stable state of the system.

For example, if we have a two nodes running a webserver like `nginx` and a track script like:
```
vrrp_script chk_nginx {
       script "killall -0 nginx"
       interval 1
}
vrrp_instance VI_1 {
        ...
        track_script {
          chk_nginx
        }
        ...
}
```

then the user would expect that by default the master node will contain the node that runs nginx, but if `weight` is defined to `0` then the track script has not influence and MASTER node will be any node independently if nginx is running or not.

If existing deployment contains a configuration with no `weight` it will fix the issue keeping MASTER node the one that contains highest priority based on defined track scripts, aligning the deployment with the documentation and expected behavior.

Finally, I think setting `weight` to `0` does not seems valid in most cases and if the user really wants setting `weight` to `0` then this value should be manually provided via `weight 0` configuration line instead of being the default keepalived behavior.

pqarmitage:
Please see the last 2 paragraphs of section 2 of keepalived.SYNOPSIS (which contradicts what is stated at 2.1). A weight of 0 (which is the same as not specifying the weight) is a special case (since as you say it would otherwise do nothing itself), and it causes the vrrp instance to go to fault state if the script returns a non-zero exit code (the same also applies for track_interface).

I certainly have configurations that use this feature, and I have no doubt that others do too.

joncastro:
In that case, I agree. It would be better to update the documentation. I will close the pull request.

PR #487, opened 2016-12-19
fix doc: weight default is 0, which means tat a failure implies a FAU…

joncastro:
…LT state.


joncastro:
Issue: https://github.com/acassen/keepalived/issues/484

Issue #490, opened 2017-01-06
sorry server on reload


ze42:
When a service is down, and a `sorry_server` is used, `keepalived` trash it on reload, requiring the normal service back before being able to get back to use the `sorry_server`.

Wanting to use that service as an active/passive service, it curently kinda fails my pupose.

Tested on current master (cf3b236) and somthething near v1.2.23.

Configuration: (keepalived.conf)

```
global_defs {
   router_id foo.example.com
}
virtual_server_group group01 {
        10.2.0.1 9200
}
virtual_server group group01 {
        delay_loop 5
        lb_algo wrr
        lb_kind DR
        protocol TCP
        real_server 10.1.105.1 3306 {
                weight 1
                HTTP_GET {
                  url {
                    path /
                    status_code 200
                  }
                  connect_port 9200
                  connect_timeout 3
                  nb_get_retry 1
                  delay_before_retry 0
                }
        }
        sorry_server 10.1.105.2 3306
}
```

On load, when everything is fine:

```text ipvsadm --list
IP Virtual Server version 1.2.1 (size=4096)
Prot LocalAddress:Port Scheduler Flags
  -> RemoteAddress:Port           Forward Weight ActiveConn InActConn
TCP  10.2.0.1:9200 wrr
  -> 10.1.105.1:3306              Route   1      0          0
```

After getting RS monitored service to fail.

```
# ipvsadm --list
IP Virtual Server version 1.2.1 (size=4096)
Prot LocalAddress:Port Scheduler Flags
  -> RemoteAddress:Port           Forward Weight ActiveConn InActConn
TCP  10.2.0.1:9200 wrr
  -> 10.1.105.2:mysql             Route   1      0          0         
```

Now that the `sorry_server` is in place, a reload should be able to still keep it, until the RS is back online... but it just removes it, leaving no server at all in the pool.

```
# ipvsadm --list
IP Virtual Server version 1.2.1 (size=4096)
Prot LocalAddress:Port Scheduler Flags
  -> RemoteAddress:Port           Forward Weight ActiveConn InActConn
TCP  10.2.0.1:9200 wrr
```

If server gets back online, then fails again, `sorry_server` gets back properly...

pqarmitage:
Many thanks for your work on this. You commit has now been merged.

PR #491, opened 2017-01-09
sorry_server load/reload on failure

ze42:
Fix Sorry Server handling on start/reload

  * Do not remove it if already in place, and same server required
  * add the sorry server on start in alpha mode (quorum down at start)
  * add the new sorry server on a down VS on reload (quorum down at reload)

See issue #490


ze42:
Should check which fields are really useful in s_srv, when getting fields from existing s_srv.

PR #494, opened 2017-01-11
Fix missing documentation for 4th parameter of notify action

icflournoy:
Spent a good while trying to track down why my script wasn't executing (it was checking for 3 parameters and exiting if it didn't receive them).

The `notify_*` action documentation includes the $4 parameter (priority), but it was missing for the catch-all `notify` action.


DoItMorty:
+1

PR #503, opened 2017-01-23
allow to override build date

bmwiedemann:
to enable reproducible builds.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.

note that the configure part is tested, but the configure.ac part is not
The date call is supposed to work with both GNU date and BSD date.

### DIFF
--- a/configure
+++ b/configure
@@ -8973,7 +8973,6 @@ fi
 
 echo $INIT_TYPE
 VERSION_DATE=`date +%m/%d,%Y`
-VERSION_YEAR=`date +%Y`
 
 
 cat >>confdefs.h <<_ACEOF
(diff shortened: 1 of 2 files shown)


pqarmitage:
I think that having to manually change the copyright year is prone to error, and is likely to be forgotten since it only needs to be done once a year.

In respect of the date emitted with the -v option, if it is changed to be the date of the last git commit, it will have the effect of producing repeatable builds.

Commit 5287f03 makes the changes as outlined above, which appears to meet your needs of not having a future copyright date, and builds being reproducable, without the overhead of having to manually change dates, or forcing fake build dates.

If the above does not meet your needs, then please either submit a further patch, or raise an issue.

Many thanks for your suggestions and updates.

bmwiedemann:
many thanks for the update. If it also works with tarballs, it will be fine.

pqarmitage:
There is a file lib/git-commit.h which was already used to hold the latest commit id, and it is this file that now has the date and year. The file is included in tarballs.

PR #504, opened 2017-01-25
Make build repeatable

pqarmitage:
USe date/year of latest git commit for copyright and date output with -v option.


pqarmitage:
Not sure what happened here. Wrong branch pushed.

PR #509, opened 2017-02-01
fixes to allow building under alpine linux

### LINKED ISSUE #508: Issues building/running under alpine linux
The current build/code has a couple of issues when attempting to build and run under alpine linux (which uses the musl c library):

- minor issue: configure.ac doesn't handle an undetected init system (alpine uses openrc)
- major issue: musl doesn't really support/have a sane value for GETPW_R_SIZE_MAX/GETGR_R_SIZE_MAX which causes memory corruption when using script user/group, usually leading to a crash on startup

r0p0s3c:
Fixes #508

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1052,7 +1052,7 @@ INIT_TYPE=
 if test $? -eq 0; then
         INIT_TYPE=upstart
 else
-	if test \( `which systemctl` -a -x `which systemctl` \); then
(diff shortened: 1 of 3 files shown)


pqarmitage:
The changes have been implemented in commits 849803b and a1e142b. The check for sysconf() returning -1 is now done at runtime.

PR #523, opened 2017-02-15
Priority owner case is not working along with preemption

Navjot01:
priority should be given to priority owner case not to the pre-emption
 mode(As per the RFC).

 Check for preemption only if its not the case of priority owner.

In case one router having preemption enabled and it is acting as backup, but if we configure the same router as priority owner router in that case this should become master irrespective of the preempt state.


pqarmitage:
Many thanks for the report, and I agree there does appear to be a problem as you describe.

Have you tested your patch? So far as I can see it doesn't even compile, and if the most obvious modification is made so that it compiles, then it appears to me that the patch would stop a backup becoming master unless it was the address owner, or there were no other masters.

Further, the source version you have made the patch against is 5 months old, and some 213 commits behind upstream, so it may well be that the problem has been fixed in the intervening period (although I don't think it has been).

Would you like to fix your patch and test it to ensure that it works, and then resubmit?

pqarmitage:
I've been thinking about this while in the shower, and vrrp_state_backup() is the wrong place to do this.

A check should be made in vrrp_complete_instance() that if priority == VRRP_PRIO_OWNER, then nopreempt is not set, and if it is set, log a warning and clear nopreempt.

Navjot01:
Hi Thanks for the quick reply, my mistake i have updated wrong change, pushed the correct change.

Also synced my git.

IMO, we should do it in vrrp_state_backup() and RFC does not say that its a wrong configuration so why we need to log warning. User can configure no preempt and in case of priority owner state will change to master else if there is change in the IP  and preemption is enabled it will work as it is.

I hope it make sense.

Thanks !

pqarmitage:
The RFC states in section 3:
```
It's also possible to administratively prohibit all preemption attempts. The only exception is that a VRRP router will always become Master of any virtual router associated with addresses it owns.
```
which is a clear statement that preemption will not apply for the address owner.

The reason for logging a warning if nopreempt is set for an address owner is that if you are the address owner nopreempt cannot work, and therefore keepalived cannot work in the way the configuration requests. Further, checking this at startup means the check is made only one, rather than every time an advert is received.

Navjot01:
Changes done in vrrp_complete_instance() for this issue seems good enough to resolve this issue. Also it sounds good to have this change over here rather then in vrrp_state_backup()  as suggested by me.

I have tested with this change and it is working fine. 

Thanks :)

PR #526, opened 2017-02-16
Makefile.in: provide default values for git-commit.h

otheus:
See issue 524. Note: I don't know what to do with Makefile.am . Also note I use the $@ variable. I'm not sure why the existing Makefile doesn't use this more.


pqarmitage:
Makefile.am is the source file, Makefile.in is generated from it, so the patch needs to be made to Makefile.am.

Commit 185912c updates lib/Makefile.am to generate default definitions for GIT_DATE and GIT_YEAR.

PR #540, opened 2017-03-08
Fix build failures with the musl C library

yann-morin-1998:
Hello,

[musl](http://www.musl-libc.org/) is a strictly standards-compliant C library.

It does not define non-standard types (such as the kernel-specific `u_intXX_t`) and its headers do not internally cross-include themselves much.

This two patches fix build issues such as:  
http://autobuild.buildroot.org/results/b61/b61c2cfa949c2a20f0702522857d78651207e6d2/build-end.log

Thanks!

Regards,
Yann E. MORIN.

### DIFF
--- a/keepalived/check/libipvs.c
+++ b/keepalived/check/libipvs.c
@@ -26,6 +26,7 @@
 #include <netinet/in.h>
 #include <sys/socket.h>
 #include <arpa/inet.h>
+#include <stdint.h>
 
 #ifdef LIBIPVS_USE_NL
 #include <netlink/netlink.h>
--- a/keepalived/include/ip_vs.h
+++ b/keepalived/include/ip_vs.h
@@ -36,7 +36,7 @@ typedef struct ip_vs_stats_user ip_vs_stats_t;
 
 struct ip_vs_service_app {
 	struct ip_vs_service_user user;
-	u_int16_t		af;
+	uint16_t		af;
 	union nf_inet_addr	nf_addr;
 #ifdef _HAVE_PE_NAME_
 	char			pe_name[IP_VS_PENAME_MAXLEN];
@@ -45,7 +45,7 @@ struct ip_vs_service_app {
 
 struct ip_vs_dest_app {
 	struct ip_vs_dest_user	user;
-	u_int16_t		af;
+	uint16_t		af;
 	union nf_inet_addr	nf_addr;
 };
 
@@ -53,7 +53,7 @@ struct ip_vs_dest_app {
 struct ip_vs_service_entry_app {
 	struct ip_vs_service_entry user;
 	ip_vs_stats_t		stats;
-	u_int16_t		af;
+	uint16_t		af;
 	union nf_inet_addr	nf_addr;
 #ifdef _HAVE_PE_NAME_
 	char			pe_name[IP_VS_PENAME_MAXLEN];
@@ -64,7 +64,7 @@ struct ip_vs_service_entry_app {
 struct ip_vs_dest_entry_app {
 	struct ip_vs_dest_entry user;
 	ip_vs_stats_t		stats;
-	u_int16_t		af;
+	uint16_t		af;
 	union nf_inet_addr	nf_addr;
 
 };
@@ -84,7 +84,7 @@ struct ip_vs_get_dests_app {
 	struct ip_vs_dest_entry_app	entrytable[0];
 	} user;
 
-	u_int16_t		af;
+	uint16_t		af;
 	union nf_inet_addr	nf_addr;
 };
 
@@ -126,13 +126,13 @@ struct ip_vs_daemon_app {
 	uint16_t		sync_maxlen;
 
 	/* Multicast Port (base) */
-	u_int16_t		mcast_port;
+	uint16_t		mcast_port;
 
 	/* Multicast TTL */
-	u_int8_t		mcast_ttl;
+	uint8_t			mcast_ttl;
 
 	/* Multicast Address Family */
-	u_int16_t		mcast_af;
+	uint16_t		mcast_af;
 
 	/* Multicast Address */
 	union nf_inet_addr	mcast_group;


pqarmitage:
Re commit `check/libipvs: add missing include`, the comment at the beginning states `u_int16_t et al. are from stdint.h, so it needs to be included`. This isn't quite correct, since stdint.h defines `uint16_t` etc, and not `u_int16_t`, and the `u_int16_t` etc types are defined in /usr/include/sys/types.h.

I'm inclined not to include the `check/libipvs: add missing include` patch, since I don't think it fixes anything (and the comment is wrong), but I do agree that the other patch is sensible since it moves towards standardising the integer types we use (check/ipvswrapper.c and check/check_snmp.c should also be changed).

With regard to the `check/libipvs: add missing include` patch, I think the principle is right that we should include `stdint.h` where we are using [u]intnn_t types, but it needs rather more changing than just check/libipvs.c, since a large number of other source files use the types as well.

I note from your build log that you don't have libnl3-dev, iptables-dev and ipset-dev installed. If you build with them installed, and have the runtime versions on your target systems, you will get a better performing and functional keepalived (in fact it is having libnl3-dev installed that stopped me getting the compilation failures with musl that you were experiencing).

I also note that you have a large number of invalid options being passed to configure.

I have now merged your `keepalived/ip_vs: fix integer types` commit and added some other commits to change the remaining u_intnn_t types and to include <stdint.h>.

yann-morin-1998:
@pqarmitage Thanks for the feedback!

> Re commit check/libipvs: add missing include, the comment at the
> beginning states u_int16_t et al. are from stdint.h, so it needs to be
> included. This isn't quite correct, since stdint.h defines uint16_t
> etc, and not u_int16_t, and the u_int16_t etc types are defined in
> /usr/include/sys/types.h.

Indeed, my bad. However, I think we should not mix the types. Either we
use `uintXX_t` or `u_intXX_t`, but we do it consistently across the
code. I see you did just that in recent commits; good, I won't have to
do it (less work for me, yeah!). Thanks! :-)

> I note from your build log that you don't have libnl3-dev,
> iptables-dev and ipset-dev installed. If you build with them
> installed, and have the runtime versions on your target systems, you
> will get a better performing and functional keepalived

Maybe. However, Buildroot is a tool to create embedded systems via
cross-compilation, and packages are optional. Since dependencies on
libnl3, iptables or ipset are all optional, we do not enforce them; it's
up to the user to enable what he wants on his system.

> in fact it is having libnl3-dev installed that stopped me getting the
> compilation failures with musl that you were experiencing

Note that Buildroot is about cross-compilation, so we do not use the
build machine libs or headers (because it does not make sense to use
libs for x86-64 to build for ARM, for example).

> I also note that you have a large number of invalid options being
> passed to configure.

Yes, this has virtually no impact, since configure will just ignore
options it does not recognise. Those options are generic enough that a
lot of packages use them, so we do not repeat them for each packages;
they are part of our generic infrastructure.

> I have now merged your keepalived/ip_vs: fix integer types commit and
> added some other commits to change the remaining u_intnn_t types and
> to include <stdint.h>.

Thank you! :-)

Regards,
Yann E. MORIN.

pqarmitage:
> Maybe. However, Buildroot is a tool to create embedded systems via
> cross-compilation, and packages are optional. Since dependencies on
> libnl3, iptables or ipset are all optional, we do not enforce them; it's
> up to the user to enable what he wants on his system.

Hmm, that's a pity. Building without libnl3 (or libnl) means that keepalived cannot use IPVS with IPv6 (since it isn't supported via the sockets interface). Building without iptables-dev means that keepalived will attempt to invoke the iptables/ip6tables command by fork() and exec(), once for each address to be added/deleted, and the iptables command is horribly slow especially if there is a largish rule set. ipsets are used to overcome the inefficiencies of iptables by adding/deleting addresses to sets instead.

The real reasons for libnl3, iptables and ipset functionality being able to be disabled is in case a system really doesn't support that functionality; if the functionality is available it is rather expected that it will be used. Are you able to make libnl3, iptables and ipset be dependant packages for keepalived? It's hard to imagine anyone not wanting that functionality.

I actually implemented the ipset functionality using dynamic linking, but the iptables functionality isn't implemented that way, and it doesn't attempt to use ipsets if iptables isn't available at build time.

Perhaps this means that 4 different versions need to be build, with and without libn3, combined with with and without iptables/ipsets if you can't make libnl3, iptables and ipsets required dependencies.

> Yes, this has virtually no impact, since configure will just ignore
> options it does not recognise. Those options are generic enough that a
> lot of packages use them, so we do not repeat them for each packages;
> they are part of our generic infrastructure.
That could be a problem if we happened to use one of the options such as `--enable-shared` but in general keepalived shouldn't be build using that option.

yann-morin-1998:
> Are you able to make libnl3, iptables and ipset be dependant packages
> for keepalived?

They already are, but [conditionally](https://git.buildroot.org/buildroot/tree/package/keepalived/keepalived.mk).

So as you can guess, if a user voluntarily enabled libnl, then keepalived
will use it. Ditto iptables and ipset.

In Buildroot, we have a policy not to enforce unnecessary stuff. So,
when an upstream (like keepalived) decides that support for something is
optional, we leave it optional. If a future version of keepalived makes
those mandatory, we'd adapt to match.

Now, what you said is very interesting "it doesn't attempt to use ipsets
if iptables isn't available at build time". As I understand it, it means
that using ipset without iptables does not make sense. We should fix our
conditions, then. Thanks! :-)

> That could be a problem if we happened to use one of the options such
> as --enable-shared but in general keepalived shouldn't be build using
> that option.

`--enable-shared`, as well as `--enable-static`, are "reserved" options;
they are used by libtool to decide whether to build shared and/or static
libs. We pass it unconditionally to all autotools-based packages, and
normally only packages that actually build libs will act upon it. They
should not be used other than to reflect building shared/static libs.

Thanks again for the feedback. Much interesting! :-)

Regards,
Yann E. MORIN.

pqarmitage:
I've found the Buildroot source for the keepalived package now, so I can see how you have the options.

In keepalived.mk is the following quite right?
```
ifeq ($(BR2_PACKAGE_LIBNL),y)
KEEPALIVED_DEPENDENCIES += libnfnetlink
endif
```
I think for using libnl you want:
`KEEPALIVED_DEPENDENCIES += libnl3 libnfnetlink`
since I can't see that libnfnetlink has any dependency on libnl3.

yann-morin-1998:
> I think for using libnl you want:
> KEEPALIVED_DEPENDENCIES += libnl3 libnfnetlink
> since I can't see that libnfnetlink has any dependency on libnl3.

Thank you for spotting this issue! :-)

We have [libnl](https://git.buildroot.org/buildroot/tree/package/libnl/libnl.mk)
which is in fact version 3.2.27 (not the latest, granted, but cloe
enough).

Thanks, this discussion was very interesting. I'll be able to send a few
fixes to Buildroot! :-)

pqarmitage:
@yann-morin-1998 I have now pushed a series of patches that allow keepalived to dynamically link to libnl/libnl-3, libiptc (for iptables) and libipset at runtime. This means that if the keepalived build is configured with the --enable-dynamic-linking option, keepalived will use the libraries if they are available at runtime, otherwise it will fall back to the less functional/performant code this is used if keepalived is build without those libraries.

If this configure option is used for Buildroot, then the single binary will use the libraries if they are available, so no-one is forced to use the libraries, but the best options are chosen at run-time, depending on what is available on the target system.

In order to get this functionality, the development packages for libnl3 iptables and ipset will need to be installed at build time.

PR #568, opened 2017-03-29
In case of consecutive vrrp state notification, need to make sure that state is notified in correct order. To maintain the  order and synchronize the notification adding wait before gi- -ving control back to parent process.

Navjot01:
On changing priority we are seeing state update is not synchronized and wrong state update process is getting called causing mismatch in actual VRRP state in KA and remote module learning from the notify scripts.

Parent process should wait till notify script completes its execution, this is to avoid context switch causing mismatch in state update.

### DIFF
--- a/keepalived/vrrp/vrrp.c
+++ b/keepalived/vrrp/vrrp.c
@@ -1501,7 +1501,7 @@ vrrp_state_backup(vrrp_t * vrrp, char *buf, ssize_t buflen)
 #ifdef _WITH_SNMP_RFCV3_
 		vrrp->stats->master_reason = VRRPV3_MASTER_REASON_PRIORITY;
 #endif
-	} else if ((vrrp->nopreempt || vrrp->effective_priority != VRRP_PRIO_OWNER)
+	} else if ((vrrp->nopreempt &&  vrrp->effective_priority != VRRP_PRIO_OWNER) ||
 		   hd->priority >= vrrp->effective_priority ||
 		   (vrrp->preempt_delay &&
 		    (!vrrp->preempt_time.tv_sec ||
--- a/keepalived/vrrp/vrrp.c
+++ b/keepalived/vrrp/vrrp.c
@@ -1501,7 +1501,7 @@ vrrp_state_backup(vrrp_t * vrrp, char *buf, ssize_t buflen)
 #ifdef _WITH_SNMP_RFCV3_
 		vrrp->stats->master_reason = VRRPV3_MASTER_REASON_PRIORITY;
 #endif
-	} else if (vrrp->nopreempt ||
+	} else if ((vrrp->nopreempt || vrrp->effective_priority != VRRP_PRIO_OWNER)
 		   hd->priority >= vrrp->effective_priority ||
 		   (vrrp->preempt_delay &&
 		    (!vrrp->preempt_time.tv_sec ||
--- a/keepalived/vrrp/vrrp.c
+++ b/keepalived/vrrp/vrrp.c
@@ -1501,7 +1501,7 @@ vrrp_state_backup(vrrp_t * vrrp, char *buf, ssize_t buflen)
 #ifdef _WITH_SNMP_RFCV3_
 		vrrp->stats->master_reason = VRRPV3_MASTER_REASON_PRIORITY;
 #endif
-	} else if ((vrrp->nopreempt &&  vrrp->effective_priority != VRRP_PRIO_OWNER) ||
+	} else if ((vrrp->nopreempt || vrrp->effective_priority != VRRP_PRIO_OWNER)
 		   hd->priority >= vrrp->effective_priority ||
 		   (vrrp->preempt_delay &&
 		    (!vrrp->preempt_time.tv_sec ||
--- a/keepalived/vrrp/vrrp.c
+++ b/keepalived/vrrp/vrrp.c
@@ -1501,7 +1501,7 @@ vrrp_state_backup(vrrp_t * vrrp, char *buf, ssize_t buflen)
 #ifdef _WITH_SNMP_RFCV3_
 		vrrp->stats->master_reason = VRRPV3_MASTER_REASON_PRIORITY;
 #endif
-	} else if ((vrrp->nopreempt || vrrp->effective_priority != VRRP_PRIO_OWNER)
+	} else if (vrrp->nopreempt ||
 		   hd->priority >= vrrp->effective_priority ||
 		   (vrrp->preempt_delay &&
 		    (!vrrp->preempt_time.tv_sec ||
--- a/lib/notify.c
+++ b/lib/notify.c
@@ -111,7 +111,8 @@ script_setup(void)
 int
 notify_exec(const notify_script_t *script)
 {
-	pid_t pid;
+	pid_t pid, cpid;
+	int status = 0;
 
 	pid = fork();
 
@@ -121,9 +122,15 @@ notify_exec(const notify_script_t *script)
 		return -1;
 	}
 
-	/* In case of this is parent process */
-	if (pid)
+	/* In case of parent process, wait for the exit status of child, this is
+        * to synchronize consecutive status update.
+        */
+	if (pid) {
+                while ((cpid = wait(&status)) > 0) {
+                     log_message(LOG_INFO,"Exit status to invoke %s from %d was %d \n",cmd, (int)cpid, status);
+                }
 		return 0;
+        }
 
 #ifdef _MEM_CHECK_
 	skip_mem_dump();


pqarmitage:
So far as I can see, of these 7 commits, 3 are reversing 3 of the others, and the only actual change is commit d9db105.

I don't think we would want to pull all these commits, and their reverts, so creating a new branch and just including commit d9db105 would be a helpful start.

I absolutely see that the sequence of processing the notification scripts in the scenario you describe is non-deterministic, and that that can cause problems.

However, I think there are significant problems with commit d9db105. First of all, we cannot have the keepalived vrrp process suspended waiting for the completion of a notify script. If the notify script takes several seconds to complete, vrrp instances which are master will not be able to send adverts during that time, and backup instances might start taking over as masters. Second, if there are other child processes of the vrrp process, the wait loop will wait until they all have terminated, and will consume their exit statuses, thereby meaning that the main code expecting to receive the exit status of track scripts will miss them, time out and may cause vrrp instances to transition to backup or fault state (using waitpid() rather than wait() would solve this problem, but not the first one). Third, the log message will report the exist status of another child if it terminates, but report it as the exist status of the notify script command.

One possible solution to your problem would be to use SNMP to query the current state of the vrrp instance from within your script, so you always are handing the current status as reported by keepalived, as opposed to the status contained in the notify script.

An alternative, but this would require further patching to keepalived, would be to add a global option `notify_sequence_numbers` which would add an extra parameter to every notify script that would report an increasing sequence number. Then within the scripts it should be possible to check the last processed sequence number, and if the sequence number currently being processed is lower that the last one received, ignore the script notification.

A further alternative, akin to what you are attempting to do, would be to queue within keepalived any further notifications until a currently running notify script has completed, and then on completion of the notify script to check if there are any pending notifications and then just send the latest one. This would mean using `system_call_script()` for notify scripts in lib/notify.c rather than `system_call()`, so that keepalived is aware of the child process executing the script and can handle the termination of the script and hence run a further notification. This approach would also require adding timeouts to notify scripts.

Navjot01:
I will create another branch with changes for sure so that we can avoid bad commits, but i am not able to understand your following doubt  :
"Second, if there are other child processes of the vrrp process, the wait loop will wait until they all have terminated, and will consume their exit statuses, thereby meaning that the main code expecting to receive the exit status of track scripts will miss them, time out and may cause vrrp instances to transition to backup or fault state (using waitpid() rather than wait() would solve this problem, but not the first one). "

We have one VRRP process this process will call notify script for one instance at a time and main process will remain in wait status till notify completes. I don't see another fork till first executes successfully.

I don's see multiple child process or need of waitpid over here or probably i am not able to understand your scenario.

Regarding your first doubt, that in case notify does not complete its execution in time, master will change to backup absolutely correct, but we have minimum 3 sec to complete this execution keeping in mind minimum adv-interval, so even if we don't see execution of notify in less then 3 sec means there is some serious problem.

But best solution which i got from your reply is of maintaining sequence number but for this we need to do a lot of changes.

With my suggested changes, i have tested 255 VRRP instance's and the problem which i was facing that of wrong state update got resolved and every thing works fine.

pqarmitage:
> Second, if there are other child processes of the vrrp process, the wait loop will wait until they all have terminated, and will consume their exit statuses, thereby meaning that the main code expecting to receive the exit status of track scripts will miss them, time out and may cause vrrp instances to transition to backup or fault state (using waitpid() rather than wait() would solve this problem, but not the first one). 

There are (at least) two types of child processes that the vrrp process can create. The first, as you refer to, are the notify scripts; the second are the track scripts. These are scripts that run on a periodic basic (possibly even once per second) and a failure of one of those scripts can be configured to cause a vrrp instance to go into fault state. If one of these scripts terminates during your call to wait(), then the wait() call will consume it's exit status, and the vrrp process elsewhere will not be aware of it.

Furthermore, during the wait() call, if the notify script took several seconds to complete, vrrp instances that are in the master state would be blocked from sending adverts, and instances in the backup state would not be able to expire their timers and take over as master should that be needed.

Whereas you may have tested your change with 255 vrrp instances in your configuration, and maybe your notify scripts complete quickly, other users may have many more vrrp instances, running many track scripts and notify scripts that take several seconds to complete (the theoretical limit is 510 vrrp instances _per logical interface_).

The design of keepalived is that all the **slow** activities, e.g. script execution, are executed asynchronously so that other essential tasks are not blocked. Your patch fundementally changes that to make the execution of notify scripts synchronous, thereby blocking all other activity across all verrp instance while the script runs.

> Regarding your first doubt, that in case notify does not complete its execution in time, master will change to backup absolutely correct, but we have minimum 3 sec to complete this execution keeping in mind minimum adv-interval, so even if we don't see execution of notify in less then 3 sec means there is some serious problem.

Whereas you may use an advertisement interval of one second, the minimum allowed advertisement interval is 0.01seconds, meaning that the failover time for a backup to take over as master is 0.03 seconds. Again, others may run notify scripts that take many seconds, and this won't be a problem with the current method of working since they are effectively _fire and forget_.

> But best solution which i got from your reply is of maintaining sequence number but for this we need to do a lot of changes.

I'm not convinced that this is a very big change, but I think the better solution would be not to execute further notify scripts for a vrrp instance or group until the previous one had completed. This is more complicated that the sequence number change, but means that it can be implemented without any change to the external interface.

When considering changes to keepalived it is essential to consider what impact those changes may have on existing users with their existing configurations. We must do all we can to ensure that we don't break keepalived for existing users simply for an expedient or simple change to satisfy the needs of one or more other users.

If you are not happy producing patches for any of the above solutions, or any alternative that you can come up with that doesn't break the way keepalived currently works, then it is probably best if you log an issue, describing the problem, and then proposed solutions can be discussed and the best solution implemented. It is certainly the case that you have identified an issue that I am not aware that anyone else has raised before, but we must ensure that any solution implemented works for everyone.

pqarmitage:
It's just occurred to me that at the very start of this thread you refer to `On changing priority ...`. Could you explain how the priorities are being changed, so that I can better understand the problem. At the moment I'm not seeing why there should be multiple notify scripts running in quick succession causing the scripts to execute in 'the wrong' order.

Navjot01:
I totally agree with your explanation thanks for writing it in detail. I understand from your explanation that 
adding wait is not the best of the solution to resolve the problem in generic way.

I have also raised an issue for the same, below is the scenario where i am seeing this problem.

Backup ------------------------ Master
  prio 100                                  prio 105

Now on backup side, i keep changing priority to 108 then back to 100. After changing priority number of times on backup side eventually i see both side calling the notify with state as "Master" i.s. last state update is Master even though KA maintains the correct state but notify is not called in correct order causing the problem.

pqarmitage:
Could you please explain what it is you are doing to change the priority, since presumably you are making these changes quite rapidly. I'm not sure at the moment that I can see a way of changing priorities rapidly, unless you are using SNMP. The only other way I know of changing priorities is through track scripts, but that shouldn't be able to happen that rapidly.

How many times per minute (or second) are you changing the priorities? It would really help me if you could both let me have a copy of your configuration file, and also explain how you are changing the priorities, so that I can try and reproduce the problem here.

Navjot01:
I am not using SNMP, we have our own commands which eventually update the keepalived configuration file.
Once KA file is updated, we restart "kill -1 KA pid" the KA process so that new configuration takes place.
Now, by using priority change command "interface vlan 10 vrrp 1 priority 100" this means we are updating the priority of instance 1. 

This command is used to change the priority of vrrp instance from 100 to 108 and back to 100.

Multiple execution of priority change command cause KA process to restart again and again with new priority & this change eventually reproduce the issue.

I hope this will help you to understand the problem.

neutrino1:
Hi,
@pqarmitage @Navjot01
I have discussed my problem & observations in this [link-CPU scheduling of the called scripts by keepalived is not synchronized](https://github.com/acassen/keepalived/issues/584).
The way I observed this issue was **plugging out and then plugging in** of ethernet cable but please correct me, to my understanding **the problem is same** i.e **multiple notify scripts instantiated and now due to context switch or CPU scheduling  the order of notify scripts is not synchronized** .

My conclusion of requirment is following;
## Requirment 1:
I need to queue (FIFO) the notify scripts, plus fire & forget i.e. the scripts should take there time but should execute in order.
## Requirment 2:
 Additionaly I will need some mechanism to make sure that scripts with long  delay dont create a problem, may be I will have a queue size max limit and I will drop new notify scripts when queue size limit is reached OR may use some better approach when queue size limit is reached. 

Requirment 1 seems important, have you guys reached to some solution. Any way out, any way I can contribute to resolve the issue with your help, please.

neutrino1:
@pqarmitage @acassen 
**Is the forking in notify_exec function at line 116  of lib/notify.c necessary ?** 
What if the notify_exec function only write the script details in  the queue in order/serially (without forking), the notify_exec moves on after writing to queue  and an other child process (suppose the child process was spawned as soon as keepalived was executed) reads that queue and execute all the scripts in FIFO order. 

1. notify_exec func write in the queue the script details. 
2. After writing in the queue the main code move forwards. 
3. Another child process that was waiting for the notify scripts, reads the queue.
4. This child process execute the scripts in FIFO queue.
5. The child process after executing scripts again goes to a wait state i.e step 3.

Do you see any issues in the above?

pqarmitage:
@Navjot01 - I think what you describe is feasible, but I'm not sure I see the benefit of having a separate process to run the scripts. If the VRRP process is changed so that it becomes aware of the termination of notify scripts, then it can all be handled within the VRRP process.

I can see one potential problem with what you propose, regardless of whether it is the VRRP process or a child process that executes the scripts. If an interface is flapping, a large number of notify script requests could be created, and the queue might "overflow" in some sort of way; certainly one could envisage a scenario where the execution of the scripts cannot keep up with the generation of requests to run scripts (I think this scenario would also cause a problem with the current implementation with a large number of processes running scripts simultaneously).

I have taken an alternative approach with commit 04905cd, and added a new mechanism for notifying events, which is to deliver them via a FIFO. keepalived will write a line to the FIFO for each event, which can then be read by a separate process. Since there is no asynchronous activity, the events will be delivered in the correct order, and this approach also avoids the overhead of forking a child for each notify event. An example script for reading the FIFO is provided in doc/samples/sample/sample_notify_fifo.sh.

The lines written to the FIFO are simply the parameters that are passed to a notify script, for example:
INSTANCE "VI_1" MASTER 100

I hope you are able to use this new approach, which will both be faster than the current method of forking a child, and will deliver the events in the correct order.

Please advise if this solution will or will not work for you.

neutrino1:
Hi, 
Good work, I also tried this concept of FIFO with **message queues** and it is working in my particular case of issue #584 but my solution needs a review from Keepalived maintainers like you @pqarmitage , should I share it

pqarmitage:
@nautrino1 - yes, happy for you to share your message queues solution

neutrino1:
**Major file changed:**
Changes made in existing files are lib/notify.c and keepalived/core/main.c 
New files added are lib/msq.c & lib/msq.h
Makefile in lib and in keepalived/core needs to be replaced with Makefile_msq in lib & keepalived/core/
msq_key.txt needs to be placed in /. Used for message queue keys. The process can be made better.
keepalived-1.2.15 version 
version with message queue is attached


**Description:**
1: In main.c before starting vrrp child we instantiate a child process that will create the message queue and will wait in the read state their until some notify script command arrives.

2: When there is change in state the notify_exec function is called in keepalived, in this function now instead of forking a new child we just write the notify script cmd in the queue.

3: The child in wait state in step 1, read the notify script command in the queue and execute it using system_call as done previously in the keepalived.

4: On pkill keepalived the child process will be removed and the queue will be deleted.

5: The code is tested and is working without any crash and resolved my problem #584 till now.

**Further Work Required:**
1: On some places the code is disabled as I have not handled **signals** in the child process that was instantiated to read the scripts in the queue but the code is without compilation error.

Your kind look will help to better understanding. With your help I can make it better.
[keepalived-1.2.15-msq.tar.gz](https://github.com/acassen/keepalived/files/1019823/keepalived-1.2.15-msq.tar.gz)


[](url)

neutrino1:
For review patch for keepalived-1.2.15 with message queue implementation is attached.
you can place this patch in keepalived-1.2.15 and use command 

> patch -p1 < ../keepalived-1.2.15-msq.patch


[keepalived-patch.tar.gz](https://github.com/acassen/keepalived/files/1020673/keepalived-patch.tar.gz)
Waiting for your comments.

pqarmitage:
@neutrino1 I've had a look at your message queue code, and ported to to the current master branch (v1.3.5 commit e8421dd). I've streamlined the code, resolved one or two areas that I think are issues and added some comments where I think there are some further issues. I've also moved where start_msgr_child() is called from.

Attached is a diff from the current HEAD of master for the modified message queue code.
[msg.diff.txt](https://github.com/acassen/keepalived/files/1029607/msg.diff.txt)

Having thought about this though, while I believe it can be made to work safely I don't think it is the correct way to go. It wouldn't be very difficult to change keepalived so that it handles the termination of notify scripts; it can then maintain a queue per vrrp instance and when one notify script completes, if there are any entries in the queue the next one is executed (as an alternative it might work not to have a queue but to simply run a notify script for the latest status change of the instance, since the previous ones are historical and presumably aren't important since their status has been superseded). My concern is that if something occurs such as an interface flapping, it could result in a large number of script executions being queued, and could eventually lead to a denial of service (only running the script for the latest state change would significantly help this).

I think the best way forward is to use the FIFO approach for reliable sequenced notifications, and if someone wants to they can run scripts from the process that is reading the FIFO, or alternately simply incorporate the functionality into the process that is doing the reading - it should be straightforward to modify the sample_notify_fifo.sh script. A significant benefit of the FIFO approach is the if notify event can be handled within the process reading the FIFO, then the overhead of forking (actually 2 or 3 forks and 1 or two execs) per script invocation can be avoided.

I am considering whether it would be worth being able to configure a separate FIFO per vrrp instance and per virtual server.

Any thoughts from anyone on what would be best/what works for them/doesn't work for them would be welcome.

neutrino1:
> @neutrino1 I've had a look at your message queue code, and ported to to the current master branch (v1.3.5 commit e8421dd).

Thank you for the review & all the effort. On applying this msg.diff patch on current master, will it not create conflict with your FIFO implementation?

>  I've streamlined the code, resolved one or two areas that I think are issues and added some comments where I think there are some further issues. I've also moved where start_msgr_child() is called from.

I have seen your comments, should I further correct those points, beside learning if corrected is there still chance that it could be used in keepalived master code?

> Having thought about this though, while I believe it can be made to work safely I don't think it is the correct way to go. It wouldn't be very difficult to change keepalived so that it handles the termination of notify scripts; it can then maintain a queue per vrrp instance and when one notify script completes, if there are any entries in the queue the next one is executed (as an alternative it might work not to have a queue but to simply run a notify script for the latest status change of the instance, since the previous ones are historical and presumably aren't important since their status has been superseded). My concern is that if something occurs such as an interface flapping, it could result in a large number of script executions being queued, and could eventually lead to a denial of service (only running the script for the latest state change would significantly help this).

You want to run notify scripts for the latest status change, that is definitely a right idea. may be it will be like following steps;
 1: Due to status change once a notify script is running you have to wait for its graceful completion. 
 2 : Meanwhile the step 1 was not completed & if there is a status change you will keep its record but you will not load the script as the step 1 is still running. May be the record can be a queue of just size 1 or may be some , old historical will be removed either if they get executed or if replaced by a new status change.other better idea etc.
 3: Another status change occurs and you will overwrite the record that was generated in step 2  but again you will not load the script as I suppose that  the step 1 is still running but now the record of step 2 is overwritten. Step 2 has become history its script will never get chance to be executed and overwritten with step 3.
4: So on and so forth you will not load the new notify script until the step 1  is completed gracefully and only the last status change will be kept in the record.
By gracefully completion of step 1, I mean to say we don't want to interrupt this loaded script execution. As we never know that it may be in the middle of something e.g executing other process, file writing etc. 
I am not sure, can there be a use case that where some one wants to log each intermediate change (that we are calling as history and dropping), may be an option can be left were user can decide what to use i.e record only last status change OR record all status changes.
Is the current implementation handling above?
 

>  It wouldn't be very difficult to change keepalived so that it handles the termination of notify scripts; it can then maintain a queue per vrrp instance and when one notify script completes

> I think the best way forward is to use the FIFO approach for reliable sequenced notifications, and if someone wants to they can run scripts from the process that is reading the FIFO, or alternately simply incorporate the functionality into the process that is doing the reading - it should be straightforward to modify the sample_notify_fifo.sh script. A significant benefit of the FIFO approach is the if notify event can be handled within the process reading the FIFO, then the overhead of forking (actually 2 or 3 forks and 1 or two execs) per script invocation can be avoided.

You don't want to create a new process to read FIFO. You want existing process to be linked through FIFO and perform IPC, and those two process are keepalived and vrrp_instance OR vrrp instance and per virtual server. Am I right? Again Is the current implementation handling above?

May be my questions are insane and get clear once I execute the latest master code.

pqarmitage:
@neutrino1 Many thanks for your input to this and the code you have produced. The ideas you presented led to the implementation of notifcation fifos, that guarantee sequential delivery. We will, however, not implement sequential execution of notify scripts, due to the possibility of generating a longer and longer backlog of scripts to be run. Use of `notify_fifo` is the way to go for guaranteed sequential, and that is now incorporated in the current v2.0.0 release released yesterday. Please note that the doc/samples directory has a sample skeleton script that can be run to handle fifo notifications, and keepalived can be configured to run the script at startup.

If you do not feel that this is sufficient for your needs, then would you please raise an issue report identifying what your issues are with the current implementation, and there can then be a discussion about the best way forward.

PR #570, opened 2017-03-29
Log missing adverts from MASTER

rolffokkens:
VRRP prority changes are logged, but a lack of adverts is not logged. This change adds the logging of this event,


pqarmitage:
I'm not sure I understand the benefit of adding the logging of a received advert timeout. The action that immediately follows this is a transition to master (which is fuly logged), and the only reason for transitioning to master (unless the instance is the address owner) is a received advert timeout.

It seems to me that this is adding logging of additional information that can be determined from the existing logged information.

I assume I am missing something, so could you please explain what the additional logging adds to knowing what has occurred. I'm quite happy to merge the commit if I can understand the benefit.

rolffokkens:
At a certain point last week I looked in to a VM bouncing van MASTER to BACKUP and v.v. There wasn't much to go on: it sometimes just claimed MASTERship, and then a minute later it became BACKUP again. The only explanation was that the other VM drop of the network or something like that, but there was no indication as such. ICMP kept working, so what was going on? WAS the VM claiming MASTERship indeed because the other VM hadn't been sending adverts? Or was something else going on?
Debugging based on assumptions requires to validate the assumptions. And this log message just dit that.
After being absolutely sure that the adverts weren't received, I dug deeper into this. Later it appeared that the host machine's I/O system was suffering intermittent freezes, at which times the receiving VM's keepalived froze too for about 5 seconds. On defrosting it apparently handled the timeout claiming that no adverts had been received.

So in a nutschell: I was able the make the same assumption you did ("the only reason for transitioning to master is a received advert timeout") but with less certainty (I'm not a VRRP expert, not can I oversee the complete keepalived code). The aditional log message promoted my assumption to certainty, which was very helpful in the debugging process.

And why not assume other (non expert) users may benefit from this message as well?

brandt:
We were in this same position last week while investigating an unexpected failover.

We just rolled out keepalived recently, so we are in that non-expert group. This message would have helped narrow down the cause.

pqarmitage:
A message logging receive advert timeout has now been added when using the log detail (-D) command line option.

PR #571, opened 2017-03-30
add Dockerfile

osixia:
Here the pull request to add a Dockerfile.
Let me know if all required packages are installed and if the unnecessary one are deleted after compile.

Keepalived seems to need a "keepalived_script" user, i tried to add one with the following command : 
`addgroup -S keepalived_script && adduser -D -S -G keepalived_script keepalived_script`
but i still get:
```
Unable to resolve default script username 'keepalived_script' - ignoring
```
if you have any idea ?

thanks.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,50 @@
+FROM alpine:3.5
+MAINTAINER Bertrand Gouny <bertrand.gouny@osixia.net>
+
+# add keepalived sources to /tmp/keepalived-sources
+ADD . /tmp/keepalived-sources
+
+# Add keepalived default script user to make sure their IDs get assigned consistently,
+# regardless of whatever dependencies get added
+RUN addgroup -S keepalived_script && adduser -D -S -G keepalived_script keepalived_script
+
+# 1. install required libraries and tools
+# 2. compile and install keepalived
+# 3. remove keepalived sources and unnecessary libraries and tools
+RUN apk --no-cache add \
+       gcc \
+       ipset \
+       ipset-dev \
+       iptables \
+       iptables-dev \
+       libnfnetlink \
+       libnfnetlink-dev \
+       libnl3 \
+       libnl3-dev \
+       make \
+       musl-dev \
+       openssl \
+       openssl-dev \
+
+    && cd /tmp/keepalived-sources \
+    && ./configure --disable-dynamic-linking \
+    && make && make install \
+    && cd - \
+
+    && rm -rf /tmp/keepalived-sources \
+    && apk --no-cache del \
+        gcc \
+        ipset-dev \
+        iptables-dev \
+        libnfnetlink-dev \
+        libnl3-dev \
+        make \
+        musl-dev \
+        openssl-dev
+
+# set keepalived as image entrypoint with --dont-fork and --log-console (to make it docker friendly)
+# define /usr/local/etc/keepalived/keepalived.conf as the configuration file to use
+ENTRYPOINT ["/usr/local/sbin/keepalived","--dont-fork","--log-console", "-f","/usr/local/etc/keepalived/keepalived.conf"]
+
+# example command to customise keepalived daemon:
+# CMD ["--log-detail","--dump-conf"]


pqarmitage:
Manually running the `addgroup -S keepalived_script` and `adduser -D -S -G keepalived_script keepalived_script` commands in the container shows that -S, -D and -G do not appear to be valid options.

Running `groupadd keepalived_script`, and `useradd -g keepalived_script keepalived_script` appear to work in the container. You may want to add some further options (I found `useradd --help` and `groupadd --help` useful); it appears that `useradd` and `groupadd` are significantly more up-to-date executables that `adduser` and `addgroup`.

osixia:
Alpine is very light weight, so a container can be up very quickly.
```
$ docker images
REPOSITORY                           TAG                 IMAGE ID            CREATED             SIZE
debian                               jessie              8cedef9d7368        8 days ago          123 MB
alpine                               3.5                 4a415e366388        3 weeks ago         3.99 MB
```

The keepalived image with alpine will be around 22mb uncompressed compared to 123mb for the debian jessie baseimage.

The dockerfile is now in the root directory of the sources, that will allow the docker hub to build the image automatically.

You can also build and push the image manually by doing:
```
docker build -t keepalived .
```
`-t` to set the image tag (here keepalived) 
`.` means send this directory to build context

the keepalived tag is ok if you want to use the image locally but if you want to push the image to the docker hub, the image must be tagged to something like login/imagename:tag

then you can push the image to the dockerhub:
```
docker push login/imagename:tag
```

when the Dockerfile will be ok and in the sources we see how to set up the automatic build on the docker hub :)

osixia:
for example to start a container with the custom config in /data/keepalived.conf on the host:
```
docker run --cap-add=NET_ADMIN --net=host -v /data/keepalived.conf:/usr/local/etc/keepalived/keepalived.conf keepalived --log-detail --dump-conf
```

pqarmitage:
I've resolved the
```
Unable to resolve default script username 'keepalived_script' - ignoring
```
error. It was my stupid fault - I was setting the the buffer size for getpwnam after calling getpwnam!

pqarmitage:
When I try running
`docker run --cap-add=NET_ADMIN --net=host -v /data/keepalived.conf:/usr/local/etc/keepalived/keepalived.conf keepalived --log-detail --dump-conf`

in the docker logs I get the error message
`Configuration file '/usr/local/etc/keepalived/keepalived.conf' open problem (Permission denied) - skipping`
and indeed if I try opening the file from within the container I get permission denied.

Are there some permissions needed on the /data directory or keepalived/conf file to allow this to work?

pqarmitage:
I've sorted the problem of opening the conf file when mounted with -v. I needed to add the `--security-op label=disable` option.

pqarmitage:
Keepalived adds network  configuration (iptables, ipsets and ipvs) in the host system. Terminating the container with `docker rm -f keepalived` doesn't allow keepalived to cleanup after itself, and so the network configuration gets left in host configuration.

In order to make keepalived terminate cleanly, it needs to be sent a SIGTERM. I can make it work with `docker exec keepalived kill -TERM $(docker exec keepalived cat /var/run/keepalived.pid)`
but I would really like something like
`docker exec keepalived kill -TERM \$(cat /var/run/keepalived.pid)`
where cat /var/run/keepalived.pid is executed in the container, but I can't work out how to do that.

pqarmitage:
The `--disable-dynamic-linking` configure option is no longer required since commit f7cd991. However, since it is known that all the required libraries exist in the container, it probably makes sense to keep the option, since it is designed to allow keepalived to run if libipset isn't installed on the target system.

pqarmitage:
I've made some minor changes to the setup.

1. I've added a make target docker, so that `make docker` will create the docker image
2. I've added docker/README that summarises the information @osixia has provided for creating and using docker containers
3. The keepalived.conf file installed by `make install` isn't ideal for running keepalived, so I've created a very simple keepalived.conf file in docker/ which is now the configuration file installed into the image.

osixia:
thanks for merging this :)

i just checkout the master and try to build the image and had the following error:
```
check/libcheck.a(ipvswrapper.o): In function `ipvs_talk':
/tmp/keepalived-sources/keepalived/check/ipvswrapper.c:238: undefined reference to `write_stacktrace'
make[2]: *** [Makefile:462: keepalived] Error 1
```
is that just a dev bug ?

It would be better after setting up the docker automated build to not dev on the master but in an other branch to always offer a functionnal image here is why:

most of the time when people try a docker image they just do something like:
```
docker run mysql 
```
this is a shortname for running mysql:latest docker image tag

With an automated build we can set a config to tell docker that the *latest* docker tag correspond to a git branch, master for example, so when a commit is pushed on the master branch this build the docker image and tag it *latest*.

Furthermore it can also build a docker image and tag it based on the git tag, see the config example below:

capture du 2017-04-01 10-32-40

So when a git tag is created it would build the image and tag with the same git tag name.

Let me know if this is not clear :) 

Also to set up the automated build this repository owner, or maybe a colloborator (not sure, we can try if you want just add me as collaborator)  must create a docker hub account and link is github account to it.

To create an automated build, once loggged in the docker hub : 

In the top bar:
Create ->  Create Automated Build

Choose Github and select keepalived repository.

Add a description and click "Click here to customize"
Enter the same config as the image.

Done :)

here the full doc: 
https://docs.docker.com/docker-hub/builds/

osixia:
Just saw the new docker readme  nice :)

i will probably make a pull request with some adjustments
for example `docker exec -it keepalived bash` do not work with the alpine baseimage (bash is not installed only sh) it would becomes: `docker exec -it keepalived sh`

pqarmitage:
You're right that the compile error is a dev bug. I've merge a commit that resolves it now.

Please do make any changes you think useful - I'm not a docker expert myself.

I'm a collaborator so I'll have a go at setting up an automated build later.

PR #581, opened 2017-04-18
logger: output timestamps to console logs

markrechler:
Thought this might be helpful for people not using/leveraging syslog facilities. Tracking events in daemontools logs (console logs) was problematic. Let me know if I can change anything or would prefer a different approach.

### DIFF
--- a/lib/logger.c
+++ b/lib/logger.c
@@ -24,6 +24,7 @@
 
 #include <stdio.h>
 #include <stdbool.h>
+#include <time.h>
 
 #include "logger.h"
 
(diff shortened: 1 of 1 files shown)


pqarmitage:
I think this is a very sensible and useful addition. My only comment is that I think I would prefer use of `localtime_r()` rather than `localtime()`, since keepalived can be built to be multithreaded, if using DBus support. I know elsewhere `localtime()` is used, but it would be good not to make the situation worse.

markrechler:
Thanks for the feedback @pqarmitage. Adjusted to use `localtime_r()`. Let me know if I can tweak anything else.

PR #587, opened 2017-05-15
do some wait when exec script to avoid out-of-order execution

glbrtchen:


pqarmitage:
Please see comments in pull request #568 for reasons why this commit cannot be merged.

glbrtchen:
Ok. I know why the commit cannot by accepted. The reason why we change the code is using the notify_script to save the state of current VRRP, but the execution is not in a sequence. I think SNMP could solve my problem, but not so friendly. If there were a restful api could be better.

pqarmitage:
Commit 04905cd provides an alternative method to receive notification of events, which will be delivered in the correct order. Please see the comment in pull request #568 for how the mechanism for notifying events via a FIFO work.

Issue #492, opened 2017-01-09
reload inhibited servers


ze42:
Hello,

Removing an inhibited server from the pool, and reloading does NOT remove it.

Adding it later with a reload would probably fail, as it is already there, though not managed by keepalived anymore.

clear_diff tends to check if ISALIVE, instead of checking if the RS is set.

PR #596, opened 2017-06-08
Delete disabled inhibit servers at reload

pandax381:
Hi,

I fixed a problem that some server removed from the new configuration will not be deleted at reload.

(I think that this pull request solves issue of #492)

+ current state

  ```
  # ipvsadm -L
    -> RemoteAddress:Port     Forward Weight ActiveConn InActConn
  TCP  192.168.100.1:http lc
    -> 192.168.2.2            Masq    0      0          0         
    -> 192.168.2.3            Masq    1      0          0     
  ```

  Inhibit_on_failure is set in 192.168.2.2

+ delete realserver `192.168.2.2` from virtualserver & reload

  ```
  # ipvsadm -L
    -> RemoteAddress:Port     Forward Weight ActiveConn InActConn
  TCP  192.168.100.1:http lc
    -> 192.168.2.2            Masq    0      0          0         
    -> 192.168.2.3            Masq    1      0          0     
  ```

  `192.168.2.2` has been deleted from the new configuration, but it remains in the IPVS entry after reloading.

  If the weight is not 0 (that's alive), it's deleted from the IPVS entry after reloading.

  ```
  # ipvsadm -L
    -> RemoteAddress:Port     Forward Weight ActiveConn InActConn
  TCP  192.168.100.1:http lc
    -> 192.168.2.3            Masq    1      0          0     
  ```

    In this case, it will be deleted regardless of the ActiveConn.

I think that it's better to delete non alived servers as well.

We can delete it after confirming that there is no active connection, but it seems to very troublesome.

(If I have not overlooked something...)

Regards,

### DIFF
--- a/keepalived/check/ipwrapper.c
+++ b/keepalived/check/ipwrapper.c
@@ -800,8 +800,12 @@ clear_diff_s_srv(virtual_server_t *old_vs, real_server_t *new_rs)
 				SET_ALIVE(old_rs);
 			old_rs->inhibit = 0;
 		}
-		if (ISALIVE(old_rs))
+		if (ISALIVE(old_rs)) {
+			log_message(LOG_INFO, "Removing sorry server %s from VS %s"
+					    , FMT_RS(old_rs, old_vs)
+					    , FMT_VS(old_vs));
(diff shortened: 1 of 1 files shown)


pqarmitage:
@pandax31 Many thanks for these fixes.

PR #599, opened 2017-06-09
Add state EQUAL

ryno83:
Probably a good idea to add the EQUAL state in the doc.

### DIFF
--- a/doc/source/configuration_synopsis.rst
+++ b/doc/source/configuration_synopsis.rst
@@ -134,7 +134,7 @@ VRRP Instance Definitions Synopsis
             (or **notify_fault** “ /path_to_script/script_fault.sh <arg_list>”)
     }
     **vrrp_instance** string {
-        **state** *MASTER|BACKUP*
+        **state** *MASTER|BACKUP|EQUAL*
         **interface** string
         **mcast_src_ip** @IP
         **lvs_sync_daemon_interface** string


pqarmitage:
I'm not clear of the reason for the suggestion of adding `EQUAL`. The `state` parameter only accepts `MASTER` or `BACKUP` - see function vrrp_state_handler() in keepalived/vrrp/vrrp_parser.c

Can you provide an explanation of the reasons for adding `EQUAL`.

ryno83:
Because I'm using EQUAL in my setup and it works fine. It's not in the official documentation and this is why I would like to add it. 

I found this https://serverfault.com/questions/834168/what-does-mean-equal-state-of-keepalived-config and this was the exact behaviour I wanted for my setup. Then I tried the EQUAL state even if it's not in the official doc and it work fine and as expected.

But now... I don't understand how it's possible if the state parameter only accept MASTER or BACKUP. Maybe it's because I use the same priority on both nodes ?

pqarmitage:
You should have an entry in your logs saying something like `unknown state 'EQUAL', defaulting to BACKUP`.

Try removing the `state EQUAL` line from your config and see if keepalived still behaves the same way. You could also try changing `EQUAL` to `RENAUD` and see what happens then. I think you will find that the behaviour is exactly the same!

ryno83:
Indeed !
thank you

PR #604, opened 2017-06-21
build: add basic/initial .travis.yml

commodo:
For me, this produces green builds
https://travis-ci.org/commodo/keepalived/builds/245245914

All that's needed now, is for user `acassen` to
go to `https://travis-ci.org/` login with the Github
account, import repos from Github, and enable build
for keepalived [a checkbox/button].

This should serve as a small bump into adding CI/CD to keepalived.
It could encourage/allow others to write some tests/comp tests via the build [something like `make tests` or whatever].
Or even add more configuration matrix/builds ; to try building keepalived with more options enabled/disabled.

Signed-off-by: Alexandru Ardelean <ardeleanalex@gmail.com>


pqarmitage:
Many thanks for providing this basic travis-ci configuration. As you say, it gives us a point to start from.

I've done some experimenting with this to see how useful this can be.

First of all I've had to add the installation of some development libraries (as specified in the INSTALL file), but there seems to be a more significant issue regarding the version of the kernel headers.

I added `dist: trusty` to .travis.yml so that it builds on a not so old version of Ubuntu (14.04 rather than 12.04). The problem is that, although the kernel is v4.4.0, the userspace kernel header files, provided by package linux-libc-dev, are from a v3.13.0 kernel. keepalived interacts quite significantly with the kernel, and uses quite a number of features (if they are available) which are only available in post 3.13 kernels (a quick count shows about 20 comments in configure.ac and C source files referring to post 3.13 kernels). I've worked out a way to update the kernel headers to v4.4.0, but unfortunately there is quite a bit of code that cannot be tested with travis-ci due to needing a later kernel.

I've now pushed the updates for the above, and if you have any comments about them, please add them to this log.

commodo:
I wonder if you missed or did not activate something for the "build passing tag"
or did i goof something there ?

pqarmitage:
This hasn't yet been set up on travis-ci, which is why nothing it is still showing `build unknown` on the project's from page on github. I will need to ask Alexandre if he can activate the project on travis, if he is happy to do so.

commodo:
ack

PR #633, opened 2017-08-09
Adds an extra signal that dumps both data and stats on json format to…

dclabaut-ovh:
… /tmp/keepalived.json. The option has to be enabled at compilation using --enable-json (can also be added to debian/rules if you build for debian). Adds dependency to libjson-c when activated.


keepalived/vrrp/vrrp_print.c
```diff
@@ -418,3 +421,182 @@ vrrp_print_stats(void)
...
+	file = fopen ("/tmp/keepalived.json","w");
+	if (!file) {
+		log_message(LOG_INFO, "Can't open /tmp/keepalived.json (%d: %s)",
+			errno, strerror(errno));
+		return;
+	}
+
+	list l = vrrp_data->vrrp;
+	if LIST_ISEMPTY(vrrp_data->vrrp) {
+		log_message(LOG_INFO, "vrrp_data->vrrp is empty");
```
phob0s-pl: You forgot to close file there.

dclabaut-ovh: Thank you. It's fixed now.

keepalived/vrrp/vrrp_print.c
```diff
@@ -418,3 +421,182 @@ vrrp_print_stats(void)
...
+		return;
+	}
+
+	element e, f;
+	vrrp_t *vrrp;
+	struct json_object *array, *vips;
```
pqarmitage: The above three lines should be at the start of the function, or at the start of the loop, depending on how you resolve the next comment.

keepalived/vrrp/vrrp_print.c
```diff
@@ -418,3 +421,182 @@ vrrp_print_stats(void)
...
+	array = json_object_new_array();
+
+	for (e = LIST_HEAD(l); e; ELEMENT_NEXT(e)) {
+		vrrp = ELEMENT_DATA(e);
+		list v = vrrp->vip;
+		struct json_object *instance_json, *json_stats, *json_data;
```
pqarmitage: There is inconsistency about where some of the variables that are only used in the loop are declared. Either they should all be declared at the beginning of the function, or all at the start of the loop.

keepalived/vrrp/vrrp_print.c
```diff
@@ -418,3 +421,182 @@ vrrp_print_stats(void)
...
+		json_object_object_add(json_data, "strict_mode",
+			json_object_new_int(vrrp->strict_mode));
+		json_object_object_add(json_data, "vmac_flags",
+			json_object_new_int(vrrp->vmac_flags));
+		json_object_object_add(json_data, "vmac_ifname",
+			json_object_new_string(vrrp->vmac_ifname));
+		json_object_object_add(json_data, "master_priority",
+			json_object_new_int(vrrp->master_priority));
+		json_object_object_add(json_data, "last_transition",
+			json_object_new_int(vrrp->last_transition.tv_sec));
```
pqarmitage: I suggest for variables of type timeval_t type declare a function
```
+static inline double
timeval_to_double(const timeval_t *t)
{
       /* The casts are necessary to avoid conversion warnings */
       return (double)t->tv_sec + (double)t->tv_usec / TIMER_HZ_FLOAT;
}
```
and in lib/timer.h after the definition of TIMER_HZ add
`#define TIMER_HZ_FLOAT  1000000.0`

and then line 473 would become:
`json_object_new_double(timeval_to_double(&vrrp->last_transition));`

keepalived/vrrp/vrrp_print.c
```diff
@@ -418,3 +421,182 @@ vrrp_print_stats(void)
...
+		json_object_object_add(json_data, "garp_delay",
+			json_object_new_int(vrrp->garp_delay));
```
pqarmitage: This should be `vrrp->garp_delay / TIMER_HZ` so that it is reported in seconds.

keepalived/vrrp/vrrp_print.c
```diff
@@ -418,3 +421,182 @@ vrrp_print_stats(void)
...
+		json_object_object_add(json_data, "garp_refresh",
+			json_object_new_int(vrrp->garp_refresh.tv_sec));
+		json_object_object_add(json_data, "garp_refresh_timer",
+			json_object_new_int(vrrp->garp_refresh_timer.tv_sec));
```
pqarmitage: use timeval_to_double()

keepalived/vrrp/vrrp_print.c
```diff
@@ -418,3 +421,182 @@ vrrp_print_stats(void)
...
+		json_object_object_add(json_data, "vmac_ifname",
+			json_object_new_string(vrrp->vmac_ifname));
+		json_object_object_add(json_data, "master_priority",
+			json_object_new_int(vrrp->master_priority));
+		json_object_object_add(json_data, "last_transition",
+			json_object_new_int(vrrp->last_transition.tv_sec));
+		json_object_object_add(json_data, "garp_delay",
+			json_object_new_int(vrrp->garp_delay));
+		json_object_object_add(json_data, "garp_refresh",
+			json_object_new_int(vrrp->garp_refresh.tv_sec));
```
pqarmitage: It's not worth using timeval_to_double() here since it can only be a whole number of seconds, so just cast this to int to avoid the conversion warning.

keepalived/vrrp/vrrp_print.c
```diff
@@ -418,3 +421,182 @@ vrrp_print_stats(void)
...
+		json_object_object_add(json_data, "garp_refresh_timer",
+			json_object_new_int(vrrp->garp_refresh_timer.tv_sec));
+		json_object_object_add(json_data, "garp_rep",
+			json_object_new_int(vrrp->garp_rep));
+		json_object_object_add(json_data, "garp_refresh_rep",
+			json_object_new_int(vrrp->garp_refresh_rep));
+		json_object_object_add(json_data, "garp_lower_prio_delay",
+			json_object_new_int(vrrp->garp_lower_prio_delay));
```
pqarmitage: This should be `(int)(vrrp->garp_lower_prio_delay / TIMER_HZ)` so that it is reported in seconds.

keepalived/vrrp/vrrp_print.c
```diff
@@ -418,3 +421,182 @@ vrrp_print_stats(void)
...
+
+		// Dump data to json
+		json_object_object_add(json_data, "iname",
+			json_object_new_string(vrrp->iname));
+		json_object_object_add(json_data, "dont_track_primary",
+			json_object_new_int(vrrp->dont_track_primary));
+		json_object_object_add(json_data, "skip_check_adv_addr",
+			json_object_new_int(vrrp->skip_check_adv_addr));
+		json_object_object_add(json_data, "strict_mode",
+			json_object_new_int(vrrp->strict_mode));
```
pqarmitage: If compiling with configure option `--enable-conversion-checks` a large number of conversion warnings are generated. This seems to be in the main because the json-c library doesn't have `json_object_new_uint` or `json_object_new_uint64` functions. However, in order to suppress the warnings, the value can either be cast to an int or int64_t, or a unsigned int can be used with the json_object_new_int64() function. I haven't identified all the cases below.

keepalived/vrrp/vrrp_print.c
```diff
@@ -418,3 +421,182 @@ vrrp_print_stats(void)
...
+				json_object_new_string(ipaddr));
+		}
+		json_object_object_add(json_data, "vips", vips);
+
+		json_object_object_add(json_data, "promote_secondaries",
+			json_object_new_boolean(vrrp->promote_secondaries));
+		json_object_object_add(json_data, "evip_add_ipv6",
+			json_object_new_boolean(vrrp->evip_add_ipv6));
+		json_object_object_add(json_data, "master_adver_int",
+			json_object_new_int(vrrp->master_adver_int));
```
pqarmitage: This should be `json_object_new_double(vrrp->master_adver_int / TIMER_HZ_FLOAT))` so that it is reported in seconds.

keepalived/vrrp/vrrp_print.c
```diff
@@ -418,3 +421,182 @@ vrrp_print_stats(void)
...
+		json_object_object_add(json_data, "accept",
+			json_object_new_int(vrrp->accept));
+		json_object_object_add(json_data, "iptable_rules_set",
+			json_object_new_boolean(vrrp->iptable_rules_set));
+		json_object_object_add(json_data, "nopreempt",
+			json_object_new_boolean(vrrp->nopreempt));
+		json_object_object_add(json_data, "preempt_delay",
+			json_object_new_int(vrrp->preempt_delay));
```
pqarmitage: Use `(int)(vrrp->preempt_delay / TIMER_HZ)` so that it is reported in seconds.

keepalived/vrrp/vrrp_print.c
```diff
@@ -418,3 +421,182 @@ vrrp_print_stats(void)
...
+		json_object_object_add(json_data, "preempt_time",
+			json_object_new_int(vrrp->preempt_time.tv_sec));
```
pqarmitage: Use timeval_to_double

keepalived/vrrp/vrrp_print.c
```diff
@@ -418,3 +421,182 @@ vrrp_print_stats(void)
...
+		json_object_object_add(json_data, "quick_sync",
+			json_object_new_int(vrrp->quick_sync));
+		json_object_object_add(json_data, "version",
+			json_object_new_int(vrrp->version));
+		json_object_object_add(json_data, "smtp_alert",
+			json_object_new_boolean(vrrp->smtp_alert));
+		json_object_object_add(json_data, "notify_exec",
+			json_object_new_boolean(vrrp->notify_exec));
+		json_object_object_add(json_data, "ms_down_timer",
+			json_object_new_int(vrrp->ms_down_timer));
```
pqarmitage: `json_object_new_double(vrrp->ms_down_timer / TIMER_HZ_FLOAT));`

keepalived/vrrp/vrrp_print.c
```diff
@@ -418,3 +421,182 @@ vrrp_print_stats(void)
...
+		json_object_object_add(json_data, "sands",
+			json_object_new_int(vrrp->sands.tv_sec));
```
pqarmitage: Use timeval_to_double

keepalived/vrrp/vrrp_print.c
```diff
@@ -418,3 +421,182 @@ vrrp_print_stats(void)
...
+		json_object_object_add(json_data, "auth_type",
+			json_object_new_int(vrrp->auth_type));
```
pqarmitage: The above two lines need conditional compilation:
#ifdef _WITH_VRRP_AUTH_
...
#endif

keepalived/vrrp/vrrp_print.c
```diff
@@ -418,3 +421,182 @@ vrrp_print_stats(void)
...
+		json_object_object_add(json_data, "ip_id",
+			json_object_new_int(vrrp->ip_id));
+
+		// Dump stats to json
+		json_object_object_add(json_stats, "advert_rcvd",
+			json_object_new_int64(vrrp->stats->advert_rcvd));
```
pqarmitage: A number of the stats values are uint64_t and need casting to int64_t to work around the limitations of json-c.

keepalived/vrrp/vrrp_print.c
```diff
@@ -418,3 +421,182 @@ vrrp_print_stats(void)
...
+		json_object_object_add(instance_json, "stats", json_stats);
+
+		// Add instance_json to main array
+		json_object_array_add(array, instance_json);
+
+	}
+	fprintf(file, "%s", json_object_to_json_string(array));
+	fclose(file);
+}
+#endif
```
pqarmitage: The following items are probably not worth reporting, since they only have meaning internally to keepalived:
       vrrp->debug
       vrrp->evip_add_ipv6
       vrrp->fd_in
       vrrp->fd_out
       vrrp->garp_pending
       vrrp->garp_refresh_timer
       vrrp->gna_pending
       vrrp->iname
       vrrp->init_state
       vrrp->ip_id
       vrrp->iptable_rules_set
       vrrp->ms_down_timer
       vrrp->notify_exec
       vrrp->preempt_time
       vrrp->quick_sync
       vrrp->sands
       vrrp->vipset
       vrrp->vmac_flags
       vrrp->vmac_ifname
       vrrp->wantstate

The following items are reported in the standard print output, and are probably worth including in the JSON output
       vrrp->adver_int
       vrrp->auth_data
       vrrp->evip
       vrrp->ifp->ifname
       if (vrrp->ifp->vmac) if_get_by_ifindex(vrrp->ifp->base_ifindex)->ifname
       vrrp->lower_prio_no_advert
       vrrp->master_saddr
       vrrp->saddr
       vrrp->script
       vrrp->script_backup
       vrrp->script_fault
       vrrp->script_master
       vrrp->script_stop
       vrrp->track_ifp
       vrrp->track_script
       vrrp->vroutes
       vrrp->vrules

PR #664, opened 2017-10-06
Reinforce that '@' conditional config character must be 1st on line

pqarmitage:


blakehawkins:
Without having reviewed the parser code, given that beginning of line whitespace has no semantic meaning, shouldn't this be fairly trivial to allow?

In the most trivial sense, all it would require is trimming `^\s+` from the config file.

pqarmitage:
Yes it would be possible but I just didn't implement it (I think my view at the time was that I wanted the `@config-id` to stand out in the config, and so be at the beginning of the line).

Since you've made the suggestion, I've now merged commit 8abfeaa which allows whitespace before conditional configuration `@config-id`.

PR #681, opened 2017-10-20
Dev/damien.clabaut/python scripts

dclabaut-ovh:
Those are a few scripts that help with keepalived in the case of many instances:
- template_vrrp_instance, called by make_conf.py, can generate a large config file.
- vrrp_recap prints a table with most of the important info from every vrrp instance.
- isolate can change effective priorities smoothly, to move instances to the peer machine one by one at a fixed rate while monitoring the status of keepalived.

### DIFF
--- a/tools/json_tracking/isolate
+++ b/tools/json_tracking/isolate
@@ -1,0 +1,172 @@
+#!/usr/bin/env python3
+
+#####
+#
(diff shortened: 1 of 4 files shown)


pqarmitage:
I have merged these scripts into the beta branch, since track_files functionality is not yest available in the master branch.

PR #764, opened 2018-01-26
fix spelling mistakes about keyword promote_secondaries in man page

lxf1992521:


pqarmitage:
Many thanks for this, and apologies for the confusion caused by the typo.

PR #777, opened 2018-02-09
Fix build with Linux kernel headers v4.15

baruchsiach:
Linux kernel version 4.15 changed the libc/kernel headers suppression
logic in a way that introduces collisions:

In file included from ./../include/vrrp_ipaddress.h:32:0,
                 from ./../include/vrrp_arp.h:31,
                 from vrrp.c:29:
/home/peko/autobuild/instance-1/output/host/arc-buildroot-linux-uclibc/sysroot/usr/include/linux/in.h:29:3: error: redeclaration of enumerator 'IPPROTO_IP'
   IPPROTO_IP = 0,  /* Dummy protocol for TCP  */
   ^
/home/peko/autobuild/instance-1/output/host/arc-buildroot-linux-uclibc/sysroot/usr/include/netinet/in.h:33:5: note: previous definition of 'IPPROTO_IP' was here
     IPPROTO_IP = 0,    /* Dummy protocol for TCP.  */
     ^~~~~~~~~~

Include the libc netinet/in.h header first to suppress the kernel
header.

In addition, add _GNU_SOURCE to vrrp.c for the libc provided in6_pktinfo
definition.

Signed-off-by: Baruch Siach <baruch@tkos.co.il>

### DIFF
--- a/keepalived/include/vrrp_arp.h
+++ b/keepalived/include/vrrp_arp.h
@@ -24,6 +24,7 @@
 #define _VRRP_ARP_H
 
 /* system includes */
+#include <netinet/in.h>
 #include <net/ethernet.h>
 #include <net/if_arp.h>
 
--- a/keepalived/vrrp/vrrp.c
+++ b/keepalived/vrrp/vrrp.c
@@ -26,6 +26,7 @@
 #include "config.h"
 
 /* local include */
+#define _GNU_SOURCE
 #include "vrrp_arp.h"
 #include "vrrp_ndisc.h"
 #include "vrrp_scheduler.h"


pqarmitage:
Many thanks for this patch.

Symbol collisions between libc and linux headers have been a problem for a long time, and what we have now in terms of the order of including system header files is a fine balance that seems to work. 

One of the complexities of keepalived is ensuring that keepalived builds with old kernel versions - currently we have tests that go back to 2.6.15, and this means that when moving forwards with new features we have to ensure that it still works with old versions. We also have to work with other libc implementations such as musl. 

I have tried building the current version of keepalived (without this patch) with Linux 4.15 headers and for me it builds without a problem. While I agree that following kernel commit c0bace798436 it would be ideal if we could move to always include libc headers before kernel headers, and to define the relevant _UAPI_DEF_* macros as appropriate, unfortunately that isn't going to work for older kernel headers or older libc implementations.

I will test this patch with a range of kernels, with glibc and musl, and if it works for them I will merge it. Otherwise, it would be good to move forward to including headers the _proper_ way, but with conditional compilation controlled through configure.ac definitions so that older kernels/libc implementations will still work.

rohara:
This is definitely a problem in recent Fedora builds on Rawhide, too, and I've been chasing this down for a few days. Patch is much appreciated.

pqarmitage:
I realise that when I wrote above `I have tried building the current version of keepalived (without this patch) with Linux 4.15 headers and for me it builds without a problem` I was using the beta branch and not v1.4.1. I have now tried compiling with kernel headers version 4.15 and keepalived v1.4.1 and I can reproduce the the problem.

I am currently working through a number of distros to make sure that the change works on all of them, and will then merge the commit.

PR #793, opened 2018-02-28
add reload_advert_int

### LINKED ISSUE #754: reloading with small advert_int cause a switchover
When the slave keepalived is started the old master forces a switchover by going backup, even if the configuration contains nopreempt option, this happens with ka 1.3.9.
Is this the expected behavior?

> Jan 18 11:15:38 localhost Keepalived_vrrp[3222]: VRRP_Instance(VI_1) Received advert with **higher priority 253, ours 253**
> Jan 18 11:15:38 localhost Keepalived_vrrp[3222]: VRRP_Instance(VI_1) Entering BACKUP STATE

rik78:
fixes #754, based on @pqarmitage suggestion.

### DIFF
--- a/doc/keepalived.conf.SYNOPSIS
+++ b/doc/keepalived.conf.SYNOPSIS
@@ -560,6 +560,7 @@ vrrp_instance <STRING> {  # VRRP instance declaration
     virtual_router_id <INTEGER-1..255>  # VRRP VRID
     priority <INTEGER-1..255>  # VRRP PRIO
     advert_int <FLOAT>         # VRRP Advert interval (use default)
+    reload_advert_int <FLOAT>  # VRRP Advert interval during reload (if not specified use advert_int)
 
     lower_prio_no_advert [<BOOL>]  # If a lower priority advert is received, don't
                                    # send another advert. This causes adherence
--- a/keepalived/include/vrrp.h
+++ b/keepalived/include/vrrp.h
@@ -215,6 +215,7 @@ typedef struct _vrrp_t {
 	list			vroutes;		/* list of virtual routes */
 	list			vrules;			/* list of virtual rules */
 	unsigned		adver_int;		/* locally configured delay between advertisements*/
+	unsigned		reload_adver_int; /* locally configured delay between advertisements during a reload */
 	unsigned		master_adver_int;	/* In v3, when we become BACKUP, we use the MASTER's
 							 * adver_int. If we become MASTER again, we use the
 							 * value we were originally configured with.
--- a/keepalived/vrrp/vrrp_daemon.c
+++ b/keepalived/vrrp/vrrp_daemon.c
@@ -410,12 +410,43 @@ vrrp_signal_init(void)
 	signal_ignore(SIGPIPE);
 }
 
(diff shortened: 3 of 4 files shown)


pqarmitage:
@rik78 I've taken your patch and added one extra improvement - keepalived will now also send adverts immediately after a reload for any vrrp instances that are in master state. The reduces the maximum interval between adverts during a reload from 2 * advert_int + reload time to just reload_time. This may well mean that we don't need the `reload_advert_int` config option, and so I haven't incorporated that yet, although I have a patch for that ready to go if needed.

I've also developed a further patch that adds a (configurable) delay after the adverts are sent before the reload commences, since I thought there was a risk that the adverts may be discarded if the socket is closed immediately after the advert is sent. It looks as though that is not the case though.

I'd be grateful if you could try the latest code, including commit 5835fb8 and test if that resolves the problem of the backup taking over as master during a reload. If it doesn't then we can add the `reload_advert_int` patch, but I would prefer not to if we can avoid it, since it complicates the code on the beta branch, which uses pre-build advert packets, and so changing the advert interval means modifying those packets along with the ensuing checksum.

I was going to attach the other two patches to this comment, but unfortunately it wouldn't let me.

rik78:
I can confirm that after a test with
```bash
$ while true; do sudo kill -HUP <keepalive pid>; done
```

I see a lot of adverts from master and no switch happening.
```
16:35:01.466578 IP 192.168.0.10 > 224.0.0.18: VRRPv3, Advertisement
16:35:01.466592 IP 192.168.0.10 > 224.0.0.18: VRRPv3, Advertisement
16:35:01.468164 IP 192.168.0.10 > 224.0.0.22: igmp v3 report, 1 group record(s)
16:35:01.483252 IP 192.168.0.10 > 224.0.0.18: VRRPv3, Advertisement
16:35:01.483312 IP 192.168.0.10 > 224.0.0.18: VRRPv3, Advertisement
16:35:01.486116 IP 192.168.0.10 > 224.0.0.22: igmp v3 report, 1 group record(s)
16:35:01.499449 IP 192.168.0.10 > 224.0.0.18: VRRPv3, Advertisement
16:35:01.499477 IP 192.168.0.10 > 224.0.0.18: VRRPv3, Advertisement
16:35:01.499487 IP 192.168.0.10 > 224.0.0.18: VRRPv3, Advertisement
16:35:01.501110 IP 192.168.0.10 > 224.0.0.22: igmp v3 report, 1 group record(s)
16:35:01.513552 IP 192.168.0.10 > 224.0.0.18: VRRPv3, Advertisement
16:35:01.513591 IP 192.168.0.10 > 224.0.0.18: VRRPv3, Advertisement
16:35:01.516166 IP 192.168.0.10 > 224.0.0.22: igmp v3 report, 1 group record(s)
16:35:01.528617 IP 192.168.0.10 > 224.0.0.18: VRRPv3, Advertisement
16:35:01.528656 IP 192.168.0.10 > 224.0.0.18: VRRPv3, Advertisement
16:35:01.528671 IP 192.168.0.10 > 224.0.0.18: VRRPv3, Advertisement
```

PR #809, opened 2018-03-15
Support parsing of kernel versions containing '+'

sim-:
Checking out an upstream kernel (tagged release) using git, changing
anything, and then building and installing the kernel will result in in a
version string such as "4.15.10+" returned by "uname -r". We can just
treat this like '-', if it comes first.

### DIFF
--- a/keepalived/core/main.c
+++ b/keepalived/core/main.c
@@ -902,7 +902,7 @@ keepalived_main(int argc, char **argv)
 				os_major = 0;
 			else {
 				os_release = (unsigned)strtoul(end + 1, &end, 10);
-				if (*end && *end != '-')
+				if (*end && *end != '-' && *end != '+')
 					os_major = 0;
 			}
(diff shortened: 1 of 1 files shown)


pqarmitage:
@sim- What a coincidence. Only on Wednesday I produced a patch to allow a `.` to terminate the kernel version. I think we need to generalise this, and allow either any punctuation character after the release version, or even any character whatsoever. We will clearly never be able to deal with it being a digit, but as far as I can see there is no restriction on what EXTRAVERSION can be set to.

I think the check therefore should simply be:
```
if (!isdigit(end[1]))
        os_major = 0;
```
Would you be happy with this?

pqarmitage:
This patch has been incorporated into commit 63e9151 which allows further characters other than just '-' and '+' to follow the kernel release number.

Many thanks for submitting the patch which triggered the more general patch.

PR #811, opened 2018-03-19
Improve grammer

yuankunzhang:


pqarmitage:
@yuankunzhang Many thanks for the improvements.

PR #814, opened 2018-03-20
add support for SNI in SSL_GET check

### LINKED ISSUE #804: SSL_GET: SNI support
It seems the [SSL_GET check does not use SNI](https://github.com/acassen/keepalived/blob/master/keepalived/check/check_ssl.c) (no call to `SSL_set_tlsext_host_name`) when checking. This is an issue for cases where the `real_server`s have no sane fallback vhost for connections without SNI.

Any reason why this isn't done by default? If there is, anything against providing an argument to SSL_GET to use it?

Or maybe I'm missing something obvious?

costela:
This adds a `enable_sni` parameter to SSL_GET, making sure the check
passes the virtualhost in the SNI extension during SSL handshake.

It also adds support for SNI to genhash.

fixes #804

### DIFF
--- a/doc/keepalived.conf.SYNOPSIS
+++ b/doc/keepalived.conf.SYNOPSIS
@@ -835,6 +835,10 @@ virtual_server group <STRING>      {  # VS group declaration
                                   #  virtual_server.
         }
 
+        SSL_GET {
+            enable_sni            # send Server Name Indication during SSL handshake
+        }
+
         TCP_CHECK {  # TCP healthchecker
             # No additional options
         }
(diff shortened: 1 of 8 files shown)


pqarmitage:
@costela Many thanks for this patch.

pqarmitage:
@costela I've just added three very minor patches to your SNI code. The first one is just an optimisation for setting up vhost string, the second one is now superseded and the third one uses configure to check for the presence of SSL_set_tlsext_host_name() rather than relying on a specific version of OpenSSL (which I think should probably have been >= 0x10000000L) and also allows it to work with LibreSSL.

costela:
@pqarmitage thanks! All changes make sense! :)

PR #830, opened 2018-04-04
configure.ac: do not force PIE

tpetazzoni:
PIE is not necessarily supported on all architectures, so leave it up
to the user to pass the appropriate CFLAGS/LDFLAGS if he wants to use
PIE.

This fixes the build on the m68k and Microblaze architecture:

  http://autobuild.buildroot.net/results/a536f5947b3b70fdaecad1af5542572c504ad046/
  http://autobuild.buildroot.net/results/0ffbf1e8d181c9463847a5b2be6f9baa18face24/

Signed-off-by: Thomas Petazzoni <thomas.petazzoni@bootlin.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -233,8 +233,8 @@ AC_SUBST(ARFLAGS)
 # AC_PROG_LIBTOOL
 
 KA_CPPFLAGS="$kernelinc"
-KA_CFLAGS="-Wall -Wunused -Wstrict-prototypes -Wextra -g -O2 -fPIE -D_GNU_SOURCE"
-KA_LDFLAGS="-pie"
+KA_CFLAGS="-Wall -Wunused -Wstrict-prototypes -Wextra -g -O2 -D_GNU_SOURCE"
+KA_LDFLAGS=""
 KA_LIBS=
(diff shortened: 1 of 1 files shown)


pqarmitage:
@tpetazzoni I don't want to remove PIE by default where it is supported. Can we instead have a test in configure.ac to check whether PIE is supported, and remove the option if it isn't.

If we have an AC_COMPILE_IFELSE block that has CFLAGS set to -fpie, would that fail on m68k and Microblaze systems?

tpetazzoni:
A compile test with -fPIE doesn't exhibit the issue, however a link test with -pie does. So an AC_LINK_IFELSE block should do the trick, indeed. Thanks for your feedback!

pqarmitage:
@tpetazzoni Are you happy to provide a patch using AC_LINK_IFELSE? I could do so myself but I have no means of testing it on a system that doesn't support -FPIE.

pqarmitage:
@tpetazzoni I've added commit 9bd10d8 which makes a configure time check for whether PIE is supported or not. It would be helpful if you could test this and confirm whether it resolves your issue.

PR #878, opened 2018-05-22
Don't send TSM logs to stdout, but log_message

gmathon-ovh:
Hello,

TSM logs are always sent to stdout. This generates a lot of noise if keepalived is run in foreground.
This change sends TSM log using log_message, only if LOG_DETAIL_BIT is set.

Gautier


pqarmitage:
Many thanks for this patch.

I've added a further commit - d172c4a - that only enables the TSM debug code if the --enable-tsm-debug configure option is specified, since I don't think we need it in the code normally.

PR #903, opened 2018-06-07
The configure script doesn't handle '-L' flags in the output of 'pkg-config --libs'.

iyzsong:
Here is a quick fix.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -29,13 +29,13 @@ AC_DEFUN([add_pkg_config],
     fi
     add_to_var_ind_unique([${KA_PKG_PFX}_CPPFLAGS], [`$PKG_CONFIG --cflags-only-I $1`])
     add_to_var_ind_unique([${KA_PKG_PFX}_CFLAGS], [`$PKG_CONFIG --cflags-only-other $1`])
-    add_to_var_ind_unique([${KA_PKG_PFX}_LIBS], [`$PKG_CONFIG --libs $1`])
+    add_to_var_ind_unique([${KA_PKG_PFX}_LIBS], [`$PKG_CONFIG --libs-only-l $1`])
 
     if test .$3 = .remove-requires; then
       REQUIRES=`$PKG_CONFIG --print-requires $1`
       eval var=\$${KA_PKG_PFX}_LIBS
       for r in $REQUIRES; do
-	REQ_LIBS=`$PKG_CONFIG --libs $r`
+	REQ_LIBS=`$PKG_CONFIG --libs-only-l $r`
 	for l in $REQ_LIBS; do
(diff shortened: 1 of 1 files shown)


pqarmitage:
Could you please explain why we want to strip out the '-L' flags. Further, on my development system no '-L' flags are added by pkg-config --libs.

iyzsong:
Yes, in my case (GuixSD) I have libnl (all other libraries too) installed at  its own prefixed directory.

The output of `pkg-config --libs libnl-3.0` is:
```
-L/gnu/store/prbixm0ldgfj05jv6s3wzmnndmnzg60g-libnl-3.4.0/lib -lnl-3
```
Which makes the `configure.ac`  passing `-L/gnu/store/prbixm0ldgfj05jv6s3wzmnndmnzg60g-libnl-3.4.0/lib nl-3` to the `AC_CHECK_LIB` as `$NL3_LIB`.  The check will fail, and disable the libinl support.

pqarmitage:
I'm concerned that this patch may affect other users who might need the '-L' option.

Would it work to make the change at line 38 of configure.ac, but not make the change at line 32. It seems to me that if `pkg-config --libs` returns a string including '-L' options, that the 'L' option is probably necessary for linking.

iyzsong:
Yes, the library search paths (-L flags) output by `pkg-config` are useful for linking, but keepalived's current configure script parse it with some sed-fu to get the library name, which doesn't work with thoese '-L' flags.

You can test it by adding '-L' to the 'Libs' line in `libnl-3.0.pc`, eg: `Libs: -L/usr/local/lib -lnl-3`, then run `configure`, which should fail the `nl_socket_alloc` checking.

IMO, a better fix would remove the those `AC_CHECK_LIB` calls, and depends on `PKG_CHECK_MODULES` for dependency discovery  (in my OS, only the libmagic.so from the `file` package doesn't come with a pc file),  it can also specify version requirements.

pqarmitage:
@iyzsong I am concerned about removing the '-L' options from KA_LIBS, since where pkg-config --libs includes a '-L' option, it is reasonable to expect that that is required for the linker (I'm not sure why your system doesn't need it given that all your libraries seem to be in different directories).

I've produced a modified configure.ac using your ideas but preserving the '-L' options in KA_FLAGS. I've also taken the opportunity, since I was looking at the code in that area, to tidy up some of how it works.

If you could update this log to confirm whether or not the new configure.ac works for you that would be helpful.

I will have a look at changing from using AC_CHECK_LIBS to PKG_CHK_MODULES, as per your suggestion. My only concern at the moment is that we won't have separate CPPFLAGS and CFLAGS, but I'm not sure if that is an issue.

Many thanks for your input.

PR #943, opened 2018-07-07
Don't create a link-local address for vmac when vmac_xmit_base is set

idosch:
Since commit 18ec95add483 ("Make vmac_xmit_base work for IPv6
instances") VRRP advertisements are sent from the base interface and not
from the vmac interface when vmac_xmit_base is set.

Therefore, there is no need to configure a link-local address on the
vmac interface. This also means that we don't need to regenerate a
link-local address for the vmac if the link-local address was removed
from the base interface, or inherit a link-local address in case one was
configured on the base interface.

Signed-off-by: Ido Schimmel <idosch@mellanox.com>


pqarmitage:
@idosch Many thanks for this patch.

PR #956, opened 2018-07-20
Add newlines to the keepalived.stats output for better readability.

dmbaturin:
At some point, newlines have disappeared from the /tmp/keepalived.stats file generated on SIGUSR2, which made it completele unreadable for machines and hard to read for humans, so I've added the newlines back.

This is what it was like for me in 2.0.5.

````

# cat /tmp/keepalived.stats 
VRRP Instance: Bar  Advertisements:    Received: 0    Sent: 18120  Became master: 1  Released master: 0  Packet Errors:    Length: 0    TTL: 0    Invalid Type: 0    Advertisement Interval: 0    Address List: 0  Authentication Errors:    Invalid Type: 0    Type Mismatch: 0    Failure: 0  Priority Zero:    Received: 0    Sent: 0VRRP Instance: Foo  Advertisements:    Received: 0    Sent: 18120  Became master: 1  Released master: 0  Packet Errors:    Length: 0    TTL: 0    Invalid Type: 0    Advertisement Interval: 0    Address List: 0  Authentication Errors:    Invalid Type: 0    Type Mismatch: 0    Failure: 0  Priority Zero:    Received: 0    Sent: 0

````

### DIFF
--- a/keepalived/vrrp/vrrp_print.c
+++ b/keepalived/vrrp/vrrp_print.c
@@ -70,34 +70,34 @@ vrrp_print_stats(void)
 
 	for (e = LIST_HEAD(l); e; ELEMENT_NEXT(e)) {
 		vrrp = ELEMENT_DATA(e);
-		fprintf(file, "VRRP Instance: %s", vrrp->iname);
-		fprintf(file, "  Advertisements:");
-		fprintf(file, "    Received: %" PRIu64 "", vrrp->stats->advert_rcvd);
-		fprintf(file, "    Sent: %d", vrrp->stats->advert_sent);
-		fprintf(file, "  Became master: %d", vrrp->stats->become_master);
-		fprintf(file, "  Released master: %d",
+		fprintf(file, "VRRP Instance: %s\n", vrrp->iname);
+		fprintf(file, "  Advertisements:\n");
+		fprintf(file, "    Received: %" PRIu64 "\n", vrrp->stats->advert_rcvd);
+		fprintf(file, "    Sent: %d\n", vrrp->stats->advert_sent);
+		fprintf(file, "  Became master: %d\n", vrrp->stats->become_master);
+		fprintf(file, "  Released master: %d\n",
 			vrrp->stats->release_master);
-		fprintf(file, "  Packet Errors:");
-		fprintf(file, "    Length: %" PRIu64 "", vrrp->stats->packet_len_err);
-		fprintf(file, "    TTL: %" PRIu64 "", vrrp->stats->ip_ttl_err);
-		fprintf(file, "    Invalid Type: %" PRIu64 "",
+		fprintf(file, "  Packet Errors:\n");
+		fprintf(file, "    Length: %" PRIu64 "\n", vrrp->stats->packet_len_err);
+		fprintf(file, "    TTL: %" PRIu64 "\n", vrrp->stats->ip_ttl_err);
+		fprintf(file, "    Invalid Type: %" PRIu64 "\n",
 			vrrp->stats->invalid_type_rcvd);
-		fprintf(file, "    Advertisement Interval: %" PRIu64 "",
+		fprintf(file, "    Advertisement Interval: %" PRIu64 "\n",
 			vrrp->stats->advert_interval_err);
-		fprintf(file, "    Address List: %" PRIu64 "",
+		fprintf(file, "    Address List: %" PRIu64 "\n",
 			vrrp->stats->addr_list_err);
-		fprintf(file, "  Authentication Errors:");
-		fprintf(file, "    Invalid Type: %d",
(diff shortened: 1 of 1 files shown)


dmbaturin:
Travis build failed on apt-get install for some reason, but I've tested it on my machine and it seems to work fine.

pqarmitage:
Unfortunately Travis-CI does intermittently throw up these spurious build failures. When it happens I just resubmit the failed build for rebuilding and it usually works.

pqarmitage:
It seems I was rather over enthusiastic when producing commit f702a76 which merged the largely duplicated code for writing configuration after SIGUSR1 and writing the configuration to syslog after startup.

dmbaturin:
@pqarmitage Thanks!
As of Travis, if this happens again, how do I resubmit it? That build was triggered by pull request, I though a user who's not a project member cannot start builds by hand.
Now a bit of offtopic here.

As of duplication, I agree there's a lot of it, nice to see it reduced, I also don't mind to participate (though I've never worked with keepalived source code before so I may need some guidance).
Also, what do you think of making paths to dump stats/data files configurable? I'm not sure how to reconcile it with the way the data dump is triggered by signals though.

As of more immediate tasks, would you mind sync group data in the JSON output, like in the plaintext keepalived.data? Where would you put it and what field names would you use?
Since current JSON output is a list rather than a dict, we need to carefully design the change so that it doesn't break compatibility with people's existing scripts, otherwise I'm ready to work on it.

pqarmitage:
@dmbaturin I'm not sure what privileges are required for resubmitting a build. On a failed build I get a "restart job", but I see that if I am not logged in I don't get that option. I'll try and keep a look out for builds triggered by pull requests that fail and resubmit them.

I'm happy to give guidance for any changes you would like to work on. Regarding making paths to dump stats/data files configurable, I would find that very useful. There are a couple of ways this could be done: either add a command line option to specify the file name, or add an option to global_defs. If you want to use a command line option, have a look at the '-g' option. If I specify -g/tmp/keep.log, it needs to be able to create separate log files for each keepalived process running, i.e. /tmp/keep.log for the parent process, /tmp/keep_vrrp.log, /tmp/keep_check.log and /tmp/keep_bfd.log - see open_log_file() for how that is handled. It also needs to create different file names for different instances of keepalived running (using `-i` option), so if I specify `-i inst1 -g/tmp/keep.log` the log file names will be /tmp/keep_inst1.log /tmp/keep_vrrp_inst1.log etc. The same applies for specifying namespaces with the '-s' option.

My view regarding the JSON output is that anything useful should be able to be added to it. What is included at the moment is what Damien Clabaut decided to include when he implemented it. I know that Alexandre Cassen would like some changes to the JSON code, so it might be worth dropping him an email to get his thoughts. If a dict rather than a list would be better for JSON output, then perhaps that should be implemented as an option.

dmbaturin:
@pqarmitage 
>if I am not logged in I don't get that option
I don't get it even if I'm logged in, probably because only project members can initiate builds (which doesn't sounds unreasonable).

>have a look at the -g option
Thanks, I'll take a look.
So the file names for the state and statistics file should also include instance names?

>If a dict rather than a list would be better for JSON output, then perhaps that should be implemented as an option.
Yes, my main concern is backwards compatibility. An alternative would be to include sync group inside the VRRP instance dict.
Though I'm not sure how big of a concern it really is since I don't know how many people are using those files.

I was also thinking of adding alternatives ways to obtain that data, in addition to signals. Like a UNIX domain socket.
I noticed there's Dbus but it's undocumented (like JSON ;). If you allow me to ask rather than look in the code, if that information supposed to be available through DBus now?

pqarmitage:
@dmbaturin 
>So the file names for the state and statistics file should also include instance names?
The main motivator for adding instance names was to allow multiple instances of keepalived to run simultaneously with file name clashes (e.g. pid files and log files). I think this should also apply to state and stats files.

>Yes, my main concern is backwards compatibility.
I think there are three ways this could be handled.
1. An entry in global_defs to use dict based JSON output
2. Use a command line option to specify it
3. Have a configure option to specify it so that keepalived is built either to produce list base JSON output or dict based output.
I certainly don't like 2. as an option. I think it depends on the overhead in terms of code size whether option 1 or option 3 is better.

>I was also thinking of adding alternatives ways to obtain that data, in addition to signals. Like a UNIX domain socket.
Would you like the JSON output sent back on the socket as well, rather than to a file?

>I noticed there's Dbus but it's undocumented (like JSON ;).
I must make sure that when people submit patches adding functionality like DBus or JSON that they update the documentation as well!

>If you allow me to ask rather than look in the code, if that information supposed to be available through DBus now?
I'm not familiar with the detail of what is available through DBus, but a quick look at vrrp_dbus.c suggests that there is no information returned regarding the status of a sync group.

PR #973, opened 2018-08-13
improve doc spelling

atsevan:


acassen:
Documentation is under work, so most of previous doc will be tagged as deprecated. There is too much divergence between actual Keepalived features and documented ones :/

PR #1021, opened 2018-09-27
Use rbtrees for finding child processes on exit and vrrp_instance when receive advert

pqarmitage:


pqarmitage:
Closing due to an additional rbtree use

PR #1023, opened 2018-10-02
Use more rbtrees and convert some to cached rbtrees

pqarmitage:
Use rbtrees for:
- child pids to improve searching for child process defails when they reminate
- finding vrrp instances on a socket from vrid
- timeouts of vrrp instances on a socket.

Convert  rbtrees that use timeval_t sands as the key to cached rbtrees so that the first node on a tree can be efficiently accessed.


pqarmitage:
I have found a problem with this.

pqarmitage:
The problem doesn't relate to the above changes, but to having a sync group with some members set to nopreempt and others not.

PR #1090, opened 2018-12-18
Also skip route not configured with down interface

clan:
Otherwise, if keepalived has virtual_routes configured, we create
a virtual interface and bring it up and down, current code will bring
VRRP state to FAULT and never return.

```
 # ip tun add test mode ipip remote 10.0.0.1 local 10.0.0.2
 # ip link set test up
 # ip link set test down
```

### DIFF
--- a/keepalived/vrrp/vrrp_if.c
+++ b/keepalived/vrrp/vrrp_if.c
@@ -1051,7 +1051,9 @@ interface_down(interface_t *ifp)
 			/* Any route that has an oif will be tracking the interface,
 			 * so we only need to check for routes that dont specify an
 			 * oif */
-			if (!route->oif && route->configured_ifindex != ifp->ifindex)
+			/* Don't track route if it's not configured with this down
+			 * interface. */
+			if (!route->oif || route->configured_ifindex != ifp->ifindex)
 				continue;
 
 			route->set = false;


pqarmitage:
I don't think the patch is quite correct:
```
			if (!route->oif || route->configured_ifindex != ifp->ifindex)
				continue;
```
means that any virtual route without an oif configured will not be tracked, whereas if eth0 has address 10.1.0.1/24, a route specified as 10.2.0.0/24 via 10.1.0.2 will be configured on eth0, and if that route is deleted then the vrrp instance should go to fault state.

Could you please provide a copy of your keepalived configuration so that we can test what is actually happening in your case.

clan:
Yes, my fix might not correct in all case. 
For my case, as I have described in the commit message, just add a virtual_routes config, then create a virtual interface, bring it up, then down. Then keepalived will run into FAULT state.

Could you please provide a example config (or describe) for your case, so I can test and fix this pull request?

pqarmitage:
I've had a further look at the code, and it appears that if the oif for a route is not specified, then the route cannot be tracked. That being the case I think your patch is OK, but I just want to think it over a bit more before merging it.

clan:
Are there any possible side effects for this pr? :-)

PR #1126, opened 2019-02-14
Make it possible to specify decimal value for delay_loop, connect_timeout, delay_before_retry.

tamutamu:
To shorten the real server monitoring interval, I have make it possible to sepcify decimal value for following items.
- delay_loop
- connect_timeout
- delay_before_retry

```
virtual_server 192.168.1.200 80 {
  delay_loop 0.5 

    real_server 192.168.1.11 80 {
    ...
    TCP_CHECK {
      connect_timeout 0.4
      delay_before_retry 0,8
    ...
```

### DIFF
--- a/doc/man/man5/keepalived.conf.5
+++ b/doc/man/man5/keepalived.conf.5
@@ -1467,7 +1467,7 @@ The syntax for virtual_server is :
 \fBvirtual_server fwmark \fR<INTEGER> |
 \fBvirtual_server group \fR<STRING> {
     # delay timer for checker polling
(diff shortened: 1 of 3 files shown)


pqarmitage:
@tamutamu Many thanks for submitting this patch. There were one or two very small omissions from the patch (some of the parameter parsing for the virtual_server and real_server blocks weren't updated, and the config dump and SNMP outputs needed updating for the fractional parts of the times).

I have merged commits 44c6ea3 and ebd3712 which are entirely based on your patch, but with the missing bits added.

Once again, many thanks for your patch.

PR #1128, opened 2019-02-15
It enable to set zero value as mintime for delay_loop and connect_timeout.

tamutamu:
It enable to set greater than or equal to zero value as Mintime for delay_loop, connect_timeout.

### DIFF
--- a/keepalived/check/check_api.c
+++ b/keepalived/check/check_api.c
@@ -301,7 +301,7 @@ co_timeout_handler(vector_t *strvec)
 	conn_opts_t *co = CHECKER_GET_CO();
 	unsigned long timer;
 
-	if (!read_timer(strvec, 1, &timer, TIMER_MIN_SEC, UINT_MAX / TIMER_HZ, true)) {
+	if (!read_timer(strvec, 1, &timer, 1, UINT_MAX, true)) {
 		report_config_error(CONFIG_GENERAL_ERROR, "connect_timeout %s invalid - ignoring", FMT_STR_VSLOT(strvec, 1));
 		return;
 	}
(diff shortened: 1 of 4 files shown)


tamutamu:
> vs_co_timeout_handler() and rs_co_timeout_handler() in check_parser.c need to be updated.

Sorry, I missed it.  
f345acbae025e6efd54efb95ecdc3eea708ca8a0

> it is reasonable to say that the minimum is 0.000001.

That is true for sure.
I changed min value `0` to '0.000001 (1 micro-second)' by using TIMER_MIN_SEC defined in lib\timer.h.

pqarmitage:
@tamutamu Many thanks for the update. I've had a quick look at it and it is looking must better now. I'll have a proper look at it later, and if everything is OK I'll merge the commits.

tamutamu:
@pqarmitage Thank you for your review.

pqarmitage:
@tamutamu Many thanks for your work on this, which I have now merged.

There was one small issue with your patches, which for the sake of not continuing going back and forth I have added a further patch to resolve. The issue is that the calculation:
`min_time / TIMER_HZ`
produces 0 when min_time is less than TIMER_HZ, so 1 / TIMER_HZ == 0 which we were specifically trying to avoid; even (TIMER_HZ - 1)/TIMER_HZ == 0. The problem with division of integer types is that the values are truncated. When this is converted to a double as a parameter to the call of read_strvec_double(), the 0 integer value is converted to a double, but is still 0.

Commit 85b5ed2 resolves this issue by converting min_time and max_time to doubles before dividing by TIMER_HZ, thereby ensuring that the fractional parts of the results are preserved.

I have discovered an additional gcc warning flag re converting doubles/floats to integer types (as I identified in comment https://github.com/acassen/keepalived/pull/1128#pullrequestreview-204541281 above). The flag is _-Wfloat-conversion_ and it would have produced a warning when a parameter of 0.000001 (TIMER_MIN) was passed to read_timer() when it was expecting an unsigned long.

I will add _-Wfloat-conversion_ as a standard build warning, once I have fixed the three warnings that it currently produces.

PR #1141, opened 2019-02-26
Fix parse error when semicolon in the end of Scripts.

HW-lj:
Fix parse error when semicolon in the end of Scripts.
Such as config following script in virtual_server,
    quorum_up "/usr/sbin/ip addr add 192.168.122.15/32 dev lo;" 
which will lead to the failure of adding the ip address.
So we should consider ";" as part of WHITE_SPACE_STR.

### DIFF
--- a/lib/parser.c
+++ b/lib/parser.c
@@ -60,7 +60,7 @@
 
 #define BOB "{"
 #define EOB "}"
-#define WHITE_SPACE_STR " \t\f\n\r\v"
+#define WHITE_SPACE_STR " \t\f\n\r\v;"
 
 typedef struct _defs {
 	char *name;


pqarmitage:
I would suggest that having the semicolon at the end of the script is a configuration error. The solution is to remove the ';' from the configuration.

If keepalived is mishandling the semicolon at the end of the configuration, then that is what needs to be fixed, rather than allowing (and ignoring) the semicolon.

The change as proposed would mean that:
`advert_int  ;;;2`
would be parsed as though
`advert_int 2`
had been configured.

HW-lj:
Yep, the solution may be not good enough, I would think it again.

pqarmitage:
So far as I can see, the parser is working correctly. If a semicolon is specified, it is included as part of the string. For example:
```
    MISC_CHECK {
      misc_path "/etc/keepalived/scripts/vs.sh RS1;"
      misc_timeout 15
      }
```
```
Keepalive method = MISC_CHECK
script = '/home/quentin/keepalived/utils/configs/scripts/vs.sh' 'RS1;'
```
which is precisely what has been specified.

I think the answer is: if you don't want the semicolon, don't configure it! :)

HW-lj:
OK, it's fine. :(

pqarmitage:
@HW-lj Many thanks for your support of keepalived. Should you find any other issues with keepalived please do report them, or better still provide patches to fix them.

Issue #48, opened 2013-11-14
Setting of default route broken in keepalived 1.2.8 and 1.2.9


tkurbad:
Hi,

I've been using keepalived 1.2.7 successfully for over a year now to implement VRRP on our primary and backup firewalls.

My setup therefore includes sections like:

```
virtual_routes {
    default via 1.2.3.4 dev vrrp_extern_0
}
```

After upgrading to keepalived 1.2.8 and, subsequently, to 1.2.9, these route definitions stopped working, with keepalived complaining about `default` not being a valid ip-address.

I replaced the word `default` by `0.0.0.0/0` only to realize that there seems to be a bug in prefix handling, too. In fact, according to /proc/net/route, the route set by keepalived was `0.0.0.0/32`, thus being a blackhole instead of a default route definition.

I was able to circumvent the problem by setting up `notify_master` and `notify_backup` scripts that set and remove the necessary route entries by invoking the `ip route` command. However, I consider the above behavior as being a serious bug that needs to be fixed.

Best,
Torsten

Issue #44, opened 2013-10-23
Unicast across different subnets


gudmundur:
Hi,

I'm trying to set up keepalived with unicast on Amazon VPC, and within a subnet, it works like a charm. To be a bit more fault tolerant, I've set up multiple availability zones, which means that I need to place the machines in separate subnets. Now it doesn't work anymore.

The networks I have are 10.0.0.0/24 and 10.0.1.0/24 and traffic flows just fine between them. TCP, UDP, ICMP and VRRP packets are being routed.

When keepalived sends its advert, tcpdump shows a ARP who-has request, which doesn't resolve as the IP being asked for is on a separate subnet. As far as I can tell, the routing tables I have, indicate that the packets should be routed through the gateway, but it doesn't seem to be happening when the packet originates from keepalived.

Am I misunderstanding the unicast support? Shouldn't it work across networks?

cheers,
Guðmundur Bjarni

PR #1149, opened 2019-03-08
fix keepalived deadlock on signal_handler

l00478643:
It's unsafe to use syslog in signal_handler,it may cause deadlock:
#0  0x00007f7d1242248c in __lll_lock_wait_private () from /lib64/libc.so.6
#1  0x00007f7d1239e3cd in _L_lock_15626 () from /lib64/libc.so.6
#2  0x00007f7d1239b4d3 in malloc () from /lib64/libc.so.6
#3  0x00007f7d1238da5a in open_memstream () from /lib64/libc.so.6
#4  0x00007f7d1240e32a in __vsyslog_chk () from /lib64/libc.so.6
#5  0x00007f7d1240e922 in __syslog_chk () from /lib64/libc.so.6
#6  0x000055691ddd60bf in syslog (__fmt=0x55691dddeb6f "%s", __pri=6) at /usr/include/bits/syslog.h:31
#7  vlog_message (facility=6, format=<optimized out>, args=args@entry=0x7ffe4c1b8900) at logger.c:50
#8  0x000055691ddd6184 in log_message (facility=facility@entry=6, 
    format=format@entry=0x55691dde5d50 "BUG - write to signal_pipe[1] error %s - please report") at logger.c:59
#9  0x000055691ddd5805 in signal_handler (sig=1) at signals.c:100
#10 <signal handler called>
#11 0x00007f7d1242248a in __lll_lock_wait_private () from /lib64/libc.so.6
#12 0x00007f7d1239e3cd in _L_lock_15626 () from /lib64/libc.so.6
#13 0x00007f7d1239b4d3 in malloc () from /lib64/libc.so.6
#14 0x00007f7d1238da5a in open_memstream () from /lib64/libc.so.6
#15 0x00007f7d1240e32a in __vsyslog_chk () from /lib64/libc.so.6
#16 0x00007f7d1240e922 in __syslog_chk () from /lib64/libc.so.6
#17 0x000055691ddd60bf in syslog (__fmt=0x55691dddeb6f "%s", __pri=6) at /usr/include/bits/syslog.h:31
#18 vlog_message (facility=6, format=<optimized out>, args=args@entry=0x7ffe4c1b9380) at logger.c:50
#19 0x000055691ddd6184 in log_message (facility=facility@entry=6, 
    format=format@entry=0x55691dde5d50 "BUG - write to signal_pipe[1] error %s - please report") at logger.c:59
#20 0x000055691ddd5805 in signal_handler (sig=17) at signals.c:100
#21 <signal handler called>
#22 0x00007f7d12398de4 in _int_malloc () from /lib64/libc.so.6
#23 0x00007f7d1239bf04 in calloc () from /lib64/libc.so.6
#24 0x00007f7d13e5f409 in __nlmsg_alloc () from /lib64/libnl-3.so.200
#25 0x00007f7d13e5f846 in nlmsg_convert () from /lib64/libnl-3.so.200
#26 0x00007f7d13e61daf in nl_recvmsgs_report () from /lib64/libnl-3.so.200
#27 0x00007f7d13e624a9 in nl_recvmsgs () from /lib64/libnl-3.so.200
#28 0x00007f7d13e62511 in nl_wait_for_ack () from /lib64/libnl-3.so.200
#29 0x00007f7d14073ad0 in genl_ctrl_probe_by_name () from /lib64/libnl-genl-3.so.200
#30 0x00007f7d14073ceb in genl_ctrl_resolve () from /lib64/libnl-genl-3.so.200
#31 0x000055691dda0b55 in ipvs_nl_send_message (msg=msg@entry=0x556924f077a0, 
    func=func@entry=0x55691dda0b00 <ipvs_nl_noop_cb>, arg=arg@entry=0x0) at libipvs.c:302
#32 0x000055691dda515f in ipvs_add_laddr (svc=0x55692fc2ac10, laddr=0x5569359d34e0) at libipvs.c:930
#33 0x000055691dd9d273 in ipvs_talk (cmd=cmd@entry=1168, daemonrule=daemonrule@entry=0x0, 
    ignore_error=ignore_error@entry=false) at ipvswrapper.c:255
#34 0x000055691dd9d614 in ipvs_laddr_group_cmd (cmd=cmd@entry=1168, laddr_group=laddr_group@entry=0x556932f1b4e0)
    at ipvswrapper.c:626
#35 0x000055691dd9f24a in ipvs_laddr_cmd (vs=0x55692fc2ac10, cmd=1168) at ipvswrapper.c:804
#36 ipvs_cmd (cmd=cmd@entry=1168, vs=vs@entry=0x556931f77e70, rs=rs@entry=0x0) at ipvswrapper.c:874
#37 0x000055691dd9b610 in init_service_laddr (vs=0x556931f77e70) at ipwrapper.c:672
#38 init_service_vs (vs=0x556931f77e70) at ipwrapper.c:762
#39 0x000055691dd9baa1 in init_services () at ipwrapper.c:815
#40 0x000055691dd8dc45 in start_check () at check_daemon.c:277
---Type <return> to continue, or q <return> to quit---
#41 0x000055691dd8dfe5 in reload_check_thread (thread=<optimized out>) at check_daemon.c:399
#42 0x000055691ddd3bde in thread_call (thread=0x7ffe4c1ba010) at scheduler.c:963
#43 launch_scheduler () at scheduler.c:986
#44 0x000055691dd8e1af in start_check_child () at check_daemon.c:522
#45 0x000055691dd89340 in start_keepalived () at main.c:337
#46 keepalived_main (argc=2, argv=<optimized out>) at main.c:1064
#47 0x00007f7d12338455 in __libc_start_main () from /lib64/libc.so.6
#48 0x000055691dd879fe in _start ()

I wonder if we can use fprintf instead of log_message ? fprintf is safe in signal handler.
-------------------------------------------------

Signed-off-by: Lin Miaohe <linmiaohe@huawei.com>
--

### DIFF
--- a/lib/signals.c
+++ b/lib/signals.c
@@ -181,10 +181,8 @@ static void
 signal_handler(int sig)
 {
 	if (write(signal_pipe[1], &sig, sizeof(int)) != sizeof(int)) {
-		DBG("signal_pipe write error %s", strerror(errno));
 		assert(0);
-
-		log_message(LOG_INFO, "BUG - write to signal_pipe[1] error %s - please report", strerror(errno));
+		fprintf(stderr, "BUG - write to signal_pipe[1] error %s - please report\n", strerror(errno));
 	}
 }
 #endif


lib/signals.c
```diff
@@ -181,10 +181,8 @@ static void
 signal_handler(int sig)
 {
 	if (write(signal_pipe[1], &sig, sizeof(int)) != sizeof(int)) {
-		DBG("signal_pipe write error %s", strerror(errno));
 		assert(0);
-
-		log_message(LOG_INFO, "BUG - write to signal_pipe[1] error %s - please report", strerror(errno));
+		fprintf(stderr, "BUG - write to signal_pipe[1] error %s - please report\n", strerror(errno));
```
pqarmitage: We can't use fprintf, since stdout and stderr are not open (or are opened on /dev/null), and keepalived, unless specifically run in foreground, has no controlling terminal to which it could write.

lib/signals.c
```diff
@@ -181,10 +181,8 @@ static void
...
 	}
 }
 #endif
```
pqarmitage: Could you please post the output of `keepalived -v`. Since Linux 2.6.22 (and glibc 2.8) keepalived doesn't use signal handlers.

pqarmitage:
According to the GNU libc documentation, fprintf() is AS-Unsafe, so would appear to be no safer to use than syslog(), which is also AS_Unsafe

pqarmitage:
I think a patch along the following lines is probably the best way to resolve this:
```
diff --git a/lib/logger.c b/lib/logger.c
index a47d50e2..cf7e8075 100644
--- a/lib/logger.c
+++ b/lib/logger.c
@@ -29,6 +29,9 @@
 #include <fcntl.h>
 #include <string.h>
 #include <memory.h>
+#ifndef HAVE_SIGNALFD
+#include <signal.h>
+#endif
 
 #include "logger.h"
 #include "bitops.h"
@@ -151,12 +154,35 @@ vlog_message(const int facility, const char* format, va_list args)
 #endif
 	}
 
-	if (!__test_bit(NO_SYSLOG_BIT, &debug))
+	if (!__test_bit(NO_SYSLOG_BIT, &debug)) {
+#ifndef HAVE_SIGNALFD
+		sigset_t cur_set, block_set;
+		bool dont_restore = false;
+
+		sigfillset(&block_set);
+		if (sigprocmask(SIG_BLOCK, &block_set, &cur_set)) {
+			/* Yes, we are logging without disabling signals,
+			 * but it would be useful to know that this has
+			 * failed. The only error that could occur according
+			 * to sigprocmask(2) is EFAULT, which would be very
+			 * strange since the sigsets are on the stack. */
+			syslog(facility, "%s", "sigprocmask failed in vlog_message");
+
+			dont_restore = true;
+		}
+
+#endif
 #if HAVE_VSYSLOG
 		vsyslog(facility, format, args);
 #else
 		syslog(facility, "%s", buf);
 #endif
+
+#ifndef HAVE_SIGNALFD
+		if (!dont_restore)
+			sigprocmask(SIG_SETMASK, &cur_set, NULL);
+#endif
+	}
 }
 
 void
```

Please could you advise whether you think this is the right solution, or whether it should be done another way, or whether the patch can be improved.

It would also be helpful if you could explain, so that we can understand the issue and potentially provide a better solution, why code is being included that is in an `#ifndef HAVE_SIGNALFD`, since we are not aware of any use of kernels older than 2.6.32 (CentOS 6.5) and signalfd() has been available since 2.6.22, and we are considering dropping support for older kernels (but have not yet decided what is the oldest kernel version we should support).

lib/signals.c
l00478643: My kernel is base on linux-3.10,but keepalived is still old version:
[root@localhost linux]# keepalived -v
Keepalived v1.3.5 (03/19,2017), git commit v1.3.5-6-g6fa32f2

Copyright(C) 2001-2017 Alexandre Cassen, <acassen@gmail.com>

Build options:  PIPE2 LIBNL3 RTA_ENCAP RTA_EXPIRES RTA_PREF FRA_OIFNAME FRA_TUN_ID RTAX_CC_ALGO RTAX_QUICKACK LIBIPTC LIBIPSET_DYNAMIC LVS LIBIPVS_NETLINK IPVS_64BIT_STATS VRRP VRRP_AUTH VRRP_VMAC SOCK_NONBLOCK SOCK_CLOEXEC FIB_ROUTING INET6_ADDR_GEN_MODE SNMP_V3_FOR_V2 SNMP SNMP_KEEPALIVED SNMP_CHECKER SNMP_RFC SNMP_RFCV2 SNMP_RFCV3 SO_MARK

[root@localhost linux]# uname -a
Linux localhost.localdomain 3.10.0-862.14.0.1.h43.eulerosv2r7.x86_64 #6 SMP Fri Dec 28 08:27:17 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

pqarmitage:
I would strongly advise you to upgrade to keepalived v2.0.13. There have been in excess of 1600 commits since the commit you are using, and these comprise bug fixes, feature enhancements, performance enhancements etc. These updates include starting to use signalfd() so that there are no asynchronous signal handlers, which would mean that the problem that you produced the patch for no longer exists.

l00478643:
Thanks for your patch . It looks really good and it could resolve the deadlock I found in that version. But since fprintf is also AS-Unsafe, maybe you need to block signal before fprintf is called in vlog_message.
I think a patch according to your patch:
```
diff --git a/lib/signals.c b/lib/signals.c
index 2b96bcc1..f70da1b1 100644
--- a/lib/signals.c
+++ b/lib/signals.c
@@ -181,10 +181,26 @@ static void
 signal_handler(int sig)
 {
        if (write(signal_pipe[1], &sig, sizeof(int)) != sizeof(int)) {
-               DBG("signal_pipe write error %s", strerror(errno));
+               sigset_t cur_set, block_set;
+               bool dont_restore = false;
                assert(0);
 
+               sigfillset(&block_set);
+               if (sigprocmask(SIG_BLOCK, &block_set, &cur_set)) {
+                       /* Yes, we are logging without disabling signals,
+                        * but it would be useful to know that this has
+                        * failed. The only error that could occur according
+                        * to sigprocmask(2) is EFAULT, which would be very
+                        * strange since the sigsets are on the stack. */
+                       syslog(facility, "%s", "sigprocmask failed in vlog_message");
+
+                       dont_restore = true;
+               }
+
                log_message(LOG_INFO, "BUG - write to signal_pipe[1] error %s - please report", strerror(errno));
+
+               if (!dont_restore)
+                       sigprocmask(SIG_SETMASK, &cur_set, NULL);
        }
}
#endif
```

Could you please advise whether you think this is the right solution?
Or maybe we could ignore this deadlock problem, since it is only happened in the old version . And it will make code looks a little ugly if we fix it this way.
I am looking forward to your reply.

l00478643:
> I would strongly advise you to upgrade to keepalived v2.0.13. There have been in excess of 1600 commits since the commit you are using, and these comprise bug fixes, feature enhancements, performance enhancements etc. These updates include starting to use signalfd() so that there are no asynchronous signal handlers, which would mean that the problem that you produced the patch for no longer exists.

I have upgrade to keepalived v2.0.13. And I enjoy the many improvment that you worked out.Thanks for your effort.
So we may just ignore the deadlock problem since it no longer exists.

pqarmitage:
@l00478643 Thanks for your comment re fprintf() calls. I'll update the patch to include that, since we ought to have it in case anyone is using a system without signalfd(). Now that you have upgraded to 2.0.13, the patch will no longer affect you.

Many thanks for identifying the issue; there might just be one or two users out there who will benefit.

l00478643:
> @l00478643 Thanks for your comment re fprintf() calls. I'll update the patch to include that, since we ought to have it in case anyone is using a system without signalfd(). Now that you have upgraded to 2.0.13, the patch will no longer affect you.
> 
> Many thanks for identifying the issue; there might just be one or two users out there who will benefit.

It‘s my pleasure to identify the issue with you. Many thanks for your effort to improve the keepalived.

pqarmitage:
Commit b5ad7f3 blocks signals when syslog is called if async signal handlers are used.

@l00478643 Many thanks for the problem and your help with this.

l00478643:
> Commit [b5ad7f3](https://github.com/acassen/keepalived/commit/b5ad7f318429282b62144d24f1ba4eccd2d649ae) blocks signals when syslog is called if async signal handlers are used.
> 
> @l00478643 Many thanks for the problem and your help with this.

It‘s my pleasure.

PR #1156, opened 2019-03-12
fix OPENSSL_init_crypto init option

HW-lj:
we should automatic loading of the libcrypto error strings in OPENSSL_init_crypto init  .


pqarmitage:
Many thanks for this patch. We have now incorporated this change, but still use OPENSSL_INIT_NO_LOAD_CONFIG if it works.

PR #1162, opened 2019-03-15
Add more compiler warnings and various minor fixes

pqarmitage:


lib/list_head.h
```diff
@@ -152,47 +152,47 @@ static inline void list_del_init(struct list_head *entry)
 
 /**
  * list_move - delete from one list and add as another's head
- * @list: the entry to move
+ * @LIST: the entry to move
  * @head: the head that will precede our entry
  */
-static inline void list_move(struct list_head *list, struct list_head *head)
+static inline void list_move(struct list_head *LIST, struct list_head *head)
```
acassen: hmm, just 'l' instead of LIST. Cosmetics consideration, but dont like upper case.

pqarmitage: I don't know what made me use upper case. I've changes LIST to lst, since I don't like since letter variable names except as loop control variables.

PR #1164, opened 2019-03-15
fix unknow init checker state influences the result of correct health…

HW-lj:
This commit will fix two issue of acassen#1163 and acassen#1144


pqarmitage:
This pull request is superseded by pull request #1165. Please continue testing and reporting issues and providing patches. It is all improving keepalived.

HW-lj:
> Unfortunately this patch is not sufficient for all cases. For example if a reload occurs when a checker is failed, but the new config has no checkers for the real server, then the real server will not be re-added.
> 
> I am currently working on a patch that should cover all scenarios.


under this circumstance, it seems better to let the real server down as it has been failed state before, though it means the  real server will not be re-added.

PR #1180, opened 2019-03-26
fix add new rs into exist vs with alpha online

HW-lj:
When we config a virtual server and a real server with alpha [real server is unreachable], and then we add a new real server into the virtual server and reload it, the new added  real server will be online for a while.

### DIFF
--- a/keepalived/check/check_data.c
+++ b/keepalived/check/check_data.c
@@ -934,7 +934,7 @@ bool validate_check_config(void)
 
 		/* In Alpha mode also mark any checker that hasn't run as failed.
 		 * Reloading is handled in migrate_checkers() */
-		if (!reload) {
+		if (!checker->has_run) {
 			if (checker->alpha) {
 				set_checker_state(checker, false);
 				UNSET_ALIVE(checker->rs);


HW-lj:
Now, we have handled the case of inheritance in our migrate_checkers function, while the new real server with alpha would be handled in init_service_rs wrongly.

pqarmitage:
There is a question regarding what should happen when reloading a config which adds an alpha mode checker. When the new alpha mode checker is for a new real server, then the real server should start in the down state. However, if it is a new checker for an existing real server, then if the real server was up before the reload, it should not be brought down while waiting for the new alpha mode checker to run; it is better to leave the real server up until the new alpha mode checker fails (i.e. on reload when adding an alpha mode checker to a real server that is up, ignore the alpha mode).

In the light of the above I will test the patch to make sure it still works as set out above.

pqarmitage:
Commit e845645 extends the idea of this patch to also ensure that new alpha mode checkers don't cause an existing real server to be brought down at reload.

HW-lj:
Many thanks for your extention. Indeed, I think my patch can work well in the case you worried about. For the  existing real server, we do it well in migrate_checkers function. The new alpha mode checkers don't cause an existing real server to be brought down at reload, never.

pqarmitage:
Your patch doesn't work if you have one checker, which has failed, then add another alpha mode checker and restart; the real server get added to the virtual server after the new checker is successful, but since the original checker is still failed, the real server should not be added.

There is clearly something not quite right in the code, which your patch somehow relies on in order to work, but my patch somehow works around.

I suspect that with the current patch the error in the code is hidden and not causing a problem, but I will attempt to find the issue and resolve it. Once it is fixed, if an alpha mode checker is added at a restart, the real server will be brought down until the new checker is successful.

HW-lj:
Thanks for your patience. Now  I have got what you mean really.

PR #1199, opened 2019-04-03
add udp health check

HW-lj:
Hi, I have developed a UDP_CHECK to replace the way of using MISC_CHECK and scripts, which will greatly improve performance.
Meanwhile, we should do something to let it work properly. 
1. We must config the ping_group_range.
	`echo 0 0 > /proc/sys/net/ipv4/ping_group_range`
2. If we config a real server in several virtual server, the icmp_ratelimit should be canceled.
	`echo 0 > /proc/sys/net/ipv4/icmp_ratelimit`
	
PS: Due to some special reasons, someone may think ping is not safe enough, so I add a switch to let user choose whether to use ping or not.
Of course, if we do not use ping, then when the real server unexpectedly shutdown, we will consider it healthy :(

At last is an simple example config:
```
real_server 33.33.110.103 5001 {
    weight 2
    UDP_CHECK {
		connect_port 5001
		connect_timeout 15
		retry 3
		delay_before_retry 3
		#ping_check_off
	}
}
```

### DIFF
--- a/keepalived/check/Makefile.am
+++ b/keepalived/check/Makefile.am
@@ -16,7 +16,7 @@ libcheck_a_SOURCES = \
 	check_daemon.c check_data.c check_parser.c \
 	check_api.c check_tcp.c check_http.c check_ssl.c \
 	check_smtp.c check_misc.c check_dns.c check_print.c \
-	ipwrapper.c ipvswrapper.c libipvs.c
+	ipwrapper.c ipvswrapper.c libipvs.c check_udp.c check_ping.c
 
 AM_CPPFLAGS		+= -I$(srcdir)/../include -I$(srcdir)/../../lib
 
--- a/keepalived/check/check_api.c
+++ b/keepalived/check/check_api.c
@@ -48,6 +48,7 @@
 #include "bfd_event.h"
 #include "bfd_daemon.h"
 #endif
+#include "check_udp.h"
 
 /* Global vars */
 list checkers_queue;
@@ -677,6 +678,7 @@ install_checkers_keyword(void)
 	install_misc_check_keyword();
 	install_smtp_check_keyword();
 	install_tcp_check_keyword();
+	install_udp_check_keyword();
 	install_http_check_keyword();
 	install_ssl_check_keyword();
 	install_dns_check_keyword();
--- a/keepalived/check/check_api.c
+++ b/keepalived/check/check_api.c
@@ -155,6 +155,11 @@ queue_checker(void (*free_func) (void *), void (*dump_func) (FILE *, void *)
 
 	/* queue the checker */
 	list_add(checkers_queue, checker);
+	if(compare) {
+		if(rs->samecheckers == NULL)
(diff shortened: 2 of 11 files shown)


pqarmitage:
Many thanks for your work adding the new checker. There are a few changes though that are needed to the patch before it should be merged.

I will work through the patch and add comments in the next day or so.

HW-lj:
> Many thanks for your work adding the new checker. There are a few changes though that are needed to the patch before it should be merged.
> 
> I will work through the patch and add comments in the next day or so.

The commits b1873fe is nothing with udp_check, this one is just to show my some idea how to improve the migration. I add a list into real server structure to record the checkers of the same real server to decrease some useless comparison，while it may not good enough, so could you do it better to lower the complexity of migration.

acassen:
Hello, any write operations on fd MUST be submitted to I/O MUX otherwise you can introduce sync point that can introduce side effects on global scheduling.

HW-lj:
> Hello, any write operations on fd MUST be submitted to I/O MUX otherwise you can introduce sync point that can introduce side effects on global scheduling.

Sorry, I don't seem to understand what you mean. Could you elaborate on it? Thanks.

HW-lj:
Hey, guys. How long will it take for you to process this patch? I have some other bugfix patch to pull : )

pqarmitage:
There are a number of issues with this patch that mean it cannot be merged in its current form.

1. As has already been said, read and write operation must be handled via the I/O MUX, otherwise keepalived could stall while waiting for I/O operations to complete. In particular you will need to use thread_add_write() if when you attempt to send on a socket you get an EWOULDBLOCK error.
2. Don't duplicate existing code that can be re-used: check_dns.c has code for setting up UDP sockets (dns_connect_thread() etc). That code should be used, probably renamed udp_connect_thread(), and moved into layer4.c.
3. Any sockets created must have O_NONBLOCK and O_CLOEXEC set
4. ICMP is not a protocol running on top of UDP. ICMP is a protocol directly above IP/IPv6.
5. To be generally useful, the UDP check shouldn't expect to receive a UDP packet in response, but should simply ensure that there isn't an ICMP port unreachable response. There could be an option to say expect to receive a number of bytes in reply.
6. Don't use getprotobyname, use IPPROTO_ICMP etc.
7. Don't use thread_add_timer with a timeout of 0, use thread_add_event instead.
8. udp_check_t ping_check should be a bool, and use true/false with it.
9. In ping_it() the echo.sequence number will always be 1. Also, echo.id will frequently have the same value.
10. Why does SO_RCVBUF need to be set so large?
11. Error and other messages need to be corrected (e.g. UDP rather than TCP).

The above are examples of how you patch doesn't match the coding standards and practices of keepalived. I you can update your patch in line with the comments above, we will look at it again with a view to merging it.

If you have other independent patches that you would like merged, it is probably best to create them on different branches, so that they can be considered separately from each other.

HW-lj:
Thank you for your detailed advice. I will fix these as soon as I can.

PR #1246, opened 2019-04-27
fill srule correctly

HW-lj:
When we remove  vsg entry, we should fill the address. Otherwise ipvs will return errno=2.

### DIFF
--- a/keepalived/check/ipvswrapper.c
+++ b/keepalived/check/ipvswrapper.c
@@ -639,8 +639,14 @@ ipvs_group_sync_entry(virtual_server_t *vs, virtual_server_group_entry_t *vsge)
 				/* Talk to the IPVS channel */
 				ipvs_talk(IP_VS_SO_SET_ADDDEST, &srule, &drule, NULL, false);
 			}
-			else
+			else {
+				if (vsge->addr.ss_family == AF_INET6)
+					inet_sockaddrip6(&vsge->addr, &srule.nf_addr.in6);
+				else 
+					srule.nf_addr.ip = inet_sockaddrip4(&vsge->addr);
+
 				ipvs_group_range_cmd(IP_VS_SO_SET_ADDDEST, &srule, &drule, vsge);
+			}
 		}
 	}
 }
@@ -682,8 +688,14 @@ ipvs_group_remove_entry(virtual_server_t *vs, virtual_server_group_entry_t *vsge
 	if (!is_vsge_alive(vsge, vs)) {
 		if (vsge->range)
 			ipvs_group_range_cmd(IP_VS_SO_SET_DEL, &srule, NULL, vsge);
-		else
+		else {
+			if (vsge->addr.ss_family == AF_INET6)
+				inet_sockaddrip6(&vsge->addr, &srule.nf_addr.in6);
+			else 
+				srule.nf_addr.ip = inet_sockaddrip4(&vsge->addr);
+
 			ipvs_talk(IP_VS_SO_SET_DEL, &srule, NULL, NULL, false);
+		}
 	}
 }
 
--- a/keepalived/check/ipvswrapper.c
+++ b/keepalived/check/ipvswrapper.c
@@ -640,11 +640,6 @@ ipvs_group_sync_entry(virtual_server_t *vs, virtual_server_group_entry_t *vsge)
 				ipvs_talk(IP_VS_SO_SET_ADDDEST, &srule, &drule, NULL, false);
 			}
 			else {
-				if (vsge->addr.ss_family == AF_INET6)
-					inet_sockaddrip6(&vsge->addr, &srule.nf_addr.in6);
-				else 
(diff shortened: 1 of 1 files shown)


keepalived/check/ipvswrapper.c
```diff
@@ -639,8 +639,14 @@ ipvs_group_sync_entry(virtual_server_t *vs, virtual_server_group_entry_t *vsge)
 				/* Talk to the IPVS channel */
 				ipvs_talk(IP_VS_SO_SET_ADDDEST, &srule, &drule, NULL, false);
 			}
-			else
+			else {
+				if (vsge->addr.ss_family == AF_INET6)
+					inet_sockaddrip6(&vsge->addr, &srule.nf_addr.in6);
+				else 
+					srule.nf_addr.ip = inet_sockaddrip4(&vsge->addr);
+
```
pqarmitage: The above additional 4 lines seem unnecessary since ipvs_group_range_cmd() sets the address field in the srule.

keepalived/check/ipvswrapper.c
```diff
@@ -682,8 +688,14 @@ ipvs_group_remove_entry(virtual_server_t *vs, virtual_server_group_entry_t *vsge
 	if (!is_vsge_alive(vsge, vs)) {
 		if (vsge->range)
 			ipvs_group_range_cmd(IP_VS_SO_SET_DEL, &srule, NULL, vsge);
-		else
+		else {
+			if (vsge->addr.ss_family == AF_INET6)
+				inet_sockaddrip6(&vsge->addr, &srule.nf_addr.in6);
+			else 
+				srule.nf_addr.ip = inet_sockaddrip4(&vsge->addr);
+
```
pqarmitage: The srule will have been set by the call(s) to ipvs_talk()/ipvs_group_range_cmd() with IP_VS_SO_SET_DELDEST in the LIST_FOREACH block above. The only way this would not happen is if there were no real servers, but that would be a configuration error.

HW-lj:
The configuration file is as follows:
```
virtual_server_group tcp1 {
    22.22.2.134 80
    22.22.2.135 80
 }

virtual_server group tcp1 {
    delay_loop 6
    lb_algo rr
    lb_kind NAT
    protocol TCP
    persistence_timeout 120
    #alpha

    real_server 33.33.81.188 80 {
        weight 1
        TCP_CHECK {
            connect_port 80
            connect_timeout 10
            retry 3
            delay_before_retry 3
        }
    }
}
```
step1: start the keepalived.
step2: remove  22.22.2.135 80 from virtual_server_group tcp1 and reload it.
```
log: 
Apr 28 09:28:47 localhost.localdomain Keepalived_vrrp[25132]: Reloading
Apr 28 09:28:47 localhost.localdomain Keepalived_vrrp[25132]: Opening file '/etc/keepalived/keepalived.conf'.
Apr 28 09:28:47 localhost.localdomain systemd[1]: Reloaded LVS and VRRP High Availability Monitor.
Apr 28 09:28:47 localhost.localdomain Keepalived_healthcheckers[25131]: Reloading
Apr 28 09:28:47 localhost.localdomain Keepalived_healthcheckers[25131]: Got SIGHUP, reloading checker configuration
Apr 28 09:28:47 localhost.localdomain Keepalived_healthcheckers[25131]: Opening file '/etc/keepalived/keepalived.conf'.
Apr 28 09:28:47 localhost.localdomain Keepalived_healthcheckers[25131]: Initializing ipvs
Apr 28 09:28:47 localhost.localdomain Keepalived_healthcheckers[25131]: VS [[22.22.2.135]:tcp:80:0:1342177282] in group tcp1 no longer exists
Apr 28 09:28:47 localhost.localdomain Keepalived_healthcheckers[25131]: IPVS (cmd 1156, errno 2): No such file or directory
Apr 28 09:28:47 localhost.localdomain Keepalived_healthcheckers[25131]: Activating healthchecker for service [33.33.81.188]:tcp:80 for VS [tcp1]:0
```
step3: ipvsadm -ln, then you will see the removed one is still there.

keepalived/check/ipvswrapper.c
HW-lj: Yes, when we go into the ipvs_group_range_cmd, there is no problem. While  if we are single one ,that is to say, vsge->range is 0, then we will miss the set of srule.nf_addr.ip. But in fact, I think maybe we just delete the judgement of  vsge->range, whether it is zero or not , it will work well : )

keepalived/check/ipvswrapper.c
HW-lj: Indeed if the real servers are all died,  this would not happen.

HW-lj: But in fact, I think maybe we just delete the judgement of  vsge->range, whether it is zero or not , it will work well : )

pqarmitage:
Unfortunately the patches above don't quite work with the latest HEAD of the master branch. Commit 6643f59 incorporates your changes with some slight tweaks for it to fully work.

Many thanks for your work on this.

PR #1250, opened 2019-05-02
add udp health check

HW-lj:
> There are a number of issues with this patch that mean it cannot be merged in its current form.
> 
> 1. As has already been said, read and write operation must be handled via the I/O MUX, otherwise keepalived could stall while waiting for I/O operations to complete. In particular you will need to use thread_add_write() if when you attempt to send on a socket you get an EWOULDBLOCK error.
> 2. Don't duplicate existing code that can be re-used: check_dns.c has code for setting up UDP sockets (dns_connect_thread() etc). That code should be used, probably renamed udp_connect_thread(), and moved into layer4.c.
> 3. Any sockets created must have O_NONBLOCK and O_CLOEXEC set
> 4. ICMP is not a protocol running on top of UDP. ICMP is a protocol directly above IP/IPv6.
> 5. To be generally useful, the UDP check shouldn't expect to receive a UDP packet in response, but should simply ensure that there isn't an ICMP port unreachable response. There could be an option to say expect to receive a number of bytes in reply.
> 6. Don't use getprotobyname, use IPPROTO_ICMP etc.
> 7. Don't use thread_add_timer with a timeout of 0, use thread_add_event instead.
> 8. udp_check_t ping_check should be a bool, and use true/false with it.
> 9. In ping_it() the echo.sequence number will always be 1. Also, echo.id will frequently have the same value.
> 10. Why does SO_RCVBUF need to be set so large?
> 11. Error and other messages need to be corrected (e.g. UDP rather than TCP).
> 
> The above are examples of how you patch doesn't match the coding standards and practices of keepalived. I you can update your patch in line with the comments above, we will look at it again with a view to merging it.
> 
> If you have other independent patches that you would like merged, it is probably best to create them on different branches, so that they can be considered separately from each other.

1.As has already been said, read and write operation must be handled via the I/O MUX, otherwise keepalived could stall while waiting for I/O operations to complete. In particular you will need to use thread_add_write() if when you attempt to send on a socket you get an EWOULDBLOCK error.

　　In fact, ping is not done by keepalived, but the kernel does it for us. Of course, it's a little tricky. As the function sendto is nonblock, so keepalived will work well. If you want to know more, you can refer to https://stackoverflow.com/questions/8290046/icmp-sockets-linux.

2.Don't duplicate existing code that can be re-used: check_dns.c has code for setting up UDP sockets (dns_connect_thread() etc). That code should be used, probably renamed udp_connect_thread(), and moved into layer4.c.

　　The two pieces of code are exactly the same, but the error log is different. If we abstract these two into a function, then if the user has configured both DNS_CHECK and UDP_CHECK, the log information will become blurred.

3.Any sockets created must have O_NONBLOCK and O_CLOEXEC set

　　OK, I have fixed it.√

4.ICMP is not a protocol running on top of UDP. ICMP is a protocol directly above IP/IPv6.

　　In fact, if we use the traditional usage of ICMP, then we will meet such a problem: when we receive a ICMP echo reply, we must judge whether the id matches our id and then the efficiency is so low.
	Here is a little trick to create SOCK_DGRAM. If you want to know more, you can refer to https://stackoverflow.com/questions/8290046/icmp-sockets-linux.

５.To be generally useful, the UDP check shouldn't expect to receive a UDP packet in response, but should simply ensure that there isn't an ICMP port unreachable response. There could be an option to say expect to receive a number of bytes in reply.

	In fact, if (ret = recv(fd, recv_buf, sizeof(recv_buf)-1, 0) < 0) can exactly ensure that we got an ICMP port unreachable response, and only in this case will we consider the check failed. So I think we have the same idea at this point.

６.Don't use getprotobyname, use IPPROTO_ICMP etc.

	OK, I have fixed it.√

７.Don't use thread_add_timer with a timeout of 0, use thread_add_event instead.

	OK, I have fixed it.√

８.udp_check_t ping_check should be a bool, and use true/false with it.

	OK, I have fixed it.√

9.In ping_it() the echo.sequence number will always be 1. Also, echo.id will frequently have the same value.

	Indeed, the seq set here is no need.

10.Why does SO_RCVBUF need to be set so large? 

　　Considering that we can support more accurate UDP health check later, we can only accept the specified string and we think that the health check is successful.

11.Error and other messages need to be corrected (e.g. UDP rather than TCP).

　　Sorry for my carelessness! I have fixed it.

### DIFF
--- a/keepalived/core/global_data.c
+++ b/keepalived/core/global_data.c
@@ -49,13 +49,12 @@ data_t *old_global_data = NULL;
 
 /* Default settings */
 static void
-set_default_router_id(data_t *data, char *new_id)
+set_default_router_id(data_t *data, const char *new_id)
 {
 	if (!new_id || !new_id[0])
 		return;
 
-	data->router_id = MALLOC(strlen(new_id)+1);
-	strcpy(data->router_id, new_id);
+	data->router_id = STRDUP(new_id);
 }
 
 static void
@@ -227,7 +226,8 @@ init_global_data(data_t * data, data_t *prev_global_data, bool copy_network_name
 {
 	/* If this is a reload and we are running in a network namespace,
 	 * we may not be able to get local_name, so preserve it */
-	char unknown_name[] = "[unknown]";
+	const char unknown_name[] = "[unknown]";
+	char *str;
 
 	/* If we are running in a network namespace, we may not be
 	 * able to get our local name now, so re-use original */
@@ -252,8 +252,8 @@ init_global_data(data_t * data, data_t *prev_global_data, bool copy_network_name
 		/* If for some reason get_local_name() fails, we need to have
 		 * some string in local_name, otherwise keepalived can segfault */
 		if (!data->local_name) {
-			data->local_name = MALLOC(sizeof(unknown_name));
-			strcpy(data->local_name, unknown_name);
+			data->local_name = str = MALLOC(sizeof(unknown_name));
+			strcpy(str, unknown_name);
 		}
 	}
 
@@ -345,7 +345,7 @@ free_global_data(data_t * data)
 	FREE_PTR(data->router_id);
 	FREE_PTR(data->email_from);
 	FREE_PTR(data->smtp_helo_name);
-	FREE_PTR(data->local_name);
+	FREE_CONST_PTR(data->local_name);
 #ifdef _WITH_SNMP_
 	FREE_PTR(data->snmp_socket);
 #endif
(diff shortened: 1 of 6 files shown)


HW-lj:
Since I re-forked it, I closed the previous request and resubmitted it

HW-lj:
Time flies, could you share with me the confusion that you still have and the suggestions for further improvement?

pqarmitage:
@HW-lj I've only just had time to start looking at this, and am working on it at the moment.

I'll provide an update when I have reached some conclusions.

HW-lj:
For more information about ping socket, you can refer to the following links:
https://lkml.org/lkml/2011/5/10/389
https://lwn.net/Articles/550551/

HW-lj:
Emmm,  I just found out that my udp health check patch is missing. And now I will redo a patch.

pqarmitage:
@HW-lj I was working on reviewing the patch when you closed it several days ago. Now that you have re-submitted it I will continue reviewing it.

PR #1251, opened 2019-05-03
refactor switch

teknoraver:
Rewrite get_nl_msg_type() in a way which is less error prone,
shorter, and more compact.

### DIFF
--- a/keepalived/core/keepalived_netlink.c
+++ b/keepalived/core/keepalived_netlink.c
@@ -120,44 +120,31 @@ report_and_clear_netlink_timers(const char * str)
 }
 #endif
 
-static char *
+#define strval(x) [x] = #x
+
+static const char *
 get_nl_msg_type(unsigned type)
 {
-	switch (type) {
-	case RTM_NEWLINK:
-		return "RTM_NEWLINK";
-		break;
-	case RTM_DELLINK:
-		return "RTM_DELLINK";
-		break;
-	case RTM_NEWADDR:
-		return "RTM_NEWADDR";
-		break;
-	case RTM_DELADDR:
-		return "RTM_DELADDR";
-		break;
-	case RTM_NEWROUTE:
-		return "RTM_NEWROUTE";
-		break;
-	case RTM_DELROUTE:
-		return "RTM_DELROUTE";
-		break;
-	case RTM_NEWRULE:
-		return "RTM_NEWRULE";
-		break;
-	case RTM_DELRULE:
-		return "RTM_DELRULE";
-		break;
-	case RTM_GETLINK:
-		return "RTM_GETLINK";
-		break;
-	case RTM_GETADDR:
-		return "RTM_GETADDR";
-		break;
-	}
+	static const char *vals[] = {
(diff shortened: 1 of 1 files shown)


pqarmitage:
@teknoraver Many thanks for this patch. It certainly makes the source code shorter, but I'm not convinced by the less error prone claim.

If we were to adopt this as a general approach, for example see ipvs_cmd_str() in ipvswrapper.c, there is a problem with the size of the array generated when the values of the constants are large. For example using the approach with ipvs_cmd_str() generates an array of 9344 bytes, since the values start at 1153.

A way around this would be to slightly alter the code as follows:
```
#define min_val  RTM_NEWLINK
#define strval(x)  [(x)-min_val] = #x
```
and the test `if (type < sizeof(val) / sizeof(*vals) && vals[type])` becomes
```
if (type >= min_val && type - min_val < sizeof(vals) / sizeof(*vals) && vals[type - min_val]
```
This increases the code size by 2 bytes, but reduces the array size by 128 bytes compared to the original patch version.

A potential criticism of the `min_val` approach could be what happens if the values change and a different defined name becomes the actual minimum value. The answer to that is two-fold:
1. These are exposed UAPI kernel defines and so will not change
2. If any of the indices of the array are negative in the initialisation, the compiler will produce an error message and the compilation will fail.

As a summary of the code and data sizes (in bytes) using gcc 7.3.1 with -O2:
```
Code version       Code size    Data size

Switch version          23         144
Original patch          34         272
Patch with min_val      34         144
```

Equivalent figures for the function ipvs_cmd_str() are:
```
Code version       Code size    Data size

Current version        173            0
Switch version          26          120
Original patch          35         9344
Patch with min_val      37          128
```
Certainly the current code in ipvs_cmd_str() needs to be changed.

In respect of compactness (code and data size) it seems that gcc with -O2 can do at least as well as, and sometimes better than, the approach used in the patch when using a switch statement.

If the code is changed as follows:
```
#define strval(x) case x: return #x; break
const char *
get_nl_msg_type(unsigned type)
{
        switch (type) {
                strval(RTM_NEWLINK);
                strval(RTM_DELLINK);
                strval(RTM_NEWADDR);
                strval(RTM_DELADDR);
                strval(RTM_NEWROUTE);
                strval(RTM_DELROUTE);
                strval(RTM_NEWRULE);
                strval(RTM_DELRULE);
                strval(RTM_GETLINK);
                strval(RTM_GETADDR);
        };

        return "";
}
#undef strval
```
we get both the benefits of the better appearance of the source code, and also the small and faster code and smaller data size that gcc produces with a switch.

@teknoraver Would you be happy with this approach?

teknoraver:
> @teknoraver Would you be happy with this approach?

Yes, go for it!

teknoraver:
@pqarmitage just a small nitpick: discard the break following a return:

`#define strval(x) case x: return #x`

pqarmitage:
Commit 40b3746 updates the code in line with the discussion above. I have removed the break statement from the define as you suggested.

Many thanks for your suggested improvement.

PR #1260, opened 2019-05-10
genhash: ask the server to disable keep-alive

MaxKellermann:
genhash speaks only HTTP/1.0 which strictly speaking doesn't support
keep-alive; but some servers which do not support version 1.0 may
interpret the absence of a "Connection" header as "enable keep-alive",
as required by version 1.1 of the protocol.  This breaks genhash,
which waits until the server closes the connection.

To fix this, this patch adds "Connection:close" which explicitly
disables keep-alive.

### DIFF
--- a/genhash/include/http.h
+++ b/genhash/include/http.h
@@ -41,10 +41,12 @@
 /* GET processing command */
 #define REQUEST_TEMPLATE "GET %s HTTP/1.0\r\n" \
 			 "User-Agent: KeepAlive GenHash Client\r\n" \
+			 "Connection: close\r\n" \
 			 "Host: %s%s\r\n\r\n"
 
 #define REQUEST_TEMPLATE_IPV6 "GET %s HTTP/1.0\r\n" \
 			 "User-Agent: KeepAlive GenHash Client\r\n" \
+			 "Connection: close\r\n" \
 			 "Host: [%s]%s\r\n\r\n"
 
 /* Output delimiters */


pqarmitage:
@MaxKellermann Many thanks for this patch. Please see my following comments:

1. Presumably this change should also be done in include/check_http.h
2. Since HTTP/1.0 does not support Connection:, would it be better to have an HTTP/1.1 version  of the templates which include `Connection: keep-alive`, and have a configuration option e.g. `http-version 1.1` in the checker configuration block (or command line option for genhash)?

MaxKellermann:
With HTTP/1.1, you never need to send `Connection: keep-alive`, because that's the implied default.
Of course, in HTTP/1.0, the `Connection` header is not specified, but it was already widely used, and so common practice was later taken into the 1.1 standard. I think a full 1.1 migration is out of this PR's scope, and rather leave this PR a trivial change to fix one specific real-world problem.
I'll look into check_http.h, though.

pqarmitage:
I should have said `Connection: close` rather than `Connection: keep-alive`.

I'm not sure that we can just add `Connection: close` to the HTTP/1.0 header since it might break for someone if their HTTP server rejected it as invalid; after all this patch is only to work around servers that don't conform correctly.

I think if we want to add `Connection: close` to an HTTP/1.0 header, then it should be a configuration/genhash command line option. I'm happy to add that part if you agree with it.

MaxKellermann:
I don't think any web server will do the wrong thing in the presence of this unspecified header, because it was so widely used even before 1.1 existed. But of course I agree with making it optional.

MaxKellermann:
Btw. about
> after all this patch is only to work around servers that don't conform correctly

Actually, this is about working around servers that don't support HTTP/1.0 which has been obsoleted 22 years ago. Not supporting HTTP/1.0 is a reasonable thing to do.

What really does not conform to any HTTP spec is genhash, which won't read the response until the server closes the connection. The state of the connection shouldn't have anything to do with that.

pqarmitage:
Commit 2ff56f5 adds command line option `-P` to genhash. `-P 1.0C` which will make genhash send an HTTP/1.0 header, but with a `Connection: close` line. `-P 1.1` will send an HTTP/1.1 header, also with a `Connection: close` line. The default remains as it was before, i.e. an HTTP/1.0 header with no `Connection:` line.

I have also added commit 64ef2fa which adds the `http_protocol` keyword to HTTP_GET and SSL_GET checkers. Again, options `1.0C` and `1.1` can be specified, which mirror what genhash now does.

PR #1274, opened 2019-05-18
support http status_code group

HW-lj:
The origin status_code only support one specific code, now we can
support http status_code of the same class. That's to say, we can
use 1xx to represent 100-199, 2xx means 200-299 ans so on.

eg: The configure as follows:

	url {
		path /index.html
		status_code 2xx 3xx
	}

which means we consider all status_code range in [200,399] is ok.
Of course the following configure is either 200 or [300,399] is ok.

	url {
		path /index.html
		status_code 200 3xx
	}

### DIFF
--- a/keepalived/check/check_http.c
+++ b/keepalived/check/check_http.c
@@ -56,6 +56,56 @@
 #define	REGISTER_CHECKER_NEW	1
 #define	REGISTER_CHECKER_RETRY	2
 
+/* Each unsigned long is 64 bits, equals to 1 << 6 */
+#define	STATUS_CODE_ARRAY_BITS	6
+#define	STATUS_CODE_ARRAY_ENTRY_NUM	(1 << STATUS_CODE_ARRAY_BITS)
+#define	STATUS_CODE_ARRAY_MASK		(STATUS_CODE_ARRAY_ENTRY_NUM - 1)
+
+static unsigned long int set_code_bit(url_t *url, int code) {
+	return url->status_code[code >> STATUS_CODE_ARRAY_BITS] |= (1UL << (code & STATUS_CODE_ARRAY_MASK));
+}
+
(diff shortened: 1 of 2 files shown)


pqarmitage:
@HW-lj I have been somewhat concerned about these patches, and the ones submitted as pull request #1273.

The problems with the patches in pull request #1273 were:
1. It didn't compile (2 separate errors)
2. A compiler warning due to isdigit not declared
3. An unused variable compiler warning in status_code_handler()
4. A compiler warning of a shadowed declaration of i in dump_url()
5. If no status code was configured, it doesn't dump the default status codes
6. If a status code was configured, a dump would write output all status codes from 100 to 599, regardless of whether they were configured
7. longs are not necessarily 64 bits, so this wouldn't work on a system with 32 bit longs
8. reload didn't work

Pull request #1274 resolved issues 1, 3 and 8, but the remaining five problems still exist.

The problem with having all these issues, particularly submitting a patch that cannot compile, is that it suggests that there cannot have been any testing, or at least sufficient testing, of the code. Much as we appreciate patches being submitted, both to resolve problems and to add functionality, it makes our lives much harder work if we have to rework patches to ensure that they integrate properly with keepalived.

On this occasion I have prepared some additional patches that I will push after this patch has been merged, but it really would be appreciated if you could test all aspects of your patches before submitting them. While there may be the odd error in a patch, or an improvement that can be made, that is expected to be the exception rather than the rule.

Finally, to help other users make use of enhancements such as this, the documentation needs to be updated. As a very minimum, the relevant man pages should be updated.

HW-lj:
Sorry for my carelessness and thank you for your patience. I will do better in next time ლ(╹◡╹ლ). BTW, 
it’s really annoying to forget to update my local source code. Otherwise, many problems can be killed in the cradle.

PR #1277, opened 2019-05-21
Fix systemd notice caused by PIDFile referencing /var/run

SjonHortensius:
exact message: "PIDFile= references a path below legacy directory /var/run/"

please review this carefully, I removed the variable from `keepalived.service.in` as all systemd machines will output this error when PIDFile doesn't start with /run - so it's pretty much a fixed requirement.

Notice introduced in https://github.com/systemd/systemd/issues/10657

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2255,6 +2255,8 @@ GNU_STD_PATHS=No
 if test "${enable_gnu_std_paths}" = "yes"; then
   AC_DEFINE([GNU_STD_PATHS], [ 1 ], [set to enforce GNU standard paths, for .pid files etc])
   PID_DIR=$localstatedir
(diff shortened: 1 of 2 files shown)


pqarmitage:
@SjonHortensius Many thanks for reporting this, and your warning to review the patch carefully.

Further updates were needed to keepalived for this, since keepalived itself wasn't using what configure set as PID_DIR. See commit 6977a5c. The change of PID_DIR to use /run rather than /var/run is now done if /run exists and is a directory, rather than simply if INIT_TYPE is systemd.

SjonHortensius:
@pqarmitage Thanks a lot for your quick response and improved fix!

PR #1285, opened 2019-05-24
add udp check

HW-lj:


HW-lj:
If there are still many problems with this, could you prepare the patch first, then put your patch on after merging these patches? You can modify the unreasonable part directly, which may be more efficient. : )

HW-lj:
For the sake of convenience, here come the previous comments. https://github.com/acassen/keepalived/pull/1199#issuecomment-486275249
https://github.com/acassen/keepalived/pull/1250#issue-275268618

pqarmitage:
With many apologies for taking so long to merge this functionality.

Commit 303d2a1 now adds this functionality, but separated out as two checkers, a UDP_CHECK and a PING_CHECK, so that they an be used independently of each other.

PR #1292, opened 2019-05-31
handle pidfile_write correctly

HW-lj:
fprintf will fail under certain circumstances,
such as the disk is full.


pqarmitage:
I'm not convinced about this patch. We have a failure such as the disk is full and we attempt to report it by (probably) writing to the disc. It isn't a big problem if we can't write the pid file; it just makes stopping keepalived more difficult. If the disk is full, the system will have far greater problems.

HW-lj:
OK.

PR #1297, opened 2019-06-04
remove ON_SUCCESS case in http_handle_response

HW-lj:
Since we previously merged the default configuration with the specified configuration, it seems we will never come to the ON_SUCCESS case.


pqarmitage:
Many thanks for reporting this issue. This has now been resolved as part of the patch set resolving issues identified by coverity.

Issue #1299, opened 2019-06-04
Wrong escaping of characters in the JSON output


zdc:
**Describe the bug**
JSON writer escapes the `'` charter, which is not required by [ECMA-404](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf), nor [RFC4627](https://www.ietf.org/rfc/rfc4627.txt) (see "String" section). This leads to incorrect formatting of JSON output.

**To Reproduce**
Configure VRRP instance with any `notify_` script. Run keepalived and dump information into the JSON format. Check the content of the JSON file (`/tmp/keepalived.json`). Example:
```
[{"data":{"iname":"G1","dont_track_primary":0,"skip_check_adv_addr":0,"strict_mode":0,"vmac_ifname":"","ifp_ifname":"eth1","master_priority":0,"last_transition":1559296646.621354,"garp_delay":5,"garp_refresh":0,"garp_rep":5,"garp_refresh_rep":1,"garp_lower_prio_delay":5,"garp_lower_prio_rep":5,"lower_prio_no_advert":0,"higher_prio_send_advert":0,"vrid":1,"base_priority":100,"effective_priority":100,"vipset":false,"promote_secondaries":false,"adver_int":1,"master_adver_int":1,"accept":1,"nopreempt":false,"preempt_delay":0,"state":1,"wantstate":1,"version":2,"smtp_alert":false,"script_backup":"\'/bin/echo\' \'123\'","vips":["192.168.2.2/24 dev eth1 scope global"],"auth_type":0},"stats":{"advert_rcvd":0,"advert_sent":0,"become_master":0,"release_master":0,"packet_len_err":0,"advert_interval_err":0,"ip_ttl_err":0,"invalid_type_rcvd":0,"addr_list_err":0,"invalid_authtype":0,"authtype_mismatch":0,"auth_failure":0,"pri_zero_rcvd":0,"pri_zero_sent":0}}]
```
Such file does not have valid JSON syntax.

**Expected behavior**
Character `'` must be not escaped. Also, there is no direct denotement in RFC that we must escape `/` character, and JSON with unescaped `/` will be valid. But, due to security reasons and recommendation from RFC, this character should be escaped.

**Keepalived version**
`Keepalived v2.0.16 (06/04,2019), git commit v2.0.16-123-gc5aecf3`

**Distro (please complete the following information):**
 - Name: Debian
 - Version: 8.11
 - Architecture: x86_64

**Configuration file:**
```
[...]

PR #1300, opened 2019-06-04
Fix for JSON characters escaping (issue #1299)

zdc:
Changes, according to https://github.com/acassen/keepalived/issues/1299.


pqarmitage:
@zdc Many thanks for identifying the problem and providing a rapid fix.

PR #1305, opened 2019-06-12
Fix segfault when we do not config vsg

HW-lj:
When we do not config vsg, during reloading,
the old and new vsg is NULL, so we can not
use old(new)->vsg->addr_range or
old(new)->vsg->vfwmark directly.

### DIFF
--- a/keepalived/check/ipwrapper.c
+++ b/keepalived/check/ipwrapper.c
@@ -767,6 +767,9 @@ update_alive_counts(virtual_server_t *old, virtual_server_t *new)
 	list *old_l, *new_l;
 	element e;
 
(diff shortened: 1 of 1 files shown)


craiggowing:
We came across the same problem and came to the same conclusion regarding the cause and fix. We rebuilt using this patch and can confirm it has solved the segfault issue for us.

HW-lj:
> We came across the same problem and came to the same conclusion regarding the cause and fix. We rebuilt using this patch and can confirm it has solved the segfault issue for us.

lol, my pleasure : )

acassen:
thanks for investigating, just merge patchset

Issue #1301, opened 2019-06-06
http/ssl health check with two or more url


HW-lj:
Hi, all:
    When we config two or more urls in HTTP_GET,  when all the url check success, should we only check the first url or still check all urls?

pqarmitage:
All URLs should be checked.

pqarmitage:
Commits 205660d - 16cd32f resolve this issue and also implement some performance improvements.

PR #1306, opened 2019-06-12
Fix HTTP_GET with two or more urls

HW-lj:
When we config two or more urls,. With the following step
to reproduce:
1. let the two urls check success
2. modify the second url to fail and reload
3. the real server will never be offline

the reason is we will always check the first url after we have
been online. As the reply of issue:
https://github.com/acassen/keepalived/issues/1301#issuecomment-501049906
All URLs should be checked. So when we successfully check one
URL, we should keep the retry_it and check the next URL until we failed,
adding the retry_it and start another round check. Of course,
when we check all URLs successfully, we should reset retry_it
and url_it.

### DIFF
--- a/keepalived/check/check_http.c
+++ b/keepalived/check/check_http.c
@@ -917,10 +917,19 @@ epilog(thread_ref_t thread, int method, unsigned t, unsigned c)
 	bool checker_was_up;
 	bool rs_was_alive;
 
-	http_get_check->url_it += t ? t : -http_get_check->url_it;
-	checker->retry_it += c ? c : -checker->retry_it;
+	if (method == REGISTER_CHECKER_NEW) {
+		http_get_check->url_it += t ? t : -http_get_check->url_it;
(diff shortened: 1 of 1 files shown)


pqarmitage:
Unfortunately this patch isn't quite right. The retry_it counter is supposed to be per url, not the total while all URLs are checked. For example if there 

I have reworked this patch, and also found some performance optimisations which I have applied. Commits 205660d - 16cd32f implement the above fix and the performance improvements.

HW-lj:
Thank you for your working. Although, in fact, I think this is just a mechanism. My patch is to check from the first one when a url fails, and your current mechanism is to check the failed url and continually retry the url. Both mechanisms are acceptable, of course, this is up to you.

At last, thank you for your efforts of a number of enhancements again.

pqarmitage:
I think the key aspect is that both patches ensure that all the urls are checked. Retrying the failed url is simply an optimisation that enables quicker recovery once the failed url is working again.

PR #1307, opened 2019-06-12
check_dns: fix blank DNS name problem

ZhiqiangLiu26:
In dns_name_handler func, set_value(strvec) will return NULL when
DNS name is blank in configure file. It will cause
reference-null-pointer problem in dns_make_query func.

Here, we will report config error and set DNS name to DNS_DEFAULT_NAME
when DNS name is NULL.

Signed-off-by: Zhiqiang Liu <liuzhiqiang26@huawei.com>
Reported-by: Ailing Duan <duanailing1@huawei.com>

### DIFF
--- a/keepalived/check/check_dns.c
+++ b/keepalived/check/check_dns.c
@@ -509,8 +509,14 @@ dns_type_handler(const vector_t *strvec)
 static void
 dns_name_handler(const vector_t *strvec)
 {
+	char *dns_name = NULL;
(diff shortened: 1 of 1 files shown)


pqarmitage:
This seems to duplicate pull request #1308.

pqarmitage:
Please see updated comments in pull request #1308.

ZhiqiangLiu26:
> Please see updated comments in pull request #1308.

ok， I think that is a better solution. I will close the pull request.

PR #1308, opened 2019-06-13
Fix potential segfault when keywords without value

HW-lj:
As the pull request
https://github.com/acassen/keepalived/pull/1307#issue-287524770
says: when we meet the situation of like those who call
set_value, but the value is missing. Although we return the NULL
pointer, but it will still meet segfault. We can easily to reproduce
just like the config without a path:

 HTTP_GET {
 	url {
		path
		status_code 200
	}
}

So I consider it necessary to ensure all keywords using set_value
should have a value.

### DIFF
--- a/lib/parser.c
+++ b/lib/parser.c
@@ -2003,9 +2003,6 @@ read_value_block(const vector_t *strvec)
 const char * __attribute__((malloc))
 set_value(const vector_t *strvec)
 {
-	if (vector_size(strvec) < 2)
-		return NULL;
-
 	return STRDUP(vector_slot(strvec, 1));
 }
 


ZhiqiangLiu26:
Nice work!!

pqarmitage:
I'm not sure this is the best way to do this. It would be simpler to call strvec_slot even if vector_size(strvec) < 2. I will have a look at this further.

pqarmitage:
Changing set_value() to
```
const char * __attribute__((malloc))
set_value(const vector_t *strvec)
{
        return STRDUP(strvec_slot(strvec, 1));
}
```
is a simpler change and it has the same effect and avoids some code duplication.

Before merging the change, I want to make sure that it is not going to cause any problems at the other points where set_value() is called.

HW-lj:
> Changing set_value() to
> 
> ```
> const char * __attribute__((malloc))
> set_value(const vector_t *strvec)
> {
>         return STRDUP(strvec_slot(strvec, 1));
> }
> ```
> 
> is a simpler change and it has the same effect and avoids some code duplication.
> 
> Before merging the change, I want to make sure that it is not going to cause any problems at the other points where set_value() is called.

Yes, this is a better solution. I will update my patch latter.

pqarmitage:
Commit ed7ca7d incorporates the above fix; however, several of the uses of set_value() did not cause a problem if the keyword parameter was omitted, so specific checks are now made in those circumstances to log a configuration error, but not terminate keepalived.

Issue #1309, opened 2019-06-13
The build is failing after commit 7059aba in some environments


zdc:
**Describe the bug**
`keepalived` is failed to build when `_HAVE_VRRP_VMAC_` is defined, but `_HAVE_VRRP_IPVLAN_` is not. The issue was added with https://github.com/acassen/keepalived/commit/7059abacba2de666b60fec8ba16f8af0c939a363 several days ago.

**To Reproduce**
Fetch `g5eefc57` version from the repository. Run:
```
./build_setup
dh_make --createorig  -s -y -p keepalived_2.0.16-5eefc57
dpkg-buildpackage -us -uc -b
```
The build will fail with the followed error:
```
[...]

PR #1310, opened 2019-06-13
Fix building (issue https://github.com/acassen/keepalived/issues/1309)

zdc:
I was not digging very deep for this fix, so I cannot be pretty sure if this will not break some internal logic. @pqarmitage it is better to take a look for this before merging.


pqarmitage:
@zdc - Many thanks for this patch. The patch itself looks fine, but the original code is rubbish and can never be true.

Commit 1f32162 incorporates your patch, and also makes the original code work.

Issue #1311, opened 2019-06-13
False-positive notifiers during reloading configuration in VRRP BACKUP state


zdc:
**Describe the bug**
During reloading configuration (SIGHUP), if VRRP is in BACKUP state and this state is not changing after reloading, `vrrp_scheduler` generate a false-positive `send_instance_notifies`, which leads to unnecessary invoking `notify_` scripts.

**To Reproduce**
1. Configure the VRRP instance with `notify` script (or `notify_backup`). Example script content:
```
#!/bin/bash

echo -e "$@" >> /tmp/vrrp_notify_log.txt

exit 0
```
2. Run `keepalived` and put this instance to BACKUP state.
3. Reload `keepalived` configuration: `kill -HUP $(cat /var/run/keepalived.pid)`
4. Check if the configured script was executed by reviewing the content of `/tmp/vrrp_notify_log.txt` file:
```
INSTANCE G1 BACKUP 80
```

**Expected behavior**
`notify` script must be not executed, because VRRP state was not changed (BACKUP -> BACKUP).

**Keepalived version**
```
Keepalived v2.0.16 (06/12,2019), git commit v2.0.16-135-g5eefc57+

Copyright(C) 2001-2019 Alexandre Cassen, <acassen@gmail.com>

Built with kernel headers for Linux 3.16.64
Running on Linux 4.19.40-amd64-vyos #1 SMP Sat May 11 18:50:24 CEST 2019

configure options: --build=x86_64-linux-gnu --prefix=/usr --includedir=${prefix}/include --mandir=${prefix}/share/man --infodir=${prefix}/share/info --sysconfdir=/etc --localstatedir=/var --disable-silent-rules --libdir=${prefix}/lib/x86_64-linux-gnu --libexecdir=${prefix}/lib/x86_64-linux-gnu --disable-maintainer-mode --disable-dependency-tracking --enable-snmp --enable-sha1 --enable-dbus-create-instance --enable-dbus --enable-json build_alias=x86_64-linux-gnu CFLAGS=-g -O2 -fstack-protector-strong -Wformat -Werror=format-security LDFLAGS=-Wl,-z,relro CPPFLAGS=-D_FORTIFY_SOURCE=2

[...]

PR #1312, opened 2019-06-13
Fix false-positive send_instance_notifies calls (issue #1311)

zdc:
Changes, according to #1311.


pqarmitage:
Due to intervening commits I merged, this commit now has merge conflicts.

I have updated the patch and merged commit 15cfac1.

Many thanks for providing the original patch.

PR #1346, opened 2019-07-02
Remove unnecessary judgement

HW-lj:
It seems that read_double_func has already
hanlded this.


pqarmitage:
read_double_func() does not necessarily set a value less than fmax_time, so the check is necessary. It will report an error if it is outside the bounds of [fmin_time, fmax_time], but the value set in timer may be outside those bounds.

PR #1347, opened 2019-07-03
Fix potential overflow in read_timer()

HW-lj:
As the reply
https://github.com/acassen/keepalived/pull/1346#issuecomment-507678358
says:
read_double_func() does not necessarily set a value,
so read_timer() has the risk of overflow.


pqarmitage:
@HW-lj Many thanks for this patch. However, it appears that you are thinking about this in terms of integer arithmetic, which can indeed overflow. Since this is floating point arithmetic, if `timer * TIMER_HZ` exceeds the maximum valid double value, then it will have the value `inf` (infinity), which will be greater than TIMER_MAXIMUM, so the current code is correct in this respect.

What might be more appropriate would be to only do the calculation if the returned value from `read_double_strvec()` is valid, i.e. if `ret` is true. Any function calling `read_timer()` (or indeed `read_double_strvec()` must check the value returned via `res` is valid (i.e. `ret` is true) before using it. All uses of `read_timer()` do indeed check ret before using the returned value.

PR #1349, opened 2019-07-04
sll_protocol should be set to  0x806

wanrui:
Some times , send the gratuitous ARP message should set sll_protocol , let some drivers can evaluate which protocol we use


pqarmitage:
@wanrui Should ssl_protocol also be set in ndisc_send_na() in vrrp_ndisc.c?

wanrui:
@pqarmitage  yes , some drivers use `switch (vlan_get_protocol(skb))` to find which protocol use to do some special action.  if not set sll_protocol or default 0 , that will not work correct.
so should i make another pr?

wanrui:
yes , some drivers use `vlan_get_protocol(skb)`  to find which protocol use to do some special action. if not set sll_protocol or default 0 , that will not work correct.

PR #1353, opened 2019-07-12
Draft: Add logging for reasons why interfaces are going into fault state.

ctheune:
I had to hunt down a very surprising problem where an instance would go into fault state without telling me why.

Turns out having a v4 virtual ip (with v6 virtual excluded ips) will assign a link-local address for v6 but if the interface doesn't have a local v4 address then the instance will be down because it won't use the v6 address.

I have added a bunch of hopefully helpful logging statements that will now tell which interface caused a group to go into sync (otherwise you just see the group going to FAULT and _then_ the interfaces switching to FAULT) and why each interface went to FAULT state. I may have missed some code paths and I guess this could become a function and I'd also like to add more information (marked with XXX) with some log entries.

Right now I'd appreciate if you could let me know whether this has a chance of being merged ...

### DIFF
--- a/keepalived/vrrp/vrrp.c
+++ b/keepalived/vrrp/vrrp.c
@@ -1740,7 +1740,7 @@ vrrp_state_leave_master(vrrp_t * vrrp, bool advF)
 		vrrp->master_adver_int = vrrp->adver_int;
 	}
 	else if (vrrp->wantstate == VRRP_STATE_FAULT) {
-		log_message(LOG_INFO, "(%s) Entering FAULT STATE", vrrp->iname);
+		log_message(LOG_INFO, "(%s) Entering FAULT STATE (wanted)", vrrp->iname);
 		vrrp_send_adv(vrrp, VRRP_PRIO_STOP);
 	}
 	else {
--- a/keepalived/vrrp/vrrp_if.c
+++ b/keepalived/vrrp/vrrp_if.c
@@ -481,7 +481,7 @@ dump_if(FILE *fp, const void *data)
 	if (IS_VLAN(ifp)) {
 		const char *if_type =
 #ifdef _HAVE_VRRP_IPVLAN
-				      ifp->if_type == IF_TYPE_IPVLAN ? "IPVLAN" : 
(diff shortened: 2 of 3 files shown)


pqarmitage:
@ctheune Many thanks for submitting this patch. Unfortunately the patch looks to be a work in progress, since not only does it fail to compile (see the Travis-CI report above), and therefore cannot have been tested, but also there are various comments starting `/* XXX` that have been added and need resolving, and at least one typo in a log message.

For each of the `/* XXX` comments, it should be straightforward to add the interface/file name/process name to the log message.

If you could update the patch and resubmit, it would be really helpful. Also, an indication of what testing has been performed would be good.

ctheune:
Yeah, absolutely - that's why I added the "draft" in the title: I wanted to check out whether you're interested in this patch at all ... and it seems like you're interested in it. :)

I have a working version on the local machine I used for debugging that works but haven't properly run the cleaned up draft yet. I'll wrap this up and make it compile. :) Any other wishes?

ctheune:
Alright, I took some time to clean this up.

It compiles for me now again (and I did push indeed a version that I never compiled as I had missed to add a few more changes) and I tested in our lab for a couple of vectors (specifically tracked files). I did also see the logging for the very specific case of "no source address specified, no appropriate address found" work as that was what triggered me to debug this in the first place ...  :)

The logging overall is maybe a bit verbose but I didn't understand whether there's a bigger picture that I can fit those messages into, so I just tried to be as similar to the existing messages as possible ... 

Let me know if you like this and whether you need/like any more changes. I'd be happy to squash the commits for the PR if needed.

pqarmitage:
These changes are now merged in commit 1c6d7a3.

The addition of the log_message in update_added_interface() required the addition of { and } around the if block, so I wasn't able to merge your commit directly.

I have also changed it so that the track_script status changes are only logged if the -D (log detail) option is set.

Many thanks for your work on this.

ctheune:
Thanks for picking this up - I've been way to busy over the last months and haven't been able to stay on top if this (for me) side issue ... :)

PR #1370, opened 2019-08-06
Change order of IPVS commands on quorum change

Dragonn:
**Disclaimer**
I know I am using `sorry_server` feature in wrong way, but since I spent some time with debugging I want to share my findings and I believe even discussion could be helpful for the project.

**Issue**
When there is virtual server configuration with `sorry_server` exactly same like one of `real_server`, service is unable to recover properly during transition from sorry server stage back when "twin" `real_server` comes up first.

Keepalived simply failed to add `real_server` because of duplicity in IPVS, but thinks everything is OK.

I am using this setup to mitigate complete out of service when backend health checks are implemented incorrectly.

**My solution**
I can see two minor caveants in my solution:
* When `inhibit_on_failure` is used for real servers it has to be used for sorry server also, or same behavior occurs
* There is very short moment during status transition when no server is present in IPVS structure which may result in dropping some connections in very high load

### DIFF
--- a/keepalived/check/ipwrapper.c
+++ b/keepalived/check/ipwrapper.c
@@ -465,9 +465,7 @@ update_quorum_state(virtual_server_t * vs, bool init)
 				    , weight_sum
 				    , FMT_VS(vs));
 		if (vs->s_svr && ISALIVE(vs->s_svr)) {
-			/* Adding back alive real servers */
-			perform_quorum_state(vs, true);
-
+			/* Removing sorry server since we don't need it anymore */
 			log_message(LOG_INFO, "%s sorry server %s from VS %s"
 					    , (vs->s_svr->inhibit ? "Disabling" : "Removing")
 					    , FMT_RS(vs->s_svr, vs)
@@ -475,6 +473,9 @@ update_quorum_state(virtual_server_t * vs, bool init)
 
 			ipvs_cmd(LVS_CMD_DEL_DEST, vs, vs->s_svr);
 			vs->s_svr->alive = false;
+
+			/* Adding back alive real servers */
+			perform_quorum_state(vs, true);
 		}
 
 		do_vs_notifies(vs, init, threshold, weight_sum, false);
@@ -505,12 +506,12 @@ update_quorum_state(virtual_server_t * vs, bool init)
 					    , FMT_RS(vs->s_svr, vs)
 					    , FMT_VS(vs));
 
+			/* Remove remaining alive real servers */
+			perform_quorum_state(vs, false);
+
 			/* the sorry server is now up in the pool, we flag it alive */
 			ipvs_cmd(LVS_CMD_ADD_DEST, vs, vs->s_svr);
 			vs->s_svr->alive = true;
-
-			/* Remove remaining alive real servers */
-			perform_quorum_state(vs, false);
 		}
 
 		do_vs_notifies(vs, init, threshold, weight_sum, false);


pqarmitage:
@Dragonn So that we can understand exactly what you want to achieve, could you please provide a copy of your keepalived configuration.

From what I understand, you want to use the same real server as both a normal real server and a sorry server. As I think you are aware, the purpose of a sorry server is, for example, to give a response like "Sorry - service is currently unavailable" for a web service when all the real servers are down; clearly it doesn't make sense for a real server to be responding like this when normal service is available. To help us understand what is desired, could you please explain what it is you want to achieve.

The two caveats you describe are important, and we will need to resolve them. For example regarding `inhibit on failure` having to match for the sorry server and the corresponding real server, we would need to check that when parsing the configuration and handle it appropriately. in respect of the second caveat, if `inhibit_on_failure` is not used, any existing connections to the sorry server would be dropped when quorum was achieved and the sorry server was deleted and the (matching) real server reinstated. It might be that rather than the approach you have adopted, an additional keyword is added to real servers to indicate the weight they should be configured with in the event that the quorum for the virtual server is down; this could have other advantages such that several real servers could be configured in the event of quorum down (although why this would be wanted I am not sure).

Dragonn:
I completely understand original purpose of sorry server and I also know that we kinda misussed it. We use loadbalancing for internal API endpoints only, and I don't really care if I get request timeout or sorry answer.

The main reason why we used this type of configuration was when health checks were not properly implemented (LB config and backends are handled by different teams). Unfortunately this kind of scenario is not as rare as we want it to be. When somebody deploys backend version with incorrectly configured healthcheck, all real servers are marked as failed, all traffic is routed to single server (configured as sorry server) and whole service is not completely down.

My current configuration:
```
virtual_server 192.168.0.200 80 80 {
    delay_loop 1
    lb_algo wrr
    lb_kind NAT
    protocol TCP
    sorry_server 192.168.0.100 80

    real_server 192.168.0.100 80 {
        weight 1000
        inhibit_on_failure
        HTTP_GET {
           url {
                path /service-status
                status_code 200
           }
           connect_timeout 1
        }
    }
    real_server 192.168.0.101 80 {
        weight 1000
        inhibit_on_failure
        HTTP_GET {
           url {
                path /service-status
                status_code 200
           }
           connect_timeout 1
        }
    }
}
```

Recently I have discovered strange behavior when healthchecks were OK (I checked backends access logs and network dump), but single real server was missing in IPVS. I figured out that specific order of events triggers this situation.

Real example:
1) When everything is running, it looks like this:
```
  -> RemoteAddress:Port           Forward Weight ActiveConn InActConn
TCP  192.168.0.200:80 wrr
  -> 192.168.0.100:80              Masq    1000   0          0
  -> 192.168.0.101:80              Masq    1000   0          0
```

2) Both backend servers goes down and sorry server pops up:
```
  -> RemoteAddress:Port           Forward Weight ActiveConn InActConn
TCP  192.168.0.200:80 wrr
  -> 192.168.0.100:80              Masq    1000   0          0
```

3) Now when first server goes up first, it will never be really used:
```
  -> RemoteAddress:Port           Forward Weight ActiveConn InActConn
TCP  192.168.0.200:80 wrr
```
and keepalived logs error:
```
Aug 08 14:57:37 server Keepalived_healthcheckers[11342]: HTTP status code success to [192.168.0.100]:tcp:80 url(1).
Aug 08 14:57:37 server Keepalived_healthcheckers[11342]: Remote Web server [192.168.0.100]:tcp:80 succeed on service.
Aug 08 14:57:37 server Keepalived_healthcheckers[11342]: Adding service [192.168.0.100]:tcp:80 to VS [192.168.0.200]:tcp:80
Aug 08 14:57:37 server Keepalived_healthcheckers[11342]: Gained quorum 1+0=1 <= 1000 for VS [192.168.0.200]:tcp:80
Aug 08 14:57:37 server Keepalived_healthcheckers[11342]: Adding alive servers to the pool for VS [192.168.0.200]:tcp:80
Aug 08 14:57:37 server Keepalived_healthcheckers[11342]: IPVS (cmd 1159, errno 17): Destination already exists
Aug 08 14:57:37 server Keepalived_healthcheckers[11342]: Removing sorry server [192.168.0.100]:tcp:80 from VS [192.168.0.200]:tcp:80
```

As I said before, I am completely aware I am using in way it wasn't designed to, but still I consider this a bug.

pqarmitage:
What I think is really the bug here is not checking that the sorry server is different from all the real servers; it was never anticipated that someone would configure it that way, and hence the check was not made. In fact, there is no check for any duplicate real servers for the same virtual server, and that is where the real bug is, and it certainly does not work properly.

Having said the above, refusing to allow duplicates amongst the real and sorry servers for a virtual server is not going to work for you, and so I think we need to think of a solution that works for everyone.

I will continue thinking about this and come back with a proposal.

pqarmitage:
@Dragonn Many apologies for the long delay in dealing with this.

While the patch is fine is far as it goes, it needs some more work to ensure:
1. The inhibit setting matches
2. To be aware that the sorry server duplicates a real server, so that if inhibit is set it doesn't try adding both the real server and the sorry server, and similarly for when keepalived terminates.

I will merge a further patch dealing with the above once I have merged this one.

PR #1426, opened 2019-11-07
Mkosi

johannbg:
This adds initial mkosi support to keepalived and will deliver bootable fedora test vm with changes in current git, the other distro's probably needs some work since I just threw this together in a jiffie and Arch usually has some packaging quirks while Debian is usually far behind other distro's etc.  

Basically what needs to be done is

Arch
pacman -S mkosi
Debian
apt-get install -y mkosi
Fedora
dnf install mkosi
git clone https://github.com/acassen/keepalived.git
cd keepalived && sudo "mkosi" and it will built an Fedora image with current git and run it and you should be able to ssh into it as root with the root password CHANGEME.
If you want to throw away the build simply delete the clone of the repo

This setup is a throw away one time use test vm setup of the git repository as in you cannot update or install anything on the image but you can log into it and make local modifications. If the intent is to deliver a full blown vm, tightly integrated with tests or act as an deliverable for customers to deploy in cloud or on premis or on hardware this needs more work.

If the intent would be to just use mkosi to check "does keepalive build on X distro's"  stuff can be removed ( you dont need bootable vm's for that just the output from mkosi when it generates the image ).

It's just a matter of were to go from here if mkosi built images are considered useful for the project.


johannbg:
I was going to work on this further since I noticed that this pull request configuration and distro specific path breakage in my type unit ( which is supposed to be distro agnostic and follow [1] ) after I had brief window of looking at at arch ) I'm wondering what's your ( upstream  ) take is on [1] as the configure defaults for all distro in mkosi, ( configuration defaults that get for the component are located in mkosi.build file btw ) 

As well as another integration question about default sysctl/iptables/nftables settings. 
What I mean is the current version set in such manners it's "Dont touch sysctl/iptables/nftables etc setting, we will take care of that for you by default and or based on what you have configured in keepalived.conf, so admins should not touch anything. 

Or is the implementation that the administrator needs to do this and that while keepalived takes care of some of it, which we would do in the built image ( such files are located in mkosi.extra btw ).

Like if we take sysctl settings for example. 
Distro's have their own default, kernel has it's own defaults, ( which may or may not be overwritten by each other ) then systemd has it's own ( which might conflict with the other  two ) which may or may not be needed and then keepalived is also turning things on and off  so the best way is to build the image with the sysctl.d/number-keepalived.conf to well probably make the best attempt to ensure the settings get applied right? 

1. https://www.freedesktop.org/software/systemd/man/file-hierarchy.html

pqarmitage:
@johannbg 

1. keepalived should follow the Filesystem Hierarchy Standard unless it not possible to do so. Indeed I recently changed the code to use /run instead of /var/run to comply with the FHS.

2. keepalived looks after all sysctl/iptables/nftables settings that it can, even to the extent of updating net.ipv4.conf.all.rp_filter as you saw at https://github.com/systemd/systemd/issues/6282. There is no sysctl setting that a user needs to make. 

Likewise for nftables; it creates its own tables with the necessary hooks.

iptables, however, is slightly different. For iptables, the chains in which keepalived should add the rules are specified in the configuration, but there needs to be rules in higher up chains (such as INPUT, FORWARD, OUTPUT or chains that they call) that invoke the chains that keepalived is adding its rules to.

3. See above

4. I would rather not have a sysctl.d/number-keepalived.conf file, since that is not how keepalived is normally run. Just let keepalived sort out the necessary settings.

johannbg:
Hmm had forgotten about this one...
Anyways it's old and outdate PR. mkosi even has it's own github action these days so closing...

PR #1447, opened 2019-11-22
configure: fixed fwmark check inverted logic

thresheek:
Seemingly broken in f80b1197dc6b2fefef21c242ac0bd2545b5d7f7a,
--disable-fwmark since enables fwmark instead of disabling it:

$ ./configure --disable-lvs --disable-fwmark
checking for a BSD-compatible install... /usr/bin/install -c
[...]
checking the archiver (ar) interface... ar
configure: error: enable-fwmark requires lvs

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -147,7 +147,7 @@ AC_ARG_WITH(kernel-dir,
   [AS_HELP_STRING([--with-kernel-dir=DIR], [path to linux kernel source directory])],
   [AS_HELP_STRING([kernel_src_path="$withval"],], [[kernel_src_path=""])])
 AC_ARG_ENABLE(fwmark,
-  [AS_HELP_STRING([--disable-fwmark], [compile without SO_MARK support])])
+  [AS_HELP_STRING([--enable-fwmark], [compile with SO_MARK support])])
 AC_ARG_ENABLE(snmp,
   [AS_HELP_STRING([--enable-snmp], [compile with SNMP support])])
 AC_ARG_ENABLE(snmp-vrrp,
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -147,7 +147,7 @@ AC_ARG_WITH(kernel-dir,
   [AS_HELP_STRING([--with-kernel-dir=DIR], [path to linux kernel source directory])],
   [AS_HELP_STRING([kernel_src_path="$withval"],], [[kernel_src_path=""])])
 AC_ARG_ENABLE(fwmark,
-  [AS_HELP_STRING([--disable-fwmark], [compile without SO_MARK support])])
+  [AS_HELP_STRING([--enable-fwmark], [compile with SO_MARK support])])
```
pqarmitage: I don't think this is correct; fwmark should be enabled by default if LVS is enabled.

configure.ac
```diff
@@ -403,7 +403,7 @@ AS_IF([test .$enable_lvs = .no],
     AS_IF([test .$enable_libnl != .], [AC_MSG_ERROR([disable-libnl requires lvs])])
     AS_IF([test .$enable_lvs_syncd != .], [AC_MSG_ERROR([disable-lvs-syncd requires lvs])])
     AS_IF([test .$enable_lvs_64bit_stats != .], [AC_MSG_ERROR([disable-lvs-64bit-stats requires lvs])])
-    AS_IF([test .$enable_fwmark != .], [AC_MSG_ERROR([enable-fwmark requires lvs])])
+    AS_IF([test .$enable_fwmark != .no], [AC_MSG_ERROR([enable-fwmark requires lvs])])
```
pqarmitage: This would require --disable-fwmark to be specified if --disable-lvs is specified, which is superfluous.

thresheek:
Thank you for your review! I'm going to drop our usage of both options simultaneously if it's considered superfluous. This PR can be dropped.

pqarmitage:
@thresheek If you want to disable the IPVS/healthchecker functionality, then still specify `--disable-lvs`. If you are only using the VRRP and/or BFD functionality, and not using IPVS, then it will reduce the size and memory usage of the keepalived executable.

If you want to enable LVS functionality but disable the fwmark option within LVS, then specify `--disable-fwmark`.

I hope that helps.

Issue #1555, opened 2020-04-23
real_server FILE_CHECK may treat state: "success" at initialization


libesz:
**Describe the bug**
The new file_track option for real_server may treat the initial status as "success", regardless of the file content on the first read. In this case, until the first file content change the respective real_server will be enabled in the virtual_server regardless when it is failed in the reality and file content is according to this.

**To Reproduce**
Any steps necessary to reproduce the behaviour:
Configure the following:
```
[...]
track_file TF_local {
 file /tmp/node_status/127.0.0.1
 weight 0
}
[...]
virtual_server 192.168.1.1 30091 {
  delay_loop 5
  inhibit_on_failure
  lvs_sched rr
  lvs_method TUN
  protocol TCP
  real_server 127.0.0.1 30091 {
    lvs_method DR
    FILE_CHECK {
      track_file TF_local
    }
  }
}
[...]
```
If the file content contains 1 (meaning failure), the ipvs will be still programmed to allow traffic towards the real_server.

**Expected behavior**
Initial file reading shall also be effective

**Keepalived version**
Compiled from master branch, containing #1495 

**Distro (please complete the following information):**
 - Name: Alpine
 - Version 3.11
 - Architecture x86_64

**Details of any containerisation or hosted service (e.g. AWS)**
If keepalived is being run in a container or on a hosted service, provide full details

**Configuration file:**
https://pastebin.com/EMVWFs5X

**Notify and track scripts**
If any notify or track scripts are in use, please provide copies of them

**System Log entries**
https://pastebin.com/EMVWFs5X

**Did keepalived coredump?**
If so, can you please provide a stacktrace from the coredump, using gdb.

**Additional context**

libesz:
I've sent a suspected correction. Not tested with vrrp, only with real_server + FILE_CHECK.

pqarmitage:
I have tested a vrrp instance using track_file and it does read the initial state.

I will look at your patch in detail presently. I just wonder if `if (!init)` should be `if (init)`, rather than no check at all. I assume I put that test in for some reason; it might just be that I got it the wrong way around. On the other hand, I don't see that it can do any harm reading all the track files on a reload, and it might be necessary if a new track file is introduced at a reload.

libesz:
Thanks for the comment. Yes I also assumed that the check had some intent. Though I agree it may not be harmful to push forward the update. If the state did not change, I guess the next function (update_track_file_status) will early-return, so no unnecessary cycles are burned.

[...]

PR #1556, opened 2020-04-23
Track file status updated also at init.

libesz:
Otherwise the initial status may be treated as successful (depends on track_file settings) until the first change in the file content.
Fixes: https://github.com/acassen/keepalived/issues/1555


pqarmitage:
This patch is superseded by commit 94ad05a.

Issue #1575, opened 2020-05-19
fined grained namespace support


wdauchy:
Hello,

We have a setup such as:
- physical interface is in the root namespace
- VIP are placed within a dedicated namespace

We would like to be able to do the health checks from the root namespace, and configure ipvs routes in a dedicated namespace. The main advantage is to easily isolate the two types of traffic on a given host and to minimise performance penalties associated with a conventional bridge + veth setup.
For now, we understood that the only possible setup was to move all the processes within a dedicated namespace which lacks of flexibility. Other processes such as HAProxy made the choice to allow users to bind each IP in a specific netns, this allows the healtchecks and backend traffic to happen in the root netns (see https://cbonte.github.io/haproxy-dconv/2.1/configuration.html#5.1-namespace)

Do you think it would be an acceptable feature?

wdauchy:
(@pqarmitage even though I'm waiting for your approval regarding this feature, I started something on my side and might propose a patch for a minimal support of such feature)

pqarmitage:
I'm very happy to have a look at a patch.

Some thoughts that have occurred to me and will/might cause problems:
1. The use of `lvs_sync_daemon` will not work as it does currently, since the configuring of the sync daemon is handled by the VRRP process rather than the health checker process.
1. Will SNMP, JSON and DBus still work?
1. Will communication between the VRRP and BFD process and between the health checker and BFD process still work?
1. One use of keepalived with IPVS is for the virtual server addresses to be managed by the VRRP process, thereby creating master/backup IPVS servers. This will not work with the vrrp process and the healthchecker process in different network namespaces.

These are just a few pointers to some of the issues that may need to be considered when implementing different namespaces.

Is there any disadvantage running two separate instances of keepalived, one in each namespace?

wdauchy:
> 1. The use of `lvs_sync_daemon` will not work as it does currently, since the configuring of the sync daemon is handled by the VRRP process rather than the health checker process.
> 2. Will SNMP, JSON and DBus still work?
> 3. Will communication between the VRRP and BFD process and between the health checker and BFD process still work?
> 4. One use of keepalived with IPVS is for the virtual server addresses to be managed by the VRRP process, thereby creating master/backup IPVS servers. This will not work with the vrrp process and the healthchecker process in different network namespaces.
> 
> These are just a few pointers to some of the issues that may need to be considered when implementing different namespaces.

Thank you very much.
[...]

PR #1576, opened 2020-05-19
ipvs: allow to locate ipvs rules in a namespace

wdauchy:
- this patch adds `net_namespace_ipvs` parameter
- when set, ipvs socket is created in the targeted namespace

it allows to properly split traffic between healthchecks and ipvs
without having to use any complex setup with veth/bridge/iptables:
- do the healthcheck traffic in the root namespace
- receive the VIP traffic in a given namespace

the patch had the necessary plumbing for two cases:
- without libnl, we use a generic `socket_netns` function to create given
  socket - if no namespace has been specified, the effect is the same as
  before, we only create a socket. If a namespace was previously
  specified, we first switch to the namespace, then create the socket,
  and move back to the root namespace
  this function could be used in the future in any other extension while
  opening a socket.
- with libnl, it is about the same operation but we do the namespace
  changes around `genl_connect` as it is the function which creates and
  bind the socket.

All this patch is widely inspired by haproxy implementation which allows
to bind each IP in a given namespace.

this is a proposal for github issue #1575

### DIFF
--- a/doc/man/man5/keepalived.conf.5
+++ b/doc/man/man5/keepalived.conf.5
@@ -128,6 +128,11 @@ and
 # Note: the namespace cannot be changed on a configuration reload.
 \fBnet_namespace \fRNAME
 
+# run all ipvs operations within the mentioned namespace. It allows to easily
+# split the VIP traffic on a given namespace and keep the healthchecks traffic
+# in another namespace (such as the root namespace).
+\fBnet_namespace_ipvs\fR
+
 # ipsets wasn't network namespace aware until Linux 3.13, and so
 # if running with # an earlier version of the kernel, by default
 # use of ipsets is disabled if using a namespace and vrrp_ipsets
--- a/keepalived/check/ipvswrapper.c
+++ b/keepalived/check/ipvswrapper.c
@@ -42,6 +42,9 @@
 #include "logger.h"
 #include "libipvs.h"
 #include "main.h"
+#if HAVE_DECL_CLONE_NEWNET
+#include "namespaces.h"
+#endif
 
 static bool no_ipvs = false;
 
@@ -94,6 +97,12 @@ ipvs_start(void)
 			return IPVS_ERROR;
 		}
 	}
+#if HAVE_DECL_CLONE_NEWNET
+	if (global_data->network_namespace_ipvs && init_ipvs_namespaces(global_data->network_namespace_ipvs) < 0) {
+		log_message(LOG_ERR, "Unable to set network namespace for ipvs %s - exiting", global_data->network_namespace_ipvs);
+		return IPVS_ERROR;
+	}
+#endif
 
 	return IPVS_SUCCESS;
 }
@@ -105,6 +114,10 @@ ipvs_stop(void)
 		return;
 
 	ipvs_close();
+#if HAVE_DECL_CLONE_NEWNET
+	if (global_data->network_namespace_ipvs)
+		close_ipvs_namespaces();
+#endif
 }
 
 void
--- a/keepalived/check/libipvs.c
+++ b/keepalived/check/libipvs.c
@@ -53,6 +53,9 @@
 #include "logger.h"
 #include "utils.h"
 
+#if HAVE_DECL_CLONE_NEWNET
+#include "namespaces.h"
+#endif
 
 typedef struct ipvs_servicedest_s {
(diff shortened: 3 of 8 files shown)


pqarmitage:
I have merged this along with some further changes to:
1. Allow using default network namespace for IPVS
2. Simplify opening and closing the IPVS network namespace
3. Enable building when network namespaces are not supported

Many thanks for producing this patch, and I hope it is now useful for you.

libesz:
This PR broke the compilation on alpine linux (3.11):
```
  CC       track_process.o
In file included from /usr/include/netlink/handlers.h:19,
                 from /usr/include/netlink/socket.h:16,
                 from namespaces.c:171:
/usr/include/netlink/netlink-kernel.h:193:2: error: unknown type name '__u32'
  193 |  __u32 group;
      |  ^~~~~
In file included from /usr/include/netlink/netlink.h:25,
                 from /usr/include/netlink/genl/genl.h:15,
                 from namespaces.c:172:
/usr/include/linux/rtnetlink.h:393:2: error: unknown type name '__kernel_sa_family_t'
  393 |  __kernel_sa_family_t rtvia_family;
      |  ^~~~~~~~~~~~~~~~~~~~
make[2]: *** [Makefile:394: namespaces.o] Error 1
```

I am using the alpine provided build tools, with the only modification that it pulls specific source commit from github, and not the released tgz.
@wdauchy @pqarmitage Can you help to fix it? I can test the compilation if a fix comes up.

libesz:
I figured out adding some linux specific headers resolve the missing types: https://github.com/libesz/keepalived/commit/0662153a99e29f676e3077c85254daa028bdb4f9
Now I have linking errors:
```
  CCLD     keepalived
/usr/lib/gcc/x86_64-alpine-linux-musl/9.2.0/../../../../x86_64-alpine-linux-musl/bin/ld: check/libcheck.a(libipvs.o): in function `open_nl_sock':
/apk/apkbuild/src/keepalived-0662153a99e29f676e3077c85254daa028bdb4f9-snmp/keepalived/check/libipvs.c:339: undefined reference to `nl_socket_modify_err_cb'
/usr/lib/gcc/x86_64-alpine-linux-musl/9.2.0/../../../../x86_64-alpine-linux-musl/bin/ld: check/libcheck.a(libipvs.o): in function `ipvs_nl_send_message':
/apk/apkbuild/src/keepalived-0662153a99e29f676e3077c85254daa028bdb4f9-snmp/keepalived/check/libipvs.c:376: undefined reference to `nl_send_auto'
```

libesz:
After changing the netlink library dependencies from libnl to libnl3 here: https://github.com/alpinelinux/aports/blob/master/community/keepalived/APKBUILD#L12
It succeeded.
Can you confirm this is the solution?

pqarmitage:
@libesz Certainly using libnl3 rather than libnl is correct. I will need to see if I can find a system with the pre ver 3 libnl to see what needs changing.

```In file included from /usr/include/netlink/netlink.h:25,
                 from /usr/include/netlink/genl/genl.h:15,
                 from namespaces.c:172:
/usr/include/linux/rtnetlink.h:393:2: error: unknown type name '__kernel_sa_family_t'
  393 |  __kernel_sa_family_t rtvia_family;
```
Your fix for this is I think to include `linux/socket.h`, however, on my  Alpine 3.11 system /usr/include/linux/rtnetlink.h includes /usr/include/linux/netlink.h which includes at line 6 /usr/include/linux/socket.h (with a comment /* for __kernel_sa_family_t */). I can only assume that you have a different set of kernel headers on your system; the kernel headers I have are for 4.19.36.

```
  CC       track_process.o
In file included from /usr/include/netlink/handlers.h:19,
                 from /usr/include/netlink/socket.h:16,
                 from namespaces.c:171:
/usr/include/netlink/netlink-kernel.h:193:2: error: unknown type name '__u32'
  193 |  __u32 group;
```
This should be resolved by using libnl3.

libesz:
Correct, reverting the imports and using libnl3 works. Thanks! Will send a notification about this to the alpine package maintainers as soon new keepalived release is announced. So far it compiled also with libnl.

pqarmitage:
@libesz libnl3 should have been used in preference to libnl for many years now. It is probably worth notifying the alpine package maintainers now rather than waiting for the next release of keepalived.

PR #1608, opened 2020-06-06
build: fix and simplify NetSNMP compilation

vincentbernat:
First, this reverts changes made in #416 (commit 328ec2824022). With
lldpd, I don't use `--netsnmp-agent-libs --external-libs` and I never
had an issue with Ubuntu 14.04. From my understanding, the code to fix
mangle `NETSNMP_LIBS` to fix issues with Fedora is then not needed
anymore because the issue appears only because of `--external-libs`,
so the code can be removed as well. This was added in #858 (commit
057c279fb843).

Then, this also reverts changes introduced in #1262, notably commit
63fe1f4c7d3e. Seperating CPPFLAGS from CFLAGS also reorder some flags,
making headers in `/usr/lib/*/perl/*/CORE` used before the ones in
`lib/`. This happens for `parser.h` and leads to this error when
NetSNMP is linked to libperl. This happens with out-of-tree builds
when `-I$(srcdir)/../include -I$(srcdir)/../../lib` is appended too
late.

```
In file included from ../../../keepalived/core/main.c:54:
/usr/lib/x86_64-linux-gnu/perl/5.30/CORE/parser.h:15:5: error: unknown type name ‘YYSTYPE’
   15 |     YYSTYPE val;    /* semantic value */
      |     ^~~~~~~
```

After investigating a bit, I have a less invasive fix to propose. I'll send another one with this fix. I keep this PR here to maybe have a change to discuss the first change (and check everything passes in Travis).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2156,33 +2156,9 @@ if test "$enable_snmp" = yes -o \
   if test "$NETSNMP_CONFIG" = no; then
     AC_MSG_ERROR([*** unable to find net-snmp-config])
   fi
-  NETSNMP_LIBS_AGENT=`${NETSNMP_CONFIG} --netsnmp-agent-libs`
-  NETSNMP_LIBS_EXT=`${NETSNMP_CONFIG} --external-libs`
-  NETSNMP_LIBS="$NETSNMP_LIBS_AGENT $NETSNMP_LIBS_EXT"
+  NETSNMP_LIBS=`${NETSNMP_CONFIG} --agent-libs`
   NETSNMP_CFLAGS="`${NETSNMP_CONFIG} --base-cflags`"
   NETSNMP_CPPFLAGS="-DNETSNMP_NO_INLINE"
-
-  # net-snmp-config adds compiler and linker options that were set at the time
-  # net-snmp was built, and this can include spec files that may not exist
-  # on the system building keepalived. We need to check if any spec files
-  # are specified, and if they do not exist on this system, then remove them
-  # from NETSNMP_LIBS or NETSNMP_CFLAGS.
-  # For further information, see https://bugzilla.redhat.com/show_bug.cgi?id=1544527
-  # and the other bugs referred to in it.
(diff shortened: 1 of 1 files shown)


vincentbernat:
I have pushed a less invasive fix in #1609. However, I still think the first change of this PR (not using `--external-libs` is worth discussing).

pqarmitage:
On Fedora, the output of `net-snmp-config --agent-libs` is:
`-Wl,-z,relro -Wl,--as-needed -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -lm -L/usr/lib64 -lnetsnmpmibs -lsensors -ldl -lrpm -lrpmio -lnetsnmpagent -Wl,--enable-new-dtags -Wl,-z,relro -Wl,--as-needed -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -lnetsnmp -lssl -lssl -lcrypto -lm`

`/usr/lib/rpm/redhat/redhat-hardened-ld` is not installed by default, nor is it installed by any of the packages that keepalived needs in order to build, so it seems that  #858 (commit 057c279) is still needed. Also, `-lrpm -lrpmio` seem unnecessary, and these are not included with `--external-libs` or `--netsnmp-agent-libs`.

So I think, at least for RedHat based systems, checking for the -specs= files is necessary, and using `--external-libs` and --netsnmp-agent-libs` is better than `--agent-libs`, and if it works for other distros I am included to leave it as it is.

vincentbernat:
OK, I just read the bugzilla. I never noticed the problem as I suppose my build environment (OBS) is including the missing package. https://build.opensuse.org/public/build/home:vbernat/Fedora_33/x86_64/lldpd/_log

NetSNMP upstream is difficult to work with. The problem around the invasive flags needed to build an agent lasts forever and any solution is rejected.

pqarmitage:
I've just had a look at https://build.opensuse.org/public/build/home:vbernat/Fedora_33/x86_64/lldpd/_log and indeed rpm-libs and redhat-rpm-config appear to be installed, which are the two packages that provide the extra files.

PR #1609, opened 2020-06-06
build: fix out-of-tree builds when SNMP is enabled

vincentbernat:
NetSNMP will add Perl headers to the search path in CPPFLAGS. It
contains `parser.h` which will conflict with Keepalived's one. Just
move the search path for other directories before the remaining
CPPFLAGS to fix this.

### DIFF
--- a/genhash/Makefile.am
+++ b/genhash/Makefile.am
@@ -4,13 +4,13 @@
 #
 # Copyright (C) 2001-2017 Alexandre Cassen, <acassen@gmail.com>
 
-AM_CPPFLAGS		= $(KA_CPPFLAGS) $(DEBUG_CPPFLAGS)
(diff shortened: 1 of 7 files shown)


pqarmitage:
I prefer this commit, for now at least.

Would it be reasonable to use:
`AM_CPPFLAGS		= -I$(top_srcdir)/keepalived/include -I $(top_srcdir)/lib`
everywhere (except for lib and genhash which just use):
`AM_CPPFLAGS		= -I $(top_srcdir)/lib`
since it gives more consistency?

vincentbernat:
Yes, maybe this will be easier to use that. Note you sometimes put a space after `-I`, sometimes not.

pqarmitage:
Yes, I noticed that. It should be consistent now following commit 0e2fc82.

PR #1614, opened 2020-06-09
Fix Document bug: Update configuration_synopsis.rst

dev-itsheng:
The keyword `smtp_connect_timeout` in table is wrongly `smtp_connection_timeout`, so fix it.

### DIFF
--- a/doc/source/configuration_synopsis.rst
+++ b/doc/source/configuration_synopsis.rst
@@ -27,7 +27,7 @@ global_defs                 identify the global def configuration block
 notification_email          email accounts that will receive the notification mail  List
 notification_email_from     email to use when processing “MAIL FROM:” SMTP command  List
 smtp_server remote SMTP     server to use for sending mail notifications            alphanum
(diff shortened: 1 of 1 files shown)


pqarmitage:
Many thanks for these updates. Currently the documentation in the doc/source tree is way out of date. The up-to-date configuration documentation is in doc/man/man5/keepalived.conf.5. If you would like to update configuration_synopsis.rst from doc/man/man5/keepalived.conf.5 the updates would be very greatfully received :).

dev-itsheng:
It's the least i could do.

In fact, I saw the [old doc](https://www.keepalived.org/doc/configuration_synopsis.html) from Bing, so when I found some wrong in page, I try to search the keyword from this repo and fix it.

If the old doc pages are still display in website, I'd glad to update it from new doc.

pqarmitage:
The web site currently has no link to the old doc pages since they are so out of date. If the pages can be brought up-to-date, then the link can be restored on the website.

The old documentation pages are still accessible at https://www.keepalived.org/doc/.

PR #1627, opened 2020-06-16
Fix some format of new document

dev-itsheng:
I found some format problem in new document, such as:

1.  Four propertys' name aren't be bold.

    
    image

    image

2.  Some indentations are Tabs instead of spaces, resulting in inconsistent display.

    image

3.  Some spaces of indentations whose quantity is not 4 but 2. 

    image

So I try to fix it.


pqarmitage:
@dev-itsheng Many thanks for your help with the keepalived project.

PR #1632, opened 2020-06-18
bugfix epoll_count is reversed

zhouxudong199:
fix issue#1631
keepalived running too long，which lead to print “scheduler：epoll_wait error ：invalid argument”

### DIFF
--- a/lib/scheduler.c
+++ b/lib/scheduler.c
@@ -649,6 +649,7 @@ thread_event_cancel(const thread_t *thread_cp)
 	rb_erase(&event->n, &m->io_events);
 	if (event == m->current_event)
 		m->current_event = NULL;
+	thread_events_resize(m, -1);
 	FREE(thread->event);
 	return 0;
 }


pqarmitage:
Many thanks for diagnosing the problem and providing the solution.

TowerLee:
Whether another problem is caused, the memory usage to increase continuously,  for func  thread_envents_resize para delta is always 1, malloc size  increases each time

pqarmitage:
@TowerLee There are two separate places where `thread_events_resize()` is called, in function `thread_event_new()` and in function `thread_event_cancel()`. The call in `thread_event_cancel()` has a delta value of -1.

I am very happy to investigate this further, and add some diagnostics to check what is happening, but in order for me to do so meaningfully you will need to provide a copy of your configuration, and what you are doing to see that the memory usage is increasing.

You state that with the delta being `1, the malloc size increases each time`. I am not sure this is the case; the malloc size will only increase if `!(m->epoll_count < m->epoll_size)`.

From a quick look at the code, the value of m->epoll_count will be limited by the number of open file descriptors, and therefore m->epoll_size <= m->epoll_count + THREAD_EPOLL_REALLOC_THRESH.

For optimal memory efficiency, we want m->epoll_size <= m->epoll_count + THREAD_EPOLL_REALLOC_THRESH **- 1**. This is most simply resolved by changing
```
        if (m->epoll_count < m->epoll_size)
                return 0;
```
to
```
        if (m->epoll_count <= m->epoll_size)
                return 0;
```

TowerLee:
Thanks, get it ,for my code version is not include mr 6867661

zhouxudong199:
Sorry , I read this email just now.
Glad to help you.

发件人: TowerLee [mailto:notifications@github.com]
发送时间: 2021年1月7日 9:14
收件人: acassen/keepalived <keepalived@noreply.github.com>
抄送: Zhouxudong (EulerOS) <zhouxudong8@huawei.com>; Author <author@noreply.github.com>
主题: Re: [acassen/keepalived] bugfix epoll_count is reversed (#1632)


Thanks, get it ,for my code version is not include mr 6867661<https://github.com/acassen/keepalived/commit/68676610261f7d21ee93fde79401d3f6ffa18f84>

Issue #1637, opened 2020-06-19
Shutdown real server ，keepalived will not offline real server using dns checker


wangxp006:
**Describe the bug**
Shutdown real server ，keepalived will not offline real server using dns checker

**To Reproduce**

**startup keepalived  and check real server online status**
[root@openEuler ~]# ipvsadm -Ln
IP Virtual Server version 1.2.1 (size=4096)
Prot LocalAddress:Port Scheduler Flags
  -> RemoteAddress:Port           Forward Weight ActiveConn InActConn
UDP  22.22.54.6:53 rr
  -> 22.22.73.181:53   

login  real server  and delete the ip address
ip addr del dev enp4s0 22.22.73.181/16

real server will not offline  in keepalived
[root@openEuler ~]# ipvsadm -Ln
IP Virtual Server version 1.2.1 (size=4096)
Prot LocalAddress:Port Scheduler Flags
  -> RemoteAddress:Port           Forward Weight ActiveConn InActConn
UDP  22.22.54.6:53 rr
  -> 22.22.73.181:53    

**Expected behavior**
real server will offline 

[root@openEuler ~]# ipvsadm -Ln
IP Virtual Server version 1.2.1 (size=4096)
Prot LocalAddress:Port Scheduler Flags
  -> RemoteAddress:Port           Forward Weight ActiveConn InActConn
UDP  22.22.54.6:53 rr

**Keepalived version**
keepalived-2.0.20

**Details of any containerisation or hosted service (e.g. AWS)**
If keepalived is being run in a container or on a hosted service, provide full details

**Configuration file:**

keepalived.conf Configuration File :

! Configuration File for keepalived
global_defs {
        checker_log_all_failures on
 }

virtual_server 22.22.54.6 53 {
    delay_loop 7
    lb_algo rr
    lb_kind DR
    protocol UDP
    #alpha
    hysteresis 0
    quorum 1
    quorum_up "/usr/sbin/ip addr add 22.22.54.6 dev lo"

    real_server 22.22.73.181 53 {
        weight 2
        DNS_CHECK {
            connect_port 53
            connect_timeout 3
            retry 2
            type A
            name lvs-test.com
            bindto 22.22.1.69
        }
    }
}

wangxp006:
Solution is here #1638

pqarmitage:
@wangxp006 Thank you very much for identifying this bug and finding the root cause of the problem; however I don't think your patch #1638 quite resolves the issue in a fully safe way.

[...]

PR #1638, opened 2020-06-19
Fix abnormally DNS checker timer after RS port is down

wangxp006:
thread->sands is letter than time_now, timeout will be a extremely large value。
Timeout should be assigned to the value configured in the configuration file


pqarmitage:
Please see commits 7898118 (and 48c2e6e) as described in issue #1637.

PR #1640, opened 2020-06-20
fix-persistence_timeout-range

zhouxudong199:
although it's type is unsigned ,timeout is must <=UINT_MAX / HZ.  ip_vs_add_service function in ipvs module  check it by u->timeout > (UINT_MAX / HZ).  LVS_MAX_TIMEOUT value is 
appropriate. at the same time ipvsadm use 31 days always.

### DIFF
--- a/keepalived/check/check_parser.c
+++ b/keepalived/check/check_parser.c
@@ -506,7 +506,7 @@ pto_handler(const vector_t *strvec)
 		return;
 	}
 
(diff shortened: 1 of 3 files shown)


zhouxudong199:
sloved problem detail https://github.com/acassen/keepalived/issues/1641

thanks for your review

pqarmitage:
@zhouxudong199  Many thanks for this fix.

PR #1645, opened 2020-06-22
change same lvs_sync_daemon keys range

zhouxudong199:
because  donnot know which is reasonable. I commit for each one.
and each commit log has my understanding.
thanks for your time.
these new min values , are same as keepalived/manpage ,and same as ipvsadm.


zhouxudong199:
problem deteils in #1646

PR #1654, opened 2020-06-25
Fix build with Linux kernel headers v4.15

ffontaine:
Same issue that was fixed by 947248af144bcab6376ccddab8dc40f313b14281 but raised again when bumping keepalived from version 2.0.15 to 2.1.2 in buildroot

Fixes:
 - http://autobuild.buildroot.org/results/db7f149f63e9180b22460caa74850673362aa17c

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/keepalived/vrrp/vrrp_arp.c
+++ b/keepalived/vrrp/vrrp_arp.c
@@ -23,6 +23,7 @@
 #include "config.h"
 
 /* system includes */
+#include <netinet/in.h>
 #include <unistd.h>
 #include <net/ethernet.h>
 #include <net/if_arp.h>
--- a/keepalived/vrrp/vrrp_ndisc.c
+++ b/keepalived/vrrp/vrrp_ndisc.c
@@ -23,11 +23,11 @@
 #include "config.h"
 
 /* system includes */
+#include <netinet/in.h>
 #include <unistd.h>
 #include <net/ethernet.h>
 #include <linux/if_packet.h>
 #include <netinet/icmp6.h>
-#include <netinet/in.h>
 #include <stdint.h>
 #include <errno.h>
 
--- a/keepalived/vrrp/vrrp_vmac.c
+++ b/keepalived/vrrp/vrrp_vmac.c
@@ -23,6 +23,7 @@
 #include "config.h"
 
 /* global include */
+#include <netinet/in.h>
 #ifdef NETLINK_H_NEEDS_SYS_SOCKET_H
 #include <sys/socket.h>
 #endif


pqarmitage:
This is very strange and I would like to understand the problem further before merging this patch. I can build keepalived successfully with Fedora Linux 4.15 headers using the `--with-kernel-dir` configure option. Also snap builds with Linux 4.15 headers (using Ubuntu kernel headers) works OK, see for example https://launchpad.net/~build.snapcraft.io/+snap/500e626278a3b3940f697bd21072402d/+build/1015704/+files/buildlog_snap_ubuntu_xenial_amd64_500e626278a3b3940f697bd21072402d_BUILDING.txt.gz which builds keepalived with multiple different kernel versions.

So the problem isn't simply building with 4.15 kernel headers, and we need to identify the real cause of the problem. Possibly there is a cross compilation problem, but since keepalived builds natively with Linux 4.15 headers that might suggest we have some cross compilation issues that need resolving.

It would be good if we could work together to find the root cause of the problem, and then fix that. It may be a combination of versions of kernel and libc headers. Are you using musl libc headers? it may be that we need to resolve a combination in relation to that combination.

I have noticed the following entries in the build log you linked to:
```
configure: WARNING: using cross tools not prefixed with host triplet
configure: WARNING: Cannot determine if unaligned access supported. Assuming yes.
configure: WARNING: Cannot determine if need to OPENSSL_init_crypto() problem. Assuming yes for safety.
configure: WARNING: Cannot determine if need to call g_type_init(). Assuming yes for safety.
```
and we should resolve these for cross compilation.

Also
```
configure: WARNING: unrecognized options: --disable-gtk-doc, --disable-gtk-doc-html, --disable-doc, --disable-docs, --disable-documentation, --with-xmlto, --with-fop, --enable-ipv6, --disable-nls, --disable-static, --enable-shared, --disable-libiptc
```
should be resolved.

`*** WARNING - this build will not support IPVS with IPv6. Please install libnl/libnl-3 dev libraries to support IPv6 with IPVS.` should be resolved.

Please do not push keepalived v2.1.2 since we have found a couple of problems since then, which are fixed in v2.1.3, but we have found another problem since and so you should use commit 37054e6 or later. We will release v2.1.4 in the next few days after further testing.

ffontaine:
> It would be good if we could work together to find the root cause of the problem, and then fix that. It may be a combination of versions of kernel and libc headers. Are you using musl libc headers? it may be that we need to resolve a combination in relation to that combination.

For now, we have only a single build failure for nios architecture with a glibc toolchain using gcc 7.3.1 and kernel headers 4.15: 
- http://autobuild.buildroot.org/?reason=keepalived-2.1.2
- http://autobuild.buildroot.org/results/db7/db7f149f63e9180b22460caa74850673362aa17c//keepalived-2.1.2/config.log
> 
> I have noticed the following entries in the build log you linked to:
> 
> ```
> configure: WARNING: using cross tools not prefixed with host triplet
> configure: WARNING: Cannot determine if unaligned access supported. Assuming yes.
> configure: WARNING: Cannot determine if need to OPENSSL_init_crypto() problem. Assuming yes for safety.
> configure: WARNING: Cannot determine if need to call g_type_init(). Assuming yes for safety.
> ```
First WARNING is ok, it is raised because buildroot passes `--host=nios2-buildroot-linux-gnu`.
Last WARNINGs are due to the usage of `AC_RUN_IFELSE` in configure.ac. For obvious reasons, you can't run anything when cross-compiling.
> 
> ```
> configure: WARNING: unrecognized options: --disable-gtk-doc, --disable-gtk-doc-html, --disable-doc, --disable-docs, --disable-documentation, --with-xmlto, --with-fop, --enable-ipv6, --disable-nls, --disable-static, --enable-shared, --disable-libiptc
> ```
 buildroot is passing a bunch of default options to all autotools package (e.g. `--disable-gtk-doc, --disable-gtk-doc-html, --disable-doc, --disable-docs, --disable-documentation, --with-xmlto, --with-fop, --enable-ipv6, --disable-nls, --disable-static, --enable-shared`)
 
However, `--{enable,disable}-iptc` was added back in 2017 when adding version 1.3.2: https://git.buildroot.net/buildroot/commit/?id=9a7ac39605e1bf416db033b009e6c09f54f44fdf. I'll fix it.

> `*** WARNING - this build will not support IPVS with IPv6. Please install libnl/libnl-3 dev libraries to support IPv6 with IPVS.` should be resolved.

libnl is currently only optional with keepalived, so we don't enable it by default on buildroot as we're building embedded systems. If needed, you can make it mandatory in configure.ac.

I'll try to investigate a little more to find the root cause perhaps @baruchsiach can help?

pqarmitage:
This whole issue is very strange so I have set up a buildroot build environment to track down what is going on.
  
First of all I decided to target keepalived/vrrp/vrrp_arp.c compilation failure. In order to see what the preprocessor was doing, I added `-E -dD -dI -save-temps` to CPPFLAGS in a copy of keepalived/vrrp/Makefile - called Makefile.pqa. Then in keepalvied/vrrp, execute `make -f Makefile.pqa vrrp_arp.o`. vrrp_arp.o is now a text file with information about what the preprocessor is doing.

I looked at what was causing the duplicate definition of IPPROTO_IGMP. What is strange is that the error produced is:
```
/home/buildroot/autobuild/instance-0/output-1/host/nios2-buildroot-linux-gnu/sysroot/usr/include/linux/in.h:33:3: error: redeclaration of enumerator 'IPPROTO_IGMP'
   IPPROTO_IGMP = 2,  /* Internet Group Management Protocol */
   ^
/home/buildroot/autobuild/instance-0/output-1/host/nios2-buildroot-linux-gnu/sysroot/usr/include/netinet/in.h:46:5: note: previous definition of 'IPPROTO_IGMP' was here
     IPPROTO_IGMP = 2,    /* Internet Group Management Protocol. */
```
and given the include collisions that <linux/libc-compat.h> is intended to afford, including <netinet/in.h> BEFORE <linux/in.h> should be the right way to do it, and indeed a trivial test program shows that that is the way around that works.
The definition that controls whether we get the problem or not is `__UAPI_DEF_IN_IPPROTO`.

Searching for libc-compat.h in vrrp_arp.o helps highlight the cause of the problem.

It turns out that the problem is resolved by kernel commit da360299b673 (which was first in Linux 4.16), which stopped <linux/if_ether.h> including <linux/libc-compat.h>. The problem comes about because the double-definition protection definitions that <linux/libc-compat.h> sets up are set for all header files it helps on the first time it is included, so for example:
```
#include <linux/if.h>
#include <netinet/in.h>
#include <linux/in.h>
```
reports duplicate defintions between <netinet/in.h> and <linux/in.h>, whereas
```
#include <netinet/in.h>
#include <linux/in.h>
```
does not cause duplicate definitions. This is because in the first case the definition of `__UAPI_DEF_IN_IPPROTO` is established when <linux/if.h> is included, and it is set to 1 since <netinet/in.h> has not been included, whereas in the second case, the definition of `__UAPI_DEF_IN_IPPROTO` is established when <linux/in.h> is included and it is set to 0.

So why doesn't the problem occur when building snaps using Linux 4.15 headers? The reason is that Linux 4.15.7 resolved the issue, the headers used for building the keepalived snap are 4.15.18, and the kernel headers in the codesourcery-niosII toolchain are version 4.15.5.
@ffontaine Is there a reason you are still using such an old version of the Codesourcery Nios II toolchain? There have been 8 releases since the 2018.05 version that you are using.

Now that we understand why the includes of <netinet/in.h> are necessary when including early versions of Linux 4.15 headers, I am happy to include them, although I will add comments explaining why.

The next question was, why did the addition of `#include <netinet/in.h>` from commit 947248a "disappear` since `git log -p keepalived/include/vrrp_arp.h` shows no patch that removed it. If transpires that it was merge commit 37d98e8 resolving a merge conflict.

So, now we can incorporate the patch, but I will add some comments explaining the need.

pqarmitage:
Further to your comments above:

>> `*** WARNING - this build will not support IPVS with IPv6. Please install libnl/libnl-3 dev libraries to support IPv6 with IPVS.` should be resolved.

>libnl is currently only optional with keepalived, so we don't enable it by default on buildroot as we're building embedded systems. If needed, you can make it mandatory in configure.ac.

The reason libnl is only optional is that if it is not available we fall back to using the original socket interface for configuring IPVS. However, the socket interface does not support IPv6 configuration, and therefore libnl (and preferably libnl3) should be installed if it is available, in order to enable the full functionality, hence the warning that IPVS IPv6 functionality will not be supported.

Since we support systems back to Linux 2.6.19 at the moment we have to be very flexible in what to expect in the build environment. However, where enhanced functionality is supported by installing additional packages, those packages really should be installed.

Similarly, with the buildroot config for keepalived neither iptables nor nftables support is enabled due to the development packages not being installed for the buildroot build (this appears to be part of the .config created). Again, superior functionality is achieved through having at least one of them installed (preferably nftables, but if iptables is used, then ipsets should be enabled if possible). My view is that if you want keepalived to work properly, then the keepalived package should select BR2_PACKAGE_LIBNL and BR2_PACKAGE_LIBNFTNL.

Perhaps for where distros are building keepalived as a package, we should require, for example, an explicit `--without-libnl` and `--without-firewall`, and even `--without-ipsets` if iptables is being used, if those features are not wanted. It has concerned me for some time that various distros are building keepalived in a sub-optimal way due to the lack of installing certain packages, and we need to find a way to resolve that.

pqarmitage:
Commit d47ae3b is an update of your commit 6ad62a6, with only including <netinet/in.h> if it is necessary.

Once we have released keepalived v2.1.4 I will update this log.

pqarmitage:
On my build of buildroot I found another problem with building keepalived relating to autoconf defining `malloc` as rpl_malloc and likewise for `realloc` resulting to the linker report undefined symbols `rpl_malloc` and `rpl_realloc`. Commit cb07929 resolves this problem, and following this commit keepalived now builds successfully in my buildroot environment for the niosII architecture.

pqarmitage:
@ffontaine keepalived v2.1.4 has now been released.

PR #1662, opened 2020-06-27
[doc] MCast address normally is in 244.0.0.0/24

wodry:
Clearify, that in normal case, a Keepalived multicast group address is of RFC5771 type "Local subnetwork" address and therefore must be in 244.0.0.0/24.

### DIFF
--- a/doc/man/man5/keepalived.conf.5
+++ b/doc/man/man5/keepalived.conf.5
@@ -233,8 +233,8 @@ and
     # (default: local host name)
     \fBrouter_id \fR<STRING>
 
-    # Multicast Group to use for IPv4 VRRP adverts
-    # (default: 224.0.0.18)
+    # Multicast Group to use for IPv4 VRRP adverts (default: 224.0.0.18).
+    # An RFC5771 "Local subnetwork" address must be in 244.0.0.0/24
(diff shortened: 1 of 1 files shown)


pqarmitage:
Unfortunately we can't merge this pull request since where you have stated 244.0.0.0/24 it should be 224.0.0.0/24.

Also, whereas the title of this pull request is "MCast address **normally** is in 244.0.0.0/24", the actual patch says that the address **MUST** be in 244.0.0.0/24.

My view is that if anyone is wanting to change the multicast address (which of course then doesn't comply with RFC5798), then it can be presumed that they know what they are doing,  and so don't need to be told what the valid addresses are. However, I am open to being convinced to the contrary, 

Whereas https://www.iana.org/assignments/multicast-addresses/multicast-addresses.xhtml states `The range of addresses between 224.0.0.0 and 224.0.0.255, inclusive, is reserved for the use of routing protocols and other low-level topology discovery or maintenance protocols, such as gateway discovery and group membership reporting`, it doesn't state that _ONLY_ 224.0.0.0/24 can be used for that purpose.

Before adding that only 224.0.0.0/24 can be used, I would want to see evidence that no other addresses in 224.0.0.0/4 would work with Linux.

PR #1663, opened 2020-06-28
[doc]  Typically don't change default MCast address

wodry:
Since I have two keepalived clusters (old one and new one) in the same LAN in parallel,
I wanted to be sure that they do not intefere with VRRP messages. Since authentication is deprecated and should no be used,
I did not want to only rely on the virtual router ID, but also use a separate multicast address for the second cluster.
So I set vrrp_mcast_group4 to 224.0.**10**.18 without researching multicast address assigning (and here in the doc was nothing said about which address one could set).
But with 224.0.**10**.18 there came the message "Receive advertisement timeout" and both nodes had the virtual IP addresses (split brain).
So, VRRP simply did not work.
Then I researched and saw that the network where the default address 224.0.0.18 is in, is 224.0.0.0/24.
So I changed vrrp_mcast_group4 to 224.0.0.**2**8 and with this address, VRRP works.

I do not know what other networks could work also, but at least address 224.0.**10**.18 does not work.
So, with "normal" I mean using the network in which the default address is in.

### DIFF
--- a/doc/man/man5/keepalived.conf.5
+++ b/doc/man/man5/keepalived.conf.5
@@ -234,7 +234,8 @@ and
     \fBrouter_id \fR<STRING>
 
     # Multicast Group to use for IPv4 VRRP adverts
-    # (default: 224.0.0.18)
+    # Defaults to the RFC5798 IANA assigned VRRP multicast address 224.0.0.18
+    # which You typically do not want to change.
     \fBvrrp_mcast_group4 \fR224.0.0.18
 
     # Multicast Group to use for IPv6 VRRP adverts


wodry:
Unfortunateley, I could not manage to squash and rebase my commit  in PR #1662 (my patch-4 repo), so I created a new PR.

wodry:
If only address 224.0.0.18 complies with VRRP specification (RFC5798), I guess I will stick with the default, to be safe. I wonder, who will ever want to set some other address against the spec.
So I have to rely on that the VRRP packets do not interfere because they have different vrouter-IDs?
Or what is best practice to separate multiple VRRP instances?

pqarmitage:
I too have wondered why anyone would want to use a different multicast address, but it was easy to provide the option of specifying it, and so why not allow it?

Using different vrouter-IDs (VRIDs) is exactly the way to do it, and this usage is the sole reason why VRIDs exist. I permanently have two different IPv4 VRIDs and one IPv6 VRID in use on my main network, so there is absolutely no problem with doing as you have suggested and using different VRIDs.

wodry:
OK, that's very valuable information. Suqashed and rebased, so users can benefit.

PR #1660, opened 2020-06-27
Support rollback if necessary configuration is missing during reload

wangxp006:
After the user modifies the keepalived configuration, the process of keepalived will exit if necessary configuration items are missing during the reload process.

For example,  configuration is as follows：
virtual_server 22.22.22.237 80 {
    delay_loop 6
    lb_algo rr
    lb_kind DR
    protocol TCP
    #alpha
    hysteresis 0
    quorum 1
    quorum_up "/usr/sbin/ip addr add 22.22.22.237 dev lo"

    real_server 22.22.83.49 80 {
        weight 1
        HTTP_GET {
            url {
                path /
            }
            connect_port    **# missing checker port**
            connect_timeout 5
            retry 3
            delay_before_retry 3
        }
    }
}
[root@openEuler keepalived]# systemctl status keepalived
● keepalived.service - LVS and VRRP High Availability Monitor
   Loaded: loaded (/usr/lib/systemd/system/keepalived.service; disabled; vendor preset: disabled)
   Active: inactive (dead)

Jun 27 20:58:10 openEuler Keepalived_healthcheckers[28594]: Opening file '/etc/keepalived/keepalived.conf'.
Jun 27 20:58:10 openEuler Keepalived_healthcheckers[28594]: (Line 26) *** Configuration line starting `connect_port` is missing a parameter after keyword `connect_port` at word position 2
Jun 27 20:58:10 openEuler Keepalived_healthcheckers[28594]: (Line 26) WARNING - number '' outside range [1, 65535]
Jun 27 20:58:10 openEuler Keepalived_healthcheckers[28594]: (Line 26) *** Configuration line starting `connect_port` is missing a parameter after keyword `connect_port` at word position 2
Jun 27 20:58:10 openEuler Keepalived_healthcheckers[28594]: (Line 26) Invalid checker connect_port ''
Jun 27 20:58:10 openEuler Keepalived[28593]: pid 28594 exited with permanent error CONFIG. Terminating
Jun 27 20:58:10 openEuler Keepalived[28593]: CPU usage (self/children) user: 0.000000/0.000000 system: 0.002874/0.014846
Jun 27 20:58:10 openEuler systemd[1]: Reloaded LVS and VRRP High Availability Monitor.
Jun 27 20:58:10 openEuler Keepalived[28593]: Stopped Keepalived v2.1.3 (unknown)
Jun 27 20:58:10 openEuler systemd[1]: keepalived.service: Succeeded.

[...]

PR #1669, opened 2020-06-29
Support rollback if necessary configuration is missing during reload

wangxp006:
This PR is to fix the problem mentioned in #1660

### DIFF
--- a/keepalived/bfd/bfd_daemon.c
+++ b/keepalived/bfd/bfd_daemon.c
@@ -163,8 +163,10 @@ start_bfd(__attribute__((unused)) data_t *prev_global_data)
 		set_process_name(global_data->bfd_process_name);
 
 	/* If we are just testing the configuration, then we terminate now */
-	if (__test_bit(CONFIG_TEST_BIT, &debug))
+	if (__test_bit(CONFIG_TEST_BIT, &debug)) {
+		free_bfd_data(bfd_data);
 		return;
+	}
 	bfd_complete_init();
 
 	/* Post initializations */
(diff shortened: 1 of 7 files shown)


wangxp006:
Hi, is there anything else in this PR that needs to be modified, or is this requirement worthless? if you have any suggestions, please let me know。

pqarmitage:
I have been thinking about the patch over the last few weeks and trying to work out the best way forward.

I think there are a number of problems with the patch:
1. It leaks a not insignificant amount of memory
1. It only checks if a required parameter is missing, and not whether there is other invalid configuration -  e.g. a vrrp instance specifies a non-existent interface
1. It makes the parent process start monitoring netlink messages for interface, address, route etc changes
1. The global_data of the parent process is updated to the new config regardless of whether the new config is loaded
1. Not only is `CONFIG_MISSING_VAULE` misspelt, but it also isn't used
1. This feature needs to be enabled by a global_defs keyword, e.g. `test_config_before_reload`
1. It would be helpful to log that the reload failed.

I don't think this approach is likely to work, and even if all the issues identified above were resolved, the potential for modifying the parent process when a reload is aborted is too risky.

The only way I can see this safely being implemented is if the parent process forks another keepalived instance running it as `keepalived --config-test=/dev/null -f CONFIG_FILE ...` where ... is all relevant command line options, e.g. -P, -C -B -s, -i, -x (and possibly others). The parent process then uses the exit code of that process to determine if it is safe to reload the configuration. If might be useful to be able to specify a file name after `test_config_before_reload` and use that instead of /dev/null in `--config-test=/dev/null`, so that any errors can be checked by the user.

wangxp006:
Thank you for your valuable comments. I will close this PR and revise it according to your comments

Issue #1679, opened 2020-07-02
tcp checker and dns checker will not go offline when interface link to RS is down


wangxp006:
**Describe the bug**
tcp checker and dns checker will not go offline when interface link to RS is down

**To Reproduce**
1. startup keepalived
2. delete ip address of the interface linked to RS
TCP  and DNS checker will not go offline 
[root@openEuler keepalived]# ipvsadm -Ln
IP Virtual Server version 1.2.1 (size=4096)
Prot LocalAddress:Port Scheduler Flags
  -> RemoteAddress:Port           Forward Weight ActiveConn InActConn
TCP  22.22.94.33:5001 rr
  -> 22.22.103.237:5001           Route   1      0          0         
UDP  22.22.94.33:53 rr
  -> 22.22.98.217:53              Route   2      0          0   

**Expected behavior**
[root@openEuler keepalived]# ipvsadm -Ln                            
IP Virtual Server version 1.2.1 (size=4096)
Prot LocalAddress:Port Scheduler Flags
  -> RemoteAddress:Port           Forward Weight ActiveConn InActConn
TCP  22.22.94.33:5001 rr
UDP  22.22.94.33:53 rr

**Keepalived version**
keepalived-2.0.20

**Configuration file:**
```
global_defs {
        checker_log_all_failures on
 }


virtual_server 22.22.94.33 53 {
    delay_loop 6
    lb_algo rr
    lb_kind DR
    protocol UDP
    alpha
    hysteresis 0
    quorum 1
    quorum_up "/usr/sbin/ip addr add 22.22.94.33 dev lo"

    real_server 22.22.98.217 53 {
        weight 2
        DNS_CHECK {
            connect_port 53
            connect_timeout 1
            retry 1
            type A
            name lvs-test.com
            bindto 22.22.116.103
        }
    }
}

virtual_server 22.22.94.33 5001 {
    delay_loop 6
    lb_algo rr
    lb_kind DR
    protocol TCP
    alpha
    hysteresis 0
    quorum 1
    quorum_up "/usr/sbin/ip addr add 22.22.94.33 dev lo"

    real_server 22.22.103.237 5001 {
        weight 1
        TCP_CHECK {
            connect_port 22
            connect_timeout 10
            retry 3
            delay_before_retry 3
            bindto 22.22.116.103
        }
    }

}
```

**System Log entries**
[root@openEuler keepalived]# grep "22.22.98.217" /var/log/messages 
[...]

PR #1680, opened 2020-07-02
tcp checker and dns checker will not go offline when interface link to RS is down

wangxp006:
tcp checker and dns checker will not go offline when interface link to RS is down
Detail is here #1679

### DIFF
--- a/keepalived/check/check_dns.c
+++ b/keepalived/check/check_dns.c
@@ -433,10 +433,7 @@ dns_connect_thread(thread_ref_t thread)
 	/* handle connection status & register check worker thread */
 	if (socket_connection_state(fd, status, thread, dns_check_thread, co->connection_to)) {
 		close(fd);
-		dns_log_message(thread, LOG_INFO,
-				"UDP socket bind failed. Rescheduling.");
-		thread_add_timer(thread->master, dns_connect_thread, checker,
-				 checker->delay_loop);
+		dns_final(thread, true, "UDP socket bind failed. Rescheduling.");
 	}
 }
 
--- a/keepalived/check/check_tcp.c
+++ b/keepalived/check/check_tcp.c
@@ -226,6 +226,7 @@ tcp_connect_thread(thread_ref_t thread)
 			tcp_epilog(thread, false);
 		} else {
 			log_message(LOG_INFO, "TCP socket bind failed. Rescheduling.");
+			tcp_epilog(thread, false);
 			thread_add_timer(thread->master, tcp_connect_thread, checker,
 					checker->delay_loop);
 		}


wangxp006:
Hi guys, why is it still in the pending state after such a long time? I checked the  CI result  and it's passed.

pqarmitage:
@wangxp006 I think you will find that keepalived is one of the most responsive projects in terms of dealing with issue reports and pull requests. With many projects, issues and pull requests can languish for months without being dealt with, whereas with keepalived it generally varies from a few hours to a few days.

With all pull requests we always check them thoroughly before merging them to attempt to ensure that no unwitting side effects have been missed, and that the change hasn't missed some use cases that may affect other users of keepalived. I think it is fair to say that in most cases, apart from the most trivial of changes, we generally have to go back to the submitter because of some issue with the pull request. This all takes time and effort on our part, and so it is not reasonable to expect all pull requests to be merged in a matter of hours; we have other things to do as well :).

As an example with your patch for the TCP checker, after a tcp bind failure, your 'fix' causes two copies of the tcp_checker to be scheduled to be run, one by the existing `thread_add_timer()` call after `delay_loop` seconds, and one by the new `tcp_epilog()` call you have added after `delay_before_retry` seconds. Each of these in turn wil cause a further two instances of the checker to be scheduled. This means that in effect there is a periodic doubling of the number of instances of the checker running, causing an exponential growth in the number of checkers. This will, in time, overwhelm the system, certainly causing keepalived problems, and potentially causing the whole system to run out of resources. Don't forget, keepalived runs with root privileges, and potentially with escalated real-time scheduling priorities, so getting something wrong like this in keepalived can have devastating consequences.

My initial assessment is, yes, it was correct to add the call to `tcp_epilog()`, but you should have removed the call to `thread_add_timer()` on the line below. Do you want to submit an updated patch, or shall I just fix it? I haven't had a detailed look at the DNS_check update.

pqarmitage:
Having thought about this some more, I think the way the checkers currently behave is the way that it is intended. The TCP_CHECK and DNS_CHECK are intended to check that the remote system is down/not working. A local failure, which is what the bind failure is, does not indicate that the remote system is not working.

For example, suppose that on the system running keepalived its link to connect to the DNS server is down, but the real server still can connect to the DNS server. In this circumstance it would be wrong for the DNS check to fail, since the real server still can perform as needed.

I think what you want is an ADDRESS_CHECK, which will check if a local address is configured, possibly on a specified interface, and will fail if the address is missing. You possibly also want an INTERFACE_CHECK, to check if an interface is up. That will achieve what you are wanting to do, and covers the scenario you describe in issue #1679 of the address being deleted or the interface is down. It doesn't alter the current behaviour of TCP_CHECK and DNS_CHECK which currently check that the service is running on the remote system, and are not checking for a local failure.

It would be possible to add a configure option `fail_on_local_failure` for the checkers (it doesn't just apply to TCP_CHECK and DNS_CHECK), but I think the ADDRESS_CHECK and INTERFACE_CHECK give more flexibility and have other uses.

We would gladly consider patches to add an ADDRESS_CHECK and an INTERFACE_CHECK, but note that these should not be done by polling, but rather responding to netlink messages, as the VRRP code currently does.

wangxp006:
HI @pqarmitage , First of all , I must apologize for my unclear expression.My poor English caused this misunderstanding.
 The PR I wanted to say was in the meaning of pending, referring to the PR showing CI check is pending. But I checked that CI check is passed. I guessed that it was the reason that the CI status was not synchronized to the PR. My intention is not to say that PR has never been merged in.
I very much agree that keepalived is one of the most efficient projects on gitbhub,I do see very few projects on github where someone can reply to isuue and PR so efficiently. I pay tribute to your hard work.。
And , I apologize for my previous wrong expression again.Next time I leave a message, I will carefully check my comments to prevent misunderstandings again.
 Have a  nice day

wangxp006:
I will carefully read your comments for these two patches

pqarmitage:
@wangxp006 Many apologies if I misinterpreted your intended message. I must say that your English seems very good to me!

I have thought about this further and I think the simple way forward with your patches is to add a checker option `fail on local failure`, and it is only if that is set that your changes are used.

So the TCP checker code would be something like
```
                  if (checker->co->fail_on_local_failure)
                        tcp_epilog(thread, false);
                  else
			thread_add_timer(thread->master, tcp_connect_thread, checker,
					checker->delay_loop);
```
and
```
          if (!checker->co->fail_on_local_failure) {
		dns_log_message(thread, LOG_INFO,
				"UDP socket bind failed. Rescheduling.");
		thread_add_timer(thread->master, dns_connect_thread, checker,
				 checker->delay_loop);
          } else
		dns_final(thread, true, "UDP socket bind failed. Rescheduling.");
```
and then the `fail_on_local_failure` keyword needs to be added via the `install_checker_common_keywords()` in check_api.c.

I think the UDP_CHECK, PING_CHECK, SMTP_CHECK, HTTP_CHECK (and SSL_CHECK) would also benefit from this.

Please also don't forget to update doc/man/man5/keepalived.conf.

wangxp006:
Hi @pqarmitage , I still  have a little question about this issue. After readed the implementation of UDP checker, I found that UDP will call udp_epilog(thread, false) after bind failed ; this is inconsistent with the implementation of TCP, and also inconsistent with the scheme we discussed earlier.

```
if (udp_icmp_check_state(fd, status, thread, udp_check_thread, co->connection_to)) {
		close(fd);
		udp_epilog(thread, false);
	}

```
Therefore, my question is, should we change all checkers to the scheme using fail_on_local_failure discussed before, or refer to the implementation of UDP.

PR #1682, opened 2020-07-03
Document checker_log_all_failures for MISC_CHECK

wodry:
At least for MISC_CHECK, there is a special handling of checker_log_all_failures: It force-enables no_checker_emails.
At least for DNS_CHECK, there is not this force-enabling of no_checker_emails.

See check_misc.c:

    if (global_data->checker_log_all_failures || checker->log_all_failures)
    message_only = true;

and then:

    if (!message_only) {
    rs_was_alive = checker->rs->alive;
    update_svr_checker_state(script_success ? UP : DOWN, checker);
     
    if (checker->rs->smtp_alert &&
        (rs_was_alive != checker->rs->alive || !global_data->no_checker_emails)) {
    snprintf(message, sizeof(message), "=> MISC CHECK %s on service <=", script_exit_type);
    smtp_alert(SMTP_MSG_RS, checker, NULL, message);
    }
    }

### DIFF
--- a/doc/man/man5/keepalived.conf.5
+++ b/doc/man/man5/keepalived.conf.5
@@ -221,9 +221,8 @@ and
     # Sets default state for checker smtp_alerts
     \fBsmtp_alert_checker \fR<BOOL>
 
-    # Logs every failed real server check in syslog
-    # (nevertheless, SMTP alert is only sent when all retry checks failed
-    # and real server transitions to DOWN state)
+    # Logs every failed real server check in syslog.
+    # For MISC_CHECK, this force-enables no_checker_emails
     \fBchecker_log_all_failures \fR<BOOL>
 
     # Don't send smtp alerts for fault conditions


pqarmitage:
@wodry I think the code as it stands is correct (apart from the messages getting the number of retries wrong). `log_all_failures` is only intended to add logging of failures (and not to send emails for such failures) that do not cause the checker to transition to the fault state (in other words when the retry limit has not yet been reached). The purpose of `log_all_failures` is essentially to assist in debugging the use of checkers.

`message_only` is only set true if the retry limit has not yet been reached, and so it does not suppress the sending of the failure email when the checker transitions to fault state, i.e. once the retry limit has been reached. I have tested this and for me it works as I have described.

Based on the above, I think the proposed amendment to doc/man/man5/keepalived.conf is not correct since `log_all_failures` does not suppress checker emails.

wodry:
OK I close.
Could you tell me what no_checker_emails is for? I do not understand it. I tested with misc_check and http_check, and could not see andy difference if it was enabled or not.    (and always checker_log_all_failures set to true, but as I understand this has nothing to do with no_checker_emails  ?  Confusing

pqarmitage:
`no_checker_emails` means that the normal emails that are sent when a checker changes state will not be sent. The emails for real servers being added or removed will still be sent.

The reason for this is that some users are not interested in state changes of individual checkers, but are interested in a real server being added or removed; it simply serves to reduce noise when not wanted.

PR #1707, opened 2020-07-31
support validate configuration before reload

wangxp006:
According to the solution discussed in the previous PR(#1669), modify the function of rolling back the reload if the configuration item is incorrect

The specific modification contents are as follows:
1. Added configuration items test_config_before_reload and  reload_check_log_path.
test_config_before_reload on                //enable the function of checking the configuration file before reload
reload_check_log_path "/var/log/messages"   //Specifies the location for storing logs during configuration check. the default value is/dev/null

2. When test_config_before_reload is enabled, fork a new process.
The new process inherits the start parameter of the main process and adds the--config-test parameter to check whether the new configuration item is OK

3.After the new process is executed, check whether the validate is OK according to the exit code. if the new configuration is OK, continue to reload; otherwise, do not reload.

### DIFF
--- a/keepalived/core/global_parser.c
+++ b/keepalived/core/global_parser.c
@@ -416,6 +416,37 @@ checker_log_all_failures_handler(const vector_t *strvec)
 
 	global_data->checker_log_all_failures = res;
 }
+
+static void
+test_config_before_reload_handler(const vector_t *strvec)
+{
+	int res = true;
+
+	if (vector_size(strvec) >= 2) {
+		res = check_true_false(strvec_slot(strvec,1));
+		if (res < 0) {
+			report_config_error(CONFIG_GENERAL_ERROR, "Invalid value for test_config_before_reload specified");
+			return;
+		}
+	}
+
+	global_data->test_config_before_reload = res;
+	if (!global_data->reload_check_log_path)
+		global_data->reload_check_log_path = STRDUP("/dev/null");
+}
+
+
+static void
+reload_check_log_path_handler(const vector_t *strvec)
+{
+	if (vector_size(strvec) < 2) {
+		report_config_error(CONFIG_GENERAL_ERROR, "default reload_check_log_path  /dev/null");
+		return;
+	}
+	FREE_CONST_PTR(global_data->reload_check_log_path);
+	global_data->reload_check_log_path = set_value(strvec);
+}
+
 #endif
 #ifdef _WITH_VRRP_
 static void
@@ -1993,6 +2024,8 @@ init_global_keywords(bool global_active)
 #ifdef _WITH_LVS_
 	install_keyword("smtp_alert_checker", &smtp_alert_checker_handler);
 	install_keyword("checker_log_all_failures", &checker_log_all_failures_handler);
+	install_keyword("test_config_before_reload",   &test_config_before_reload_handler);
+	install_keyword("reload_check_log_path",   &reload_check_log_path_handler);
 #endif
 #ifdef _WITH_VRRP_
 	install_keyword("dynamic_interfaces", &dynamic_interfaces_handler);
--- a/keepalived/core/main.c
+++ b/keepalived/core/main.c
@@ -115,6 +115,7 @@
 #include "reload_monitor.h"
 #endif
 #include "warnings.h"
+#include <sys/prctl.h>
 
 /* musl libc doesn't define the following */
 #ifndef	W_EXITCODE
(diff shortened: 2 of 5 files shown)


wangxp006:
Test Case 1.    a)  Without  configuration  test_config_before_reload 
                b)  A required parameter is missing
				
				Result: Keepalived status is dead

`virtual_server 22.22.29.244 80 {
    delay_loop 6
    lb_algo rr
    lb_kind DR
    protocol TCP
    #alpha
    hysteresis 0
    quorum 1
    quorum_up "/usr/sbin/ip addr add 22.22.29.244 dev lo"

    real_server 22.22.99.47 80 {
        weight 1
        HTTP_GET {
            ## all urls should work
            url {
                path /
            }
            **connect_port**
            connect_timeout 5
            nb_get_retry 3
            delay_before_retry 3
        }
    }
}
`
[root@openEuler ~]# systemctl reload keepalived
[root@openEuler ~]# systemctl status keepalived        
● keepalived.service - LVS and VRRP High Availability Monitor
     Loaded: loaded (/usr/lib/systemd/system/keepalived.service; disabled; vendor preset: disabled)
     **Active: inactive (dead)**

Aug 12 20:41:46 openEuler Keepalived_healthcheckers[121602]: UDP connection to [22.22.99.47]:udp:53 success.
Aug 12 20:41:49 openEuler Keepalived_healthcheckers[121602]: Remote Web server [22.22.80.121]:tcp:80 succeed on service.
Aug 12 20:47:47 openEuler systemd[1]: Reloading LVS and VRRP High Availability Monitor.
Aug 12 20:47:47 openEuler Keepalived[121601]: Reloading ...
Aug 12 20:47:47 openEuler systemd[1]: Reloaded LVS and VRRP High Availability Monitor.
Aug 12 20:47:47 openEuler Keepalived[121601]: Opening file '/etc/keepalived/keepalived.conf'.
Aug 12 20:47:47 openEuler Keepalived[121601]: pid 121602 exited with permanent error CONFIG. Terminating
Aug 12 20:47:47 openEuler Keepalived[121601]: CPU usage (self/children) user: 0.000000/0.230876 system: 0.003085/0.270627
Aug 12 20:47:47 openEuler Keepalived[121601]: Stopped Keepalived v2.1.5 (unknown)
Aug 12 20:47:47 openEuler systemd[1]: keepalived.service: Succeeded.





Test Case 2.    a)  With configuration  test_config_before_reload  && reload_check_log_path "/var/log/messages"
                b)  A required parameter is missing
				
				Result:
				a)  Keepalived status is running, and the previous configuration items are kept
				b)  Error message is store in configed file using reload_check_log_path.

`
 global_defs {
        test_config_before_reload on
                reload_check_log_path "/var/log/messages"
 }


virtual_server 22.22.29.244 80 {
    delay_loop 6
    lb_algo rr
    lb_kind DR
    protocol TCP
    #alpha
    hysteresis 0
    quorum 1
    quorum_up "/usr/sbin/ip addr add 22.22.29.244 dev lo"

    real_server 22.22.99.47 80 {
        weight 1
        HTTP_GET {
            ## all urls should work
            url {
                path /
            }
            **connect_port**
            connect_timeout 5
            nb_get_retry 3
            delay_before_retry 3
        }
    }
}
`
[root@openEuler ~]# systemctl reload keepalived
[root@openEuler ~]# systemctl status keepalived
● keepalived.service - LVS and VRRP High Availability Monitor
     Loaded: loaded (/usr/lib/systemd/system/keepalived.service; disabled; vendor preset: disabled)
     **Active: active (running) (thawing) since Wed 2020-08-12 20:53:30 CST; 2min 23s ago**
    Process: 121845 ExecStart=/usr/sbin/keepalived $KEEPALIVED_OPTIONS (code=exited, status=0/SUCCESS)
    Process: 121892 ExecReload=/bin/kill -HUP $MAINPID (code=exited, status=0/SUCCESS)
   Main PID: 121846 (keepalived)
      Tasks: 2 (limit: 1039)
     Memory: 4.3M
     CGroup: /system.slice/keepalived.service
             ├─121846 /usr/sbin/keepalived -D
             └─121847 /usr/sbin/keepalived -D
			 
			 
[root@openEuler ~]# more /var/log/messages
Aug 12 20:55:41 openEuler systemd[1]: Stopping System Logging Service...
Aug 12 20:55:42 openEuler rsyslogd[121773]: [origin software="rsyslogd" swVersion="8.2002.0" x-pid="121773" x-info="https://www.rsyslog.com"] exiting on signal 15.
Aug 12 20:55:42 openEuler systemd[1]: rsyslog.service: Succeeded.
Aug 12 20:55:42 openEuler systemd[1]: Stopped System Logging Service.
Aug 12 20:55:42 openEuler audit[1]: SERVICE_STOP pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='unit=rsyslog comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? t
erminal=? res=success'
Aug 12 20:55:42 openEuler systemd[1]: Starting System Logging Service...
Aug 12 20:55:42 openEuler rsyslogd[121884]: [origin software="rsyslogd" swVersion="8.2002.0" x-pid="121884" x-info="https://www.rsyslog.com"] start
Aug 12 20:55:42 openEuler rsyslogd[121884]: imjournal: journal files changed, reloading...  [v8.2002.0 try https://www.rsyslog.com/e/0 ]
Aug 12 20:55:42 openEuler systemd[1]: Started System Logging Service.
Aug 12 20:55:42 openEuler audit[1]: SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='unit=rsyslog comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? 
terminal=? res=success'
**(/etc/keepalived/keepalived.conf: Line 27) *** Configuration line starting `connect_port` is missing a parameter after keyword `connect_port` at word position 2**
Aug 12 20:55:47 openEuler systemd[1]: Reloading LVS and VRRP High Availability Monitor.
Aug 12 20:55:47 openEuler systemd[1]: Reloaded LVS and VRRP High Availability Monitor.

wangxp006:
HI @pqarmitage , if there is any problem, please let me know. If you need me to provide a test report for more scenarios, please also tell me which scenarios to test.

pqarmitage:
@wangxp006 I was looking at this this morning and while a number of issues with the PR have occurred to me, I am not confident that I have identified them all yet.

This problems I have identified so far are:
1. For some reason the `test_config_before_reload` and `reload_check_log_path` keywords are only enabled if the code is built with healthchecker/IPVS support. Also the handler functions are defined regardless of whether healthchecker support is enabled or not.
1. The action reload_validate_fin() takes on `!validate_is_fail` is not the same action that is taken normally for a config reload. In fact, the way you have implemented it it would be simpler to run a script to validate the config and if that is successful, RESTART keepalived (of course, the same thing could be done for reloading the config, and might be simpler overall).
1. The configuration in the parent process is never updated
1. Forking another process which calls back to the main process isn't going to work. When I wrote in #1669 `forks another keepalived instance running it as keepalived --config-test=/dev/null -f CONFIG_FILE ...` I envisaged that keepalived would be exec'd so it starts up clean and doesn't inherit all the memory allocations, file descriptors etc from it's parent process (to me this was inferred by the command line options not being the same).
1. Calling `thread_add_child()` with TIMER_NEVER is not a good idea here, since we need the child process to terminate.
1. Calling `prctl(PR_SET_NAME, argv[0])` when the length of argv[0] is longer than 15 characters won't work as expected.
1. I am not clear why you use `malloc()` for argv, but `MALLOC()` for argv[sav_argc].

I think this is probably close enough now that it will be easier for me to modify your patch so that it is fully consistent with the keepalived design, unless you would particularly would like to update it yourself.

wangxp006:
Hi @pqarmitage , thank you very much for your precious time to review my code.After carefully read your review comments, I gain a lot. I think it would be very helpful if you could help me with modifing the patch. I can also learn better keepalived desgin through your modifications.

pqarmitage:
I have now merged commit 9705ba1 which is based on your patch with some updates.

Many thanks for your work on this, and apologies for the delay in merging this.

PR #1717, opened 2020-08-10
Use --with-default-config-file for installing default config file

pqarmitage:


roommen:
Tested this change on my setup and works great!
Thank you Quentin for making this quick fix 💯

Issue #1748, opened 2020-09-29
build warning using --disable-vrrp


wangxp006:
**Describe the bug**
build warning using --disable-vrrp

**To Reproduce**
1. download keepalived in keepalived-master
2. ./build_setup
3 ./configure --disable-vrrp
4. make -j20

**Expected behavior**
Build success  without warning 

**Keepalived version**
keepalived master in github

[root@randywang keepalived]# autoconf --version
autoconf (GNU Autoconf) 2.69

[root@randywang keepalived]# gcc --version
gcc (GCC) 7.3.0

** Build Log **
[root@randywang keepalived]# make -j20
Making all in lib
make[1]: Entering directory '/root/build/keepalived/lib'
make  all-am
make[2]: Entering directory '/root/build/keepalived/lib'
  CC       vector.o
  CC       html.o
  CC       parser.o
  CC       signals.o
  CC       logger.o
  CC       list_head.o
  CC       rbtree.o
  CC       process.o
  CC       json_writer.o
  CC       memory.o
  CC       utils.o
  CC       notify.o
  CC       timer.o
  CC       scheduler.o
  AR       liblib.a
make[2]: Leaving directory '/root/build/keepalived/lib'
make[1]: Leaving directory '/root/build/keepalived/lib'
Making all in keepalived
make[1]: Entering directory '/root/build/keepalived/keepalived'
Making all in core
make[2]: Entering directory '/root/build/keepalived/keepalived/core'
  CC       main.o
  CC       daemon.o
  CC       pidfile.o
  CC       layer4.o
  CC       smtp.o
  CC       global_data.o
  CC       global_parser.o
  CC       keepalived_netlink.o
  CC       namespaces.o
  CC       reload_monitor.o
main.c: In function ‘sigend’:
 main.c:1089:6: warning: assuming signed overflow does not occur when simplifying conditional to constant [-Wstrict-overflow]
   if (wait_count) { 
      ^
  AR       libcore.a
make[2]: Leaving directory '/root/build/keepalived/keepalived/core'
Making all in check
make[2]: Entering directory '/root/build/keepalived/keepalived/check'
  CC       check_daemon.o
  CC       check_data.o
  CC       check_parser.o
  CC       check_api.o
  CC       check_tcp.o
  CC       check_http.o
  CC       check_ssl.o
  CC       check_smtp.o
  CC       check_misc.o
  CC       check_dns.o
  CC       check_print.o
  CC       ipwrapper.o
[...]

PR #1749, opened 2020-09-29
fix build warning using --disable-vrrp

wangxp006:
fix build warning using --disable-vrrp
Detail is in issue #1748


pqarmitage:
Please see comments in issue #1748 for reasons why this pull request is not being applied.

@wangxp006 Many thanks for your work with helping to improve keepalived.

Issue #1750, opened 2020-09-29
build Error using --enable-debug --disable-vrrp


wangxp006:
Describe the bug
build warning using --disable-vrrp

To Reproduce

download keepalived in keepalived-master
./build_setup
3 ./configure --enable-debug --disable-vrrp 
make -j20
Expected behavior
Build success  

Keepalived version
keepalived master in github

[root@randywang keepalived]# autoconf --version
autoconf (GNU Autoconf) 2.69

[root@randywang keepalived]# gcc --version
gcc (GCC) 7.3.0

 **Build Log** 
[root@localhost keepalived]# make -j20
Making all in lib
make[1]: Entering directory '/root/opensource/keepalived/keepalived/lib'
make  all-am
make[2]: Entering directory '/root/opensource/keepalived/keepalived/lib'
  CC       assert.o
  CC       memory.o
  CC       utils.o
  CC       notify.o
  CC       timer.o
  CC       vector.o
  CC       html.o
  CC       parser.o
  CC       signals.o
  CC       logger.o
  CC       list_head.o
  CC       rbtree.o
  CC       process.o
  CC       json_writer.o
  CC       scheduler.o
  AR       liblib.a
make[2]: Leaving directory '/root/opensource/keepalived/keepalived/lib'
make[1]: Leaving directory '/root/opensource/keepalived/keepalived/lib'
Making all in keepalived
make[1]: Entering directory '/root/opensource/keepalived/keepalived/keepalived'
Making all in core
make[2]: Entering directory '/root/opensource/keepalived/keepalived/keepalived/core'
  CC       main.o
  CC       daemon.o
  CC       pidfile.o
  CC       layer4.o
  CC       smtp.o
  CC       global_data.o
  CC       global_parser.o
  CC       keepalived_netlink.o
  CC       namespaces.o
  CC       reload_monitor.o
In file included from ../../keepalived/include/vrrp_track.h:35:0,
                 from ../../keepalived/include/vrrp_scheduler.h:34,
                 from main.c:103:
../../keepalived/include/vrrp.h:139:3: error: conflicting types for ‘vrrp_sgroup_t’
 } vrrp_sgroup_t;
   ^~~~~~~~~~~~~
In file included from main.c:94:0:
../../keepalived/include/smtp.h:125:14: note: previous declaration of ‘vrrp_sgroup_t’ was here
 typedef void vrrp_sgroup_t;
              ^~~~~~~~~~~~~
[...]

PR #1751, opened 2020-09-29
fix build error with --enable-debug --disable-vrrp

wangxp006:
fix build error with --enable-debug --disable-vrrp
Details is here #1750


pqarmitage:
See issue #1750 for an explanation of why this commit wasn't merged, although part of the change has been implemented.

@wangxp006 Many thanks for your help with this, you have again contributed to improving keepalived.

PR #1755, opened 2020-10-04
parser: report_config_error when failing opening a file

wdauchy:
while using `reload_check_config`, it is useful to make sure to fail
when the config file or its includes are not possible to be open.

it will avoid a situation where a reload continues to happen while an
important config included cannot be open.

### DIFF
--- a/doc/man/man5/keepalived.conf.5
+++ b/doc/man/man5/keepalived.conf.5
@@ -700,6 +700,9 @@ and
     # if the configuration is valid.
     \fBreload_check_config \fR[LOG_FILE]
 
+    # Treat each all globals config issues as config errors
+    \fBglob_strict \fR[LOG_FILE]
+
     # reload_time_file allows a reload of keepalived to be scheduled in the future. This is
     # particularly useful if there is a master keepalived and one or more backup keepalived
     # instances and the new configuration is incompatible with the previous configuration,
--- a/keepalived/core/global_parser.c
+++ b/keepalived/core/global_parser.c
@@ -1986,6 +1986,12 @@ random_seed_handler(const vector_t *strvec)
 
 #ifndef _ONE_PROCESS_DEBUG_
 static void
+glob_strict_handler(__attribute__((unused)) const vector_t *strvec)
+{
+	glob_strict_set();
+}
+
+static void
 reload_check_config_handler(const vector_t *strvec)
 {
 	if (vector_size(strvec) >= 2) {
@@ -2223,5 +2229,6 @@ init_global_keywords(bool global_active)
 	install_keyword("reload_time_file", &reload_time_file_handler);
 	install_keyword("reload_repeat", &reload_repeat_handler);
 	install_keyword("reload_file", &reload_file_handler);
+	install_keyword("glob_strict", &glob_strict_handler);
 #endif
 }
--- a/lib/parser.c
+++ b/lib/parser.c
@@ -142,6 +142,7 @@ bool do_parser_debug;
 #ifdef _DUMP_KEYWORDS_
 bool do_dump_keywords;
 #endif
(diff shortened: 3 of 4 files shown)


pqarmitage:
I think there are one or two issues with this commit, and I am not clear what the correct solution is.

1. Please see commit 17686ee, which takes the view that a failure of a glob to match a file should not be treated as a fatal error. Specifically it gives the example of `include /etc/keepalived/puppet.d/*.conf` which may legitimately return no files.

2. I don't think adding the `report_config_error(...)` line would be the correct solution, aside from 1) above. It would be better to change the two proceeding `log_message(...)` lines to be `report_config_error(...)`.

3. Should some of the `log_message()`s in `open_conf_file()` be `report_config_error()`s? And should the `glob()` call returning a directory be an error?

It is important that whatever change is made doesn't break existing valid configurations, as perceived by the user of that configuration. So I think if the checking is to be made stricter, there would have to be new configuration keywords to enable the stricter checking. For example:
If `glob_strict` is configured use `report_config_error()` instead of `log_message()` if:
1. glob() returns GLOB_NOMATCH
1. glob() returns any other error
1. open_conf_file() attempts to open a directory
1. open_conf_file() cannot open a file
1. open_conf_file() is not a regular non-executable file
1. if chdir() fails

if `glob_strict_no_match` is configured, use `report_config_error()` instead of `log_message()` for the same as `glob_strict` except item 1) which should return an error unless the string contains a wildcard ('?', '*' or '[') character.

There is a further complication with brace matching. Suppose `/tmp/a.txt` exists and `/tmp/b.txt` does not exist. Using glob("/tmp/{a,b}.txt", ...) returns 1 file, `/tmp/a.txt` without any error indication, whereas bash, for example, returns an error that `/tmp/b.txt` does not exist. I prefer the latter behaviour, but I cannot see how that can be done using `glob()`.

@wdauchy Your thoughts on this would be appreciated so that we can work out a way forward that suits everyone.

wdauchy:
@pqarmitage small question for the implementation: how may I include "global_data.h" in "lib/parser.c" in order to test `global_data->glob_strict` value?
Thanks

pqarmitage:
@wdauchy I think you are absolutely right that (most of) the logged errors you have highlighted should be handled as config errors, and it is right that everyone should be able to benefit (even to the extent that I would prefer it to be the default, except for the need not to break existing configs).

I do agree with @apoikos that it is legitimate to have configuration directories with optional configuration, and those directories may be empty (see the output of `find /etc -name "*.d" -type d` for many examples). On the other hand, there are people such as you (and me when I originally coded it) who think the response to an `include` directive in the config file should return at least one file. In order to allow for those who want to be able to have directories where there may be config files, but they can be empty, then I think we should have `glob_strict` for those who would like your approach, and `glob_strict_no_match` for those who prefer the approach of @apoikos. If you do not want to add the `glob_strict_no_match` functionality then I am happy to add it after you have produced your patch.

I would still like a way for `include /tmp/{a,b,c}.conf` to be treated as an error if, say, /tmp/b.conf does not exist. I cannot see how to do that with `glob()` whereas bash does produce an error. Does anyone have an idea about how this can be achieved without keeplived having to parse the file name string (which we are not going to do)?

Regarding your question about `global_data->glob_strict`, I think rather than having the value in global_data, the parser for `glob_strict` in global_parser.c needs to call a function in lib/parser.c that sets a local flag. There should also be a function to retrieve the value so that it can be written when dumping keepalived.data (SIGUSR1).

wdauchy:
hello @pqarmitage 
thank you for your complete answer and support. On my side, sorry for the delay, this covid situation has a bad impact on my projects.
You will find my up to date patch. If the later is satisfying I will consider `glob_strict_no_match`.

PR #1767, opened 2020-10-28
Fix a file descriptor leak during reloads

pqarmitage:


pqarmitage:
Not waiting for Travis-CI any longer - it has got a backlog of 9,299 Linux builds.

Issue #1798, opened 2020-12-02
Cannot change vmac name on conf reload


louis-oui:
**Describe the bug**
Changing vmac name does not work on config reload. Works on config restart.

**To Reproduce**
apt install systemd-coredump -y
systemctl stop keepalived
rm /var/lib/systemd/coredump/*
ifconfig ens4 192.168.15.41/28 up
cat>/etc/keepalived/keepalived.conf <<\EOF
global_defs
{
router_id router
enable_script_security
script_user root
dynamic_interfaces
}

vrrp_sync_group group15 {
group {
vrrp
}
}

vrrp_track_file fp_tracker {
file /var/run/keepalived/fp-tracker
weight 0
}

vrrp_instance vrrp {
version 2
state BACKUP
interface ens4

use_vmac vrrp

track_file {
}

garp_master_delay 5

virtual_router_id 15

priority 200
advert_int 1.0

virtual_ipaddress {
192.168.15.38/28
}

preempt_delay 0
}
EOF
systemctl restart keepalived
ip link show vrrp
ip link show vrrpnew
sleep 10
sed -e 's|use_vmac vrrp|use_vmac vrrpnew|g' -i /etc/keepalived/keepalived.conf
kill -s SIGHUP $(pgrep keepalived | head -n1) #signal for conf reload
ip link show vrrp
ip link show vrrpnew
sleep 10
systemctl restart keepalived
ip link show vrrp
ip link show vrrpnew
coredumpctl | grep -v missing

**Observed behavior**
vmac name is not changed after reload
Dec 02 17:51:48 ha1 Keepalived_vrrp[8428]: (vrrp) Found matching interface vrrp
Dec 02 17:51:48 ha1 Keepalived_vrrp[8428]: (vrrp) vmac name mismatch vrrpnew <=> vrrp. changing to vrrp.

**Expected behavior**
vmac name changes after reload

**Keepalived version**
Keepalived v2.1.5 (07/13,2020)

Copyright(C) 2001-2020 Alexandre Cassen, <acassen@gmail.com>

Built with kernel headers for Linux 4.15.18
Running on Linux 4.15.0-123-generic #126-Ubuntu SMP Wed Oct 21 09:40:11 UTC 2020

[...]

PR #1801, opened 2020-12-04
vrrp: allow vmac rename at reload

louis-oui:
Related to https://github.com/acassen/keepalived/issues/1798

At reload, if we detect an existing vmac interface matches conf settings with
a different name and the name of the existing vmac interface is in the previous,
we delete existing to recreate a new vmac interface in order to rename.

Same behavior is kept at startup. If a vmac matching conf settings is already
found at startup, this vmac interface is used even the name is different.


pqarmitage:
I have merged a very slight modification to this as commit a43991d. Many thanks for providing this patch.

PR #1863, opened 2021-03-01
check kernel version before set IPVS_DEST_ATTR_ADDR_FAMILY

davidwtf:
Signed-off-by: Tengfei Wang <tfwang@alauda.io>

Check kernel version before set IPVS_DEST_ATTR_ADDR_FAMILY. 
On the kernel of lower version will return error if set IPVS_DEST_ATTR_ADDR_FAMILY.

### DIFF
--- a/keepalived/check/libipvs.c
+++ b/keepalived/check/libipvs.c
@@ -25,6 +25,7 @@
 #include <string.h>
 #include <stdint.h>
 #include <stdbool.h>
+#include <linux/version.h>
 
 #ifdef LIBIPVS_USE_NL
 #include <netlink/netlink.h>
@@ -52,6 +53,8 @@
 #include "utils.h"
 #include "namespaces.h"
 #include "global_data.h"
+
+#include "main.h"
 
 typedef struct ipvs_servicedest_s {
 	struct ip_vs_service_user	svc;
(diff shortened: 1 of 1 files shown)


pqarmitage:
@davidwtf Many thanks for the patch. I am not clear however why it is needed, since the existing code is:
```
#if HAVE_DECL_IPVS_DEST_ATTR_ADDR_FAMILY
        NLA_PUT_U16(msg, IPVS_DEST_ATTR_ADDR_FAMILY, dst->af);
#endif
```
and so the line of code in question will only be compiled if the kernel header <linux/ip_vs.h> defines IPVS_DEST_ATTR_ADDR_FAMILY, which means that the kernel version is either >= 3.18 or the feature has been backported into the kernel. Furthermore, there are 9 other places where IPVS_DEST_ATTR_ADDR_FAMILY is used, and they would need similar treatment to be consistent.

It is worth noting that keepalived detects which kernel features are available at compile time, not at runtime. It is not supported to build keepalived and then run it on a different system with an older kernel, and if that happens keepalived will log a warning when it starts that keepalived was build for a newer kernel.

davidwtf:
Yes, you are right. We have already checked it when compiling.
But when use Docker image, the running kernel version is usually different from the compiling kernel.

pqarmitage:
keepalived does all its checking of what kernel features are available at compile time, and most of that is done when configure runs.  There are a large number of kernel features that keepalived checks for and to do these checks at runtime would make the code somewhat less efficient. Testing against the kernel version isn't reliable since RedHat backported a number of features into the 2.6.32 kernel for RHEL6, and so it the kernel version were used to test for the feature then keepalived wouldn't use the backported features even if they were available.

The file README.kernel_versions in the root of the source tree shows what kernel features are tested for and which mainstream kernel introduced the feature. There were 40 additional kernel features that were tested for until I removed support for RHEL6 after it went EOL.

It isn't just the kernel features that vary, but also some of the libraries, such as libnftnl.

The compile time checks are absolutely necessary since otherwise compilation will fail due to the use of undefined names. To add to runtime checks as well does seem rather excessive.

Since the docker image uses the host kernel, you need to build keepalived on the host where the docker image will run, or at least a host with the same kernel headers and library headers that match the base docker image.

When we build snaps, we build several different versions of keepalived using different kernel headers, and a script in the snap container checks which keepalived version to load based on the kernel version. snap/snapcraft.yaml fill show you what it does, and you may find that technique is what you need to do if you want to run your docker image on different kernels. Alternatively you could build the docker image using the kernel headers for the oldest kernel version you will run the docker image on, but then you will lose use of any newer kernel features that might be available.

pqarmitage:
@davidwtf I am closing this pull request now since there has been no update for 10 days, and we could not merge the patch as submitted for the reasons set out above.

If you, or anyone else, would like to submit a patch that provides runtime checking for _all_ the kernel and library features that are tested for at configure/build time, then depending on the size, performance and complexity impact we would consider that for merging.

PR #1896, opened 2021-04-19
vrrp: fix dont_track_primary on IPv6 when vmac disabled

louis-oui:
Link local IPv6 deletion from the instance top interface causes
down_instance to be called.

Don't call down_instance if dont_track_primary is set on the instance.


pqarmitage:
I don't think this patch is right. When an interface goes down, IPv6 addresses are removed from the interface, whereas IPv4 addresses are not removed from the interface. Since keepalived uses the configured link local IPv6 address for sending adverts, if the address has been deleted no address exists on that interface from which to send adverts. It is therefore right that the VRRP instance goes to fault state if the IPv6 address is deleted.

Issue #1892, opened 2021-04-15
VRRP IPv6 use_vmac disable: dont_track_primary not working


louis-oui:
**Describe the bug**
On VRRP IPv6  with use_vmac disabled, dont_track_primary does not work. Interface down causes FAULT state on instance and then on group.

**To Reproduce**
Machine1 ens4 <-> Machine1 ens4
Machine1 ens5 <-> Machine2 ens5

Machine 1:
```
ip -6 addr add fd00::300:3/64 dev ens4
ip -6 addr add fd00::200:3/64 dev ens5
```

Machine 2:
```
ip -6 addr add fd00::300:4/64 dev ens4
ip -6 addr add fd00::200:4/64 dev ens5
```

Both machines at the same time:
```
ifconfig ens4 up
ifconfig ens5 up
killall keepalived
sleep 2
cat>/etc/keepalived/keepalived.conf <<\EOF
global_defs
{
router_id router
dynamic_interfaces
}

vrrp_sync_group group {
group {
vrrp
vrrp2
}
}

vrrp_instance vrrp {
version 3
state BACKUP
interface ens4

garp_master_delay 5

virtual_router_id 15

priority 200
advert_int 1.0

virtual_ipaddress {
fd00::300:1/64
}

preempt_delay 0
}

vrrp_instance vrrp2 {
version 3
state BACKUP
interface ens5

garp_master_delay 5

virtual_router_id 15

priority 200
advert_int 1.0

virtual_ipaddress {
fd00::200:1/64
}
dont_track_primary
preempt_delay 0
}
EOF

ip -6 add show ens4 | grep 300:4 && sed -e 's|priority 200|priority 150|g' -i /etc/keepalived/keepalived.conf
keepalived -D
sleep 10
ip a show ens4
sleep 10
ip -6 add show ens4 | grep 300:3 && ifconfig ens5 down
sleep 2
ip a show ens4
```

**Expected behavior**
ens5 down on machine1 does not cause FAULT state on vrrp2 machine2 instance.
Virtual address remains on ens4 machine1.

**Keepalived version**
```
Keepalived v2.2.2 (04/10,2021), git commit v2.2.2-27-g475d1bd9

Copyright(C) 2001-2021 Alexandre Cassen, <acassen@gmail.com>

Built with kernel headers for Linux 4.15.18
Running on Linux 4.15.0-123-generic #126-Ubuntu SMP Wed Oct 21 09:40:11 UTC 2020
[...]

PR #1898, opened 2021-04-20
vrrp: fix dont_track_primary on IPv6 when vmac disabled

louis-oui:
Link local IPv6 deletion from the instance top interface causes
down_instance to be called.

There are two cases where link local might be deleted:
 - link is set down. Linux kernel behaviour is to always remove
the link local IPv6 address from the interface (sysctl keep_addr_on_down
keeps IPv6 address except the link local). In that case,
dont_track_primary should allow to not put the instance into FAULT
state. The link local IPv6 address will be set back at link up.
 - IPv6 link local is removed from the interface whereas the link state
remains up. Even dont_track_primary is set, the instance must go to
FAULT state because no advertisement can't be sent.

This patch fixes the first case.

### DIFF
--- a/keepalived/core/keepalived_netlink.c
+++ b/keepalived/core/keepalived_netlink.c
@@ -1127,7 +1127,10 @@ netlink_if_address_filter(__attribute__((unused)) struct sockaddr_nl *snl, struc
 							 vrrp->ifp) &&
 						 vrrp->family == ifa->ifa_family &&
 						 vrrp->saddr.ss_family != AF_UNSPEC &&
-						 (!vrrp->saddr_from_config || is_tracking_saddr)) {
+						 (!vrrp->saddr_from_config || is_tracking_saddr) &&
+						 (vrrp->family == AF_INET ||
+						  !vrrp->dont_track_primary ||
+						  IF_ISUP(ifp))) {
 						down_instance(vrrp);
 						vrrp->saddr.ss_family = AF_UNSPEC;
 					}


keepalived/core/keepalived_netlink.c
```diff
@@ -1127,7 +1127,10 @@ netlink_if_address_filter(__attribute__((unused)) struct sockaddr_nl *snl, struc
 							 vrrp->ifp) &&
 						 vrrp->family == ifa->ifa_family &&
 						 vrrp->saddr.ss_family != AF_UNSPEC &&
-						 (!vrrp->saddr_from_config || is_tracking_saddr)) {
+						 (!vrrp->saddr_from_config || is_tracking_saddr) &&
+						 (vrrp->family != AF_INET6 || (vrrp->family == AF_INET6 &&
```
pqarmitage, line 1131: This code is not optimal, although a compiler will probably optimise it; however being not optimal makes it much harder to work out what it is doing.

The basic issue is (a || (^a && b)) is identical to (a || b)

keepalived/core/keepalived_netlink.c
```diff
@@ -1127,7 +1127,10 @@ netlink_if_address_filter(__attribute__((unused)) struct sockaddr_nl *snl, struc
...
+						  (vrrp->dont_track_primary && IF_ISUP(ifp))) ||
+						  !vrrp->dont_track_primary)) {
 						down_instance(vrrp);
```
pqarmitage, line 1134: This change can be re-written as:
```
                                                         vrrp->ifp) &&
                                                 vrrp->family == ifa->ifa_family &&
                                                 vrrp->saddr.ss_family != AF_UNSPEC &&
                                                 (!vrrp->saddr_from_config || is_tracking_saddr) &&
                                                 (vrrp->family == AF_INET || 
                                                  !vrrp->dont_track_primary ||
                                                  IF_ISUP(ifp))) {
```

So the change to the code is that if the VRRP instance is IPv6 and dont_track_primary is set and the interface is down, then the VRRP instance is not downed.

The trouble with this is that this is triggered by the source address that the VRRP instance is using has been deleted, and we cannot allow keepalived to be sending from an address that no longer exists (and might even have been moved to another system) - my recollection is that for IPv6 the kernel does not allow sending from an IP address that is not configured (but I may be wrong here).

pqarmitage (commented):
I have added some specific comments, but in summary I think the basis of the patch is not correct.

In issue #1892 you mention that `echo 1 >/proc/sys/net/ipv6/conf/<interface>/keep_addr_on_down` stops IPv6 global addresses being deleted on link down, so it seems that this patch is targeted towards being able to continue to send VRRP adverts from a link local address after that link local address has been deleted from the relevant interface. We cannot allow this to happen.

keepalived/core/keepalived_netlink.c
louis-oui: With my patch, I get that on IPv6. No crash. Same behaviour on IPv4 dont_track_primary.

Apr 26 11:02:18 ubuntu Keepalived_vrrp[1802]: (vrrp2): send advert error 22 (Invalid argument)
Apr 26 11:02:19 ubuntu Keepalived_vrrp[1802]: (vrrp2): send advert error 22 (Invalid argument)
Apr 26 11:02:20 ubuntu Keepalived_vrrp[1802]: (vrrp2): send advert error 22 (Invalid argument)
Apr 26 11:02:21 ubuntu Keepalived_vrrp[1802]: (vrrp2): send advert error 22 (Invalid argument)
Apr 26 11:02:22 ubuntu Keepalived_vrrp[1802]: (vrrp2): send advert error 22 (Invalid argument)
Apr 26 11:02:23 ubuntu Keepalived_vrrp[1802]: (vrrp2): send advert error 22 (Invalid argument)
Apr 26 11:02:24 ubuntu Keepalived_vrrp[1802]: (vrrp2): send advert error 22 (Invalid argument)
Apr 26 11:02:25 ubuntu Keepalived_vrrp[1802]: (vrrp2): send advert error 22 (Invalid argument)
Apr 26 11:02:26 ubuntu Keepalived_vrrp[1802]: (vrrp2): send advert error 22 (Invalid argument)

pqarmitage: @louis-oui I would expect the above error with your patch, since keepalived is attempting to send an advert on eth5 with a source address that is not configured. From testing it appears that EINVAL is returned both when the source address is not configured and also when the interface is administratively down.

louis-oui: @pqarmitage I confirm 'send advert error 22 (Invalid argument)' is coming from this Ipv6 patch

pqarmitage:
As discussed, ignoring the removal of addresses from an interface cannot work.

PR #1903, opened 2021-05-11
fix bug: free null pointer.

weiyanhua100:


pqarmitage:
@weiyanhua100 Many thanks for this patch, and you have certainly identified a bug, i.e. calling `recv()` when recv_buf is NULL.

Interestingly, it is permissible to call `free(NULL)` - the man page for free(3) states:
`If ptr is NULL, no operation is performed.`
so with most builds of keepalived the change:
```
- 	FREE(recv_buf);

+      if (NULL != recv_buf)
+           FREE(recv_buf);
	return;
 }
```
is not necessary, although I quite agree it is good practice to avoid calling `free()` with a NULL pointer.

However, if keepalived is built using the configure option `--enable-mem-check`, which is a debugging feature for checking the usage of malloc()/free() is correct (e.g. no double frees, no "lost" mallocs), then it does check that the ptr value of `FREE()` is not NULL, and it logs an error to the memcheck log if the ptr is NULL.

Personally I feel that `if (NULL != recv_buf)` is semantically the wrong way to make the check. In words, the check we want to make is `is recv_buf not NULL?`, and that is better represented by `if (recv_buf != NULL)`. I feel that `if (NULL != recv_buf)` looks like checking whether NULL has a particular value. Further, I prefer `if (recv_buf)` to `if (recv_buf != NULL)`, since it is more succinct.

With regards to the patch
```
+	if (NULL == recv_buf)
+		return connect_success;

	ret = recv(fd, recv_buf, *len, 0);
```
I don't think this is quite correct. Looking at the history of the code, the bug was introduced in commit 3e72832. Prior to that commit the code always called `recv()`, since, according to the comment that follows `recv()` `Ret less than 0 means the port is unreachable`.

I think in order to restore the behaviour prior to commit 3e72832 the patch should be something like:
```
@@ -423,6 +423,7 @@ enum connect_result
 udp_socket_state(int fd, thread_ref_t thread, uint8_t *recv_buf, size_t *len)
 {
        int ret;
+       char    recv_buf_local;
 
        /* Handle Read timeout, we consider it success unless require_reply is set */
        if (thread->type == THREAD_READ_TIMEOUT)
@@ -431,7 +432,11 @@ udp_socket_state(int fd, thread_ref_t thread, uint8_t *recv_buf, size_t *len)
        if (thread->type == THREAD_READ_ERROR)
                return udp_socket_error(fd);
 
-       ret = recv(fd, recv_buf, *len, 0);
+       if (recv_buf) {
+               ret = recv(fd, recv_buf, *len, 0);
+               *len = ret;
+       } else
+               ret = recv(fd, &recv_buf_local, sizeof(recv_buf_local), 0);
 
        /* Ret less than 0 means the port is unreachable.
         * Otherwise, we consider it success.
@@ -440,7 +445,6 @@ udp_socket_state(int fd, thread_ref_t thread, uint8_t *recv_buf, size_t *len)
        if (ret < 0)
                return connect_error;
 
-       *len = ret;
        return connect_success;
 }
 
```

I would be grateful for your comments regarding the above, and would be very happy if you submitted an updated patch.

weiyanhua100:
Thank you very much for the review
Has been modified, please review again.

PR #1916, opened 2021-06-09
Fix docs: children vs child

henkdswiss:


pqarmitage:
@henkdswiss Many thanks for this patch, and I quite agree that `children` should be changed to `child` where you have done so, although I think the last change should be `it simply restarts the child process`.

The difficulty I have with this change is that the description of the software design is extremely out of date. For example, there is no use of unix domain sockets for sending hellos between the processes, and there hasn't been since before I became involved with keepalived (in fact I don't know if there ever has been). As a further example there is now (potentially) a third child process, the BFD child.

My concern is that by applying this patch, it may imply that the rest of the section being patched is correct, whereas that is not currently the case. What do you think?

henkdswiss:
Hi

On Wed, 09 Jun 2021 09:05:48 -0700
Quentin Armitage ***@***.***> wrote:

> The difficulty I have with this change is that the description of the
> software design is extremely out of date. For example, there is no
> use of unix domain sockets for sending hellos between the processes,
> and there hasn't been since before I became involved with keepalived
> (in fact I don't know if there ever has been). As a further example
> there is now (potentially) a third child process, the BFD child.
> 
> My concern is that by applying this patch, it may imply that the rest
> of the section being patched is correct, whereas that is not
> currently the case. What do you think?

I agree.

IMHO it would be good to remove everything that is not correct anymore
to prevent people from wasting their time by reading it and to prevent
confusion.
Having sparse docs full of gaps that need filling but are mostly
correct is IMHO better than having complete docs that are largely out
of date and incorrect.

Thanks for your work on FLOSS and the quick response!

henk

-- 
Hendrik Jaeger
Linux System Engineer

DSwiss AG
Badenerstrasse 329
8003 Zurich
Switzerland

+41 44 515 11 11
https://www.dswiss.com

pqarmitage:
@henkdswiss I have applied this patch, and also updated software_design.rst to bring it more up to date.

PR #1921, opened 2021-06-21
optimize rs online and offline log print

kwb0523:
when rs offline,the log description is not appropriate

### DIFF
--- a/keepalived/check/ipwrapper.c
+++ b/keepalived/check/ipwrapper.c
@@ -605,7 +605,7 @@ perform_svr_state(bool alive, checker_t *checker)
 			    , alive ? (rs->inhibit) ? "Enabl" : "Add" :
 				      (rs->inhibit) ? "Disabl" : "Remov"
 			    , FMT_RS(rs, vs)
-		    	    , alive ? "to" : "from"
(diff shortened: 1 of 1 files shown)


pqarmitage (changes requested):
I agree that a log entry:
`Removing service [192.168.0.2]:tcp:80 to VS [10.0.0.1]:tcp:80`
is not right, however I don't think
`Disabling service [192.168.0.2]:tcp:80 from VS [10.0.0.1]:tcp:80`
as your patch would produce if inhibit is set works either.

I think something like
`, rs->inhibit ? "of" : alive ? "to" : "from"`
would be better.

pqarmitage:
@kwb0523 Many thanks for this patch.

kwb0523:
> I agree that a log entry:
> `Removing service [192.168.0.2]:tcp:80 to VS [10.0.0.1]:tcp:80`
> is not right, however I don't think
> `Disabling service [192.168.0.2]:tcp:80 from VS [10.0.0.1]:tcp:80`
> as your patch would produce if inhibit is set works either.
> 
> I think something like
> `, rs->inhibit ? "of" : alive ? "to" : "from"`
> would be better.

Thanks for your review ,I'm sorry for not being considerate,I have modified it according to your idea,Thanks again.

PR #1968, opened 2021-08-16
build: unbreak calculation of relative path

### LINKED ISSUE #1964: busy loop with `make install`
**Describe the bug**
Configuring with:
```
  ./configure \
    --prefix=/usr \
    --sysconfdir=/etc \
    --sbindir=/usr/bin \
    --localstatedir=/var \
    --runstatedir=/run
```
Building (`make`) succeeds, but installing (`make DESTDIR="$pkgdir" install`) results in an infinite loop.

Looks like this happens in `make install-exec-hook` and was introduced with 81b5d68d08b2661f3256e0a673a4627417aaab85.

**To Reproduce**
Configure, build and install.

**Expected behavior**
Should install.

**Keepalived version**
2.2.3

**Distro (please complete the following information):**
 - Name: Arch
 - Version: rolling
 - Architecture: x86_64

eworm-de:
Commit 81b5d68d08b2661f3256e0a673a4627417aaab85 changed genhash to be a
symbolic link to keepalived, but the code to calculate the relative path
is prone to errors, for example it can hang in infinite loop.
Replace the code with `realpath` from coreutils.

Fixes #1964

### DIFF
--- a/keepalived/Makefile.am
+++ b/keepalived/Makefile.am
@@ -76,25 +76,6 @@ endif
 
 if WITH_IPVS
 install-exec-hook:
-	@( \
-	d=`echo $(bindir) | sed -e "s:^/::"`; \
-	s=`echo $(sbindir) | sed -e "s:^/::"`; \
-	\
-	while [ 1 ]; do \
-		d1=`echo $$d | sed -e "s:/.*::"`; \
-		s1=`echo $$s | sed -e "s:/.*::"`; \
-		if [ $$d1 != $$s1 ]; then \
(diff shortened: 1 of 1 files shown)


eworm-de:
No idea if `realpath` is available on all target platforms, so handle with care!
But it works for me.

pqarmitage:
@eworm-de Many thanks for this patch. Unfortunately the busybox version of realpath does not support `--relative-to`, so I have incorporated your change for use when `realpath --relative-to` is supported.

Issue #2012, opened 2021-10-15
FIFO vrrp_notify_script does not know instance and group states at configuration reload


louis-oui:
**Describe the issue**
I use a vrrp_notify_fifo_script to set/unset iptables filter rules respectively at backup and master state.
When I stop keepalived, the script removes the filters. Basically, I put instructions to remove the rules in the start_shutdown() function of the sample script in the doc.

I noticed that at configuration reload, keepalived stops the script and starts it again.

If the state was backup, the rules are removed at FIFO script stop but not added back.

**To Reproduce**

Set notify-fifo.sh script of script section. Then:
```
ifconfig ens4 192.168.15.41/28 up
ifconfig ens5 192.168.16.41/28 up
kill $(pgrep keepalived)
sleep 1
chmod 700 /tmp/notify*.sh
rm -f /tmp/debug
cat>/etc/keepalived/keepalived.conf <<\EOF
global_defs
{
router_id router
enable_script_security
script_user root
dynamic_interfaces
vrrp_notify_fifo /tmp/notify.fifo
vrrp_notify_fifo_script /tmp/notify-fifo.sh
}

vrrp_sync_group group15 {
group {
vrrp
vrrp2
}
}

vrrp_instance vrrp {
version 2
state BACKUP
interface ens4

use_vmac vrrp

track_file {
}

garp_master_delay 5

virtual_router_id 15

priority 200
advert_int 1.0

virtual_ipaddress {
192.168.15.38/28
}

preempt_delay 0
}

vrrp_instance vrrp2 {
version 2
state BACKUP
interface ens5

use_vmac vrrp2

track_file {
}

garp_master_delay 5

virtual_router_id 15

priority 200
advert_int 1.0

virtual_ipaddress {
192.168.16.38/28
}

preempt_delay 0
}
EOF

keepalived -D
sleep 4
ifconfig ens4 down
sleep 4
iptables-save # DROP rule is there
kill -s SIGHUP $(pgrep keepalived | head -n1) #signal for conf reload
sleep 4
iptables-save # DROP rule is not there anymore but state is still FAULT
cat /tmp/debug
```
Command output
```
root@ubuntu:/tmp# keepalived -D
root@ubuntu:/tmp# sleep 4
root@ubuntu:/tmp# ifconfig ens4 down
root@ubuntu:/tmp# sleep 4
root@ubuntu:/tmp# iptables-save # DROP rule is there
[...]

PR #2014, opened 2021-10-15
vrrp: always notify FIFO script of VRRP states at reload

louis-oui:
Related to issue https://github.com/acassen/keepalived/issues/2012
Inspired by 6c8b51c8ea4836944ff1236e35490190ae9e541c

Configuration reload restarts the notify FIFO script.

If the FIFO script has instructions to (un)set things (e.g. iptables
filter rules) depending on a VRRP state. And if unset things is
requested at script stop, then, at script restart, the new script is
not notified of the VRRP states. Things that needed to be set are not.

This patch always notifies FIFO script of VRRP state at reload.


louis-oui:
Duplicate https://github.com/acassen/keepalived/pull/2020

PR #2030, opened 2021-10-29
fifo: avoid two instances of sample_notify_fifo.sh reading FIFO

louis-6wind:
At reload, the notify FIFO script is restarted.

There is a theoretical risk of corruption of the FIFO input. Since the
notify FIFO script running before the reload and the one running
afterwards both read to the same FIFO file, and if there is a delay in
scheduling the originally executing FIFO script, it is possible that
both instances of the script read the FIFO file at the same time.

This patch use a PID file to make sure that there is no concurrent
reading of the FIFO file.

### DIFF
--- a/doc/samples/sample_notify_fifo.sh
+++ b/doc/samples/sample_notify_fifo.sh
@@ -24,6 +24,7 @@ FIFO=$1
 [[ -z $FIFO ]] && echo "A FIFO name must be specified" && exit 1
 
 LOG_FILE=/tmp/${FIFO##*/}.log
+PID_FILE=/tmp/${FIFO##*/}.pid
 
 stopping()
 {
@@ -32,6 +33,7 @@ stopping()
 
 	[[ $CREATED_FIFO -eq 1 ]] && rm -f $FIFO
 
+	rm -f $PID_FILE
 	exit 0
 }
 
@@ -59,6 +61,17 @@ if [[ ! -p $FIFO ]]; then
 		exit 1
 	fi
 fi
+
+# wait for a previous instance of the script to finish
+if [ -f $PID_FILE ]; then
+	if command -v inotifywait &>/dev/null; then
+		inotifywait -e delete $PID_FILE
+	else
+		while [ -f $PID_FILE ]; do sleep 0.05; done
+	fi
+fi
+
+echo $$ >$PID_FILE
 
 # If keepalived terminates, the FIFO will be closed, so
 # read the FIFO in a loop. It keepalived hasn't opened the


pqarmitage:
@louis-oui You state above `Since the notify FIFO script running before the reload and the one running afterwards both read to the same FIFO file`; I don't think this is strictly correct. After keepalived sends SIGTERM to the old fifo script it unlinks the fifo file. After the reload keepalived creates a new fifo file, and the new script reads from that. So although the fifos both have the same file name, they are different fifos.

The only possibility I can see is that the `while read line; do` loop terminates before the old fifo script receives SIGTERM and it then opens the fifo again, but by then it is the new fifo. However, that would require the script to be able to continue executing without running the SIGTERM signal handler that sets $SHUTDOWN and I don't think this is possible since the SIGTERM is sent before the old fifo file is unlinked.

I have my concerns about using a PID file; it is always possible for an old PID file to remain if for some reason the fifo script terminates abnormally.

I would prefer an approach that checks the inode number of the fifo. This would mean that the fifo script would only ever read from the particular version of the fifo intended for that invocation of the script.

I would be interested on your comments on a patch along the following lines:
```
diff --git a/doc/samples/sample_notify_fifo.sh b/doc/samples/sample_notify_fifo.sh
index 6ca16c54..8124652f 100755
--- a/doc/samples/sample_notify_fifo.sh
+++ b/doc/samples/sample_notify_fifo.sh
@@ -71,9 +71,14 @@ fi
 # When keepalived stops it sends SIGTERM to the script and
 # afterwards send STOPPING messages, so we need to continue
 # reading the FIFO until it is closed.
+FIFO_INODE=$(stat -c "%i" $FIFO)
+[[ $? -ne 0 ]] && echo FIFO $FIFO not accessible >>$LOG_FILE && exit 1
+
 while [[ $SHUTDOWN -eq 0 ]]
 do
        [[ ! -p $FIFO ]] && echo FIFO $FIFO missing >>$LOG_FILE && exit 1
+       exec <$FIFO
+       [[ $? -ne 0 || $(stat -c "%i" $FIFO 2>/dev/null) -ne $FIFO_INODE ]] && break
 
        while read line; do
                PROLOGUE=$(echo "$(date +"%a %b %e %X %Y")": \[$PPID:$$\])
@@ -102,7 +107,7 @@ do
                else
                        echo "$PROLOGUE" $TYPE - unknown "($*)" >>$LOG_FILE
                fi
-       done < $FIFO
+       done
 
        [[ $SHUTDOWN -eq 0 ]] && echo "$PROLOGUE" FIFO CLOSED >>$LOG_FILE
 done
```

louis-oui:
> @louis-oui You state above `Since the notify FIFO script running before the reload and the one running afterwards both read to the same FIFO file`; I don't think this is strictly correct. After keepalived sends SIGTERM to the old fifo script it unlinks the fifo file. After the reload keepalived creates a new fifo file, and the new script reads from that. So although the fifos both have the same file name, they are different fifos.
You are correct. I've made a wrong conclusion because of a mistake in a test.
> The only possibility I can see is that the `while read line; do` loop terminates before the old fifo script receives SIGTERM and it then opens the fifo again, but by then it is the new fifo. However, that would require the script to be able to continue executing without running the SIGTERM signal handler that sets $SHUTDOWN and I don't think this is possible since the SIGTERM is sent before the old fifo file is unlinked.
OK
> I have my concerns about using a PID file; it is always possible for an old PID file to remain if for some reason the fifo script terminates abnormally.
OK
> I would prefer an approach that checks the inode number of the fifo. This would mean that the fifo script would only ever read from the particular version of the fifo intended for that invocation of the script.
> 
> I would be interested on your comments on a patch along the following lines:
> 
> ```
> diff --git a/doc/samples/sample_notify_fifo.sh b/doc/samples/sample_notify_fifo.sh
> index 6ca16c54..8124652f 100755
> --- a/doc/samples/sample_notify_fifo.sh
> +++ b/doc/samples/sample_notify_fifo.sh
> @@ -71,9 +71,14 @@ fi
>  # When keepalived stops it sends SIGTERM to the script and
>  # afterwards send STOPPING messages, so we need to continue
>  # reading the FIFO until it is closed.
> +FIFO_INODE=$(stat -c "%i" $FIFO)
> +[[ $? -ne 0 ]] && echo FIFO $FIFO not accessible >>$LOG_FILE && exit 1
> +
>  while [[ $SHUTDOWN -eq 0 ]]
>  do
>         [[ ! -p $FIFO ]] && echo FIFO $FIFO missing >>$LOG_FILE && exit 1
> +       exec <$FIFO
> +       [[ $? -ne 0 || $(stat -c "%i" $FIFO 2>/dev/null) -ne $FIFO_INODE ]] && break
>  
>         while read line; do
>                 PROLOGUE=$(echo "$(date +"%a %b %e %X %Y")": \[$PPID:$$\])
> @@ -102,7 +107,7 @@ do
>                 else
>                         echo "$PROLOGUE" $TYPE - unknown "($*)" >>$LOG_FILE
>                 fi
> -       done < $FIFO
> +       done
>  
>         [[ $SHUTDOWN -eq 0 ]] && echo "$PROLOGUE" FIFO CLOSED >>$LOG_FILE
>  done
> ```
I am OK with the script. I've tested it with success. However, which situation do you want to cover ? SIGTERM is sent and then FIFO is unlinked. Would it be possible that SIGTERM is not received by the script when the unlink is done ?

louis-oui:
The script is using ``CREATED_FIFO`` to determine wether the FIFO has been created by the script itself. I am also wondering in which case it is useful. Keepalived is supposed to create the FIFO.

pqarmitage:
> However, which situation do you want to cover ? SIGTERM is sent and then FIFO is unlinked. Would it be possible that SIGTERM is not received by the script when the unlink is done ?

It should not be possible for the script not to receive the SIGTERM especially since sending the SIGTERM closing the fifo keepalived does various other cleanup work. However if SIGTERM were not received, keepalived closes the fifo, which means that the read in the notify_fifo script terminates. After that keepalived unlinks the fifo. The script, after the fifo is closed will then check if the fifo file still exists, and if it does it will attempt to reopen it. If it opens the old fifo, the open (or read) will not complete since keepalived does not have the other end open. If it opens the new fifo, then my proposed inode check will ensure that it doesn't read from the new fifo.

One further improvement would be for keepalived to unlink the fifo before it closes it (see `fifo_close()` in `lib/notify.c`), which will mean that the old script cannot attempt to reopen the old fifo during a reload sequence.

louis-oui:
> One further improvement would be for keepalived to unlink the fifo before it closes it (see `fifo_close()` in `lib/notify.c`), which will mean that the old script cannot attempt to reopen the old fifo during a reload sequence.

I would mean that:
```
diff --git a/lib/notify.c b/lib/notify.c
index b5e51c61..174a330a 100644
--- a/lib/notify.c
+++ b/lib/notify.c
@@ -303,12 +303,12 @@ notify_fifo_open(notify_fifo_t* global_fifo, notify_fifo_t* fifo, thread_func_t
 static void
 fifo_close(notify_fifo_t* fifo)
 {
+	if (fifo->created_fifo)
+		unlink(fifo->name);
 	if (fifo->fd != -1) {
 		close(fifo->fd);
 		fifo->fd = -1;
 	}
-	if (fifo->created_fifo)
-		unlink(fifo->name);
 }
 
 void
```
I think we can go for both patch

pqarmitage:
This is superseded by pull request #2031.

louis-oui:
Thank you @pqarmitage for the merge of the pull request #2031 . Now we are sure the instance of FIFO can only read its own FIFO.

However, the patches don't guarantee that commands inside the read loop don't run concurrently. The old script may still do some actions when the new one starts some other.

Note also that although the trap alarm is scheduled 500ms after the shutdown, if command is running in the while loop 500ms after, trap waits for the command to finish before going to stopping function.

I will repost this pull request but with a slight modification. The new script reads the PID value in the PID if any. If the PID is not existing, it continues. Else it waits for the PID to finish before continuing. Eventually we can have a timeout. It rewrites the value in the PID file.

We have to check that it works with two reload in a row.

What do you think of that ?

louis-oui:
> I will repost this pull request but with a slight modification. The new script reads the PID value in the PID if any. If the PID is not existing, it continues. Else it waits for the PID to finish before continuing. Eventually we can have a timeout. It rewrites the value in the PID file.

https://github.com/acassen/keepalived/pull/2034

PR #2034, opened 2021-11-05
fifo: avoid concurrent instances of sample_notify_fifo.sh

louis-oui:
When keepalived reloads its configuration, the notify FIFO script is
restarted. The script that was running before the reload does not end
immediately. Then the new one can run concurrently.

Note that although the ALRM trap is scheduled 500ms after the shutdown,
if a command is running when the 500ms period is over, the script waits
for the command end before calling stopping().

Use a PID file to avoid running concurrent instances of
sample_notify_fifo.sh. If the script exists without removing the PID
file, the new instance checks that no file descriptor is open for the
PID file with the PID that it contains. Checking that the PID does not
exist in the process list is not sufficient as a PID may be reused by
another process.

### DIFF
--- a/doc/samples/sample_notify_fifo.sh
+++ b/doc/samples/sample_notify_fifo.sh
@@ -19,19 +19,40 @@
 
 CREATED_FIFO=0
 SHUTDOWN=0
+TIMEOUT=10
 
 FIFO=$1
 [[ -z $FIFO ]] && echo "A FIFO name must be specified" && exit 1
 
 LOG_FILE=/tmp/${FIFO##*/}.log
 
+if [[ -d /run ]]; then
+    PID_DIR=/run
+elif [[ -d /var/run ]]; then
+    PID_DIR=/var/run
+else
+    PID_DIR=/tmp
+fi
+PID_FILE=$PID_DIR/${FIFO##*/}.pid
+
+exiting()
+{
+	# When this script exists, this function is always executed because
+	# it is associated to the bash EXIT signal.
+    [[ $CREATED_FIFO -eq 1 ]] && rm -f $FIFO
+
+    flock -u $FD
+}
+
+reload_terminate()
+{
+    exit 0
+}
+
 stopping()
 {
 	PROLOGUE=$(echo "$(date +"%a %b %e %X %Y")": \[$PPID:$$\])
 	echo "$PROLOGUE" STOPPING >>$LOG_FILE
-
-	[[ $CREATED_FIFO -eq 1 ]] && rm -f $FIFO
-
 	exit 0
 }
 
@@ -48,7 +69,24 @@ start_shutdown()
 }
 
 trap stopping HUP INT QUIT USR1 USR2 PIPE ALRM
+trap reload_terminate QUIT
 trap start_shutdown TERM
+trap exiting EXIT
+
+exec {FD}>>"$PID_FILE"
+if ! flock -e -n $FD; then
+		# Send SIGQUIT signal to the previous instance of this script.
+		# The previous script waits for its current executed command to end
+		# before actually calling reload_terminate(). The TIMEOUT value must
+		# take this constraint into account.
+       OLD_PID=$(cat $PID_FILE)
+       if ls -l /proc/$OLD_PID/fd | grep -qw "$(readlink -f "$PID_FILE")"
+       then
+			kill -QUIT $OLD_PID
+       fi
+       flock -e --timeout $TIMEOUT $FD || exit 1
+fi
+echo $$ >"$PID_FILE"
 
 if [[ ! -p $FIFO ]]; then
 	mkfifo $FIFO


doc/samples/sample_notify_fifo.sh
```diff
@@ -112,4 +125,6 @@ do
 	[[ $SHUTDOWN -eq 0 ]] && echo "$PROLOGUE" FIFO CLOSED >>$LOG_FILE
 done
 
+rm -f "$PID_FILE"
+exec {FD}>&-
```
pqarmitage, line 129:
```
rm -f "$PID_FILE"
exec {FD}>&-
```
can be removed if there is the EXIT trap.

doc/samples/sample_notify_fifo.sh
```diff
@@ -37,6 +38,7 @@ stopping()
 
 start_shutdown()
 {
+	[[ ! -f "$PID_FILE" || "$(cat "$PID_FILE")" -ne $$ ]] && exit 0
```
pqarmitage, line 41: I don't understand the reason for this. What is the scenario you are protecting against?

doc/samples/sample_notify_fifo.sh
```diff
@@ -60,6 +62,17 @@ if [[ ! -p $FIFO ]]; then
 	fi
 fi
 
+# wait for a previous instance of this script to finish its job
+while [ -f "$PID_FILE" ] && \
+	ls -l /proc/$(cat "$PID_FILE")/fd | grep -qw "$(readlink -f "$PID_FILE")"
+do
+	sleep 0.05
+done
```
pqarmitage, line 70: Should there be a time limit (or loop count limit) here so that the script does not wait indefinitely for a previous instance to terminate? Perhaps if the previous instance of the script hasn't terminated after a couple of seconds it should be killed.

pqarmitage (changes requested):
It took me a while to understand what was this commit was doing, but I got there eventually. I do however think it is important to ensure that the PID_FILE is always removed, and that /var/run is not created if it doesn't exist.

doc/samples/sample_notify_fifo.sh
```diff
@@ -24,6 +24,7 @@ FIFO=$1
 [[ -z $FIFO ]] && echo "A FIFO name must be specified" && exit 1
 
 LOG_FILE=/tmp/${FIFO##*/}.log
+PID_FILE=/var/run/${FIFO##*/}.pid
```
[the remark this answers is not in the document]
louis-oui: > although I am not sure that [[ -n $FD ]] && exec {FD}>&- is needed since the script is exiting.

The file descriptor is always removed when the script exits.
> `trap exiting EXIT`

Does signal EXIT exist ? I am not sure

doc/samples/sample_notify_fifo.sh
louis-oui: One of my collegues suggested to me to replace the loop by a flock.
Basically, flock takes a exclusive lock on the FD until FD is removed with `exec {FD}>&-` or at exit.
```
exec {FD}>>"$PID_FILE"
if ! flock -e $FD --timeout 10; then
	if ls -l /proc/$(cat "$PID_FILE")/fd | grep -qw "$(readlink -f "$PID_FILE")"
	then
		kill $(cat "$PID_FILE")
	else
		exit 1
	fi
	flock -e $FD --timeout 10 || exit 1
fi
echo $$ >"$PID_FILE"
```
We can use flock without the timeout. Just:
`flock -e $FD`

However it would wait indefinitely if the previous script does not remove the the FD or exit.

doc/samples/sample_notify_fifo.sh
louis-oui: It was not useful with flock. See next my comment

louis-oui:
> It took me a while to understand what was this commit was doing, but I got there eventually. I do however think it is important to ensure that the PID_FILE is always removed, and that /var/run is not created if it doesn't exist.

Thank you for review @pqarmitage

louis-oui:
Set WIP because I need to test the new version of the script

doc/samples/sample_notify_fifo.sh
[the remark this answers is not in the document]
pqarmitage: > The file descriptor is always removed when the script exits.

Are you confirming that `[[ -n $FD ]] && exec {FD}>&-` is not needed?

In addition to the real signals, the valid sigspecs for trap also include EXIT, DEBUG, RETURN and ERR. As the bash man page says:
> If a sigspec is EXIT (0) the command arg is executed on exit from the shell.

pqarmitage:
@louis-oui Many thanks for the updates. I will have a look at this later this evening or tomorrow.

doc/samples/sample_notify_fifo.sh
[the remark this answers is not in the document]
louis-oui: > Are you confirming that [[ -n $FD ]] && exec {FD}>&- is not needed?

Not needed

doc/samples/sample_notify_fifo.sh
pqarmitage: I note that the man page for flock(1) states that the syntax is `flock [options] number`, and so to comply with that I think it should be `flock -e --timeout 10 $FD`.

I think the timeout should be more like 1 second or even less rather than 10 seconds; the previous instance of the script should be terminating within a few milli-seconds of the script receiving SIGTERM. Perhaps have `TIMEOUT=0.3` near the beginning of the script and use `flock -e --timeout $TIMEOUT $FD` in the two places where flock is called. Users who are then using the sample script can modify $TIMEOUT as fits their needs.

doc/samples/sample_notify_fifo.sh
```diff
@@ -24,9 +24,24 @@ FIFO=$1
...
+    PID_DIR=/var/run
+else
+    PID_DIR=/tmp
+fi
+PID_FILE=$PID_DIR/${FIFO##*/}.pid
+
+exiting()
+{
+	rm -f "$PID_FILE"
+	[[ -n $FD ]] && exec {FD}>&-
```
pqarmitage, line 39: I suggest remove
   `[[ -n $FD ]] && exec {FD}>&-`
since it is not needed.

Also add:
```
    [[ $CREATED_FIFO -eq 1 ]] && rm -f $FIFO

    flock -u $FD
```

Adding `flock -u $FD` significantly speeds up the `flock -e`. In my tests, without `flock -u` the `flock -e` took 9.78 seconds, with the `flock -u` it took 0.000008 seconds.

Moving the `rm -f $FIFO` from `stopping()` to `exiting()` ensures that the FIFO is always removed if we created it. Removing the FIFO before releasing the lock ensures that we don't remove the FIFO after the new instance of the script has checked for its existence (providing the lock is obtained before the FIFO is created - see below).

doc/samples/sample_notify_fifo.sh
```diff
@@ -24,9 +24,24 @@ FIFO=$1
...
+}
 
 stopping()
 {
+	[ -f "$PID_FILE" ] && exiting
 	PROLOGUE=$(echo "$(date +"%a %b %e %X %Y")": \[$PPID:$$\])
 	echo "$PROLOGUE" STOPPING >>$LOG_FILE
```
pqarmitage, line 47: Remove the `[[ $CREATED_FIFO -eq 1 ]] && rm -f $FIFO` since it is now in `exiting()`.

doc/samples/sample_notify_fifo.sh
```diff
@@ -47,6 +62,7 @@ start_shutdown()
 	) &
 }
```
pqarmitage, line 64: Add:
```
reload_terminate()
{
    exit 0
}
```

and
`trap reload_terminate QUIT`

so the script knows why it is being killed and can cleanup appropriately. This also means that the `flock -u` is executed.

doc/samples/sample_notify_fifo.sh
```diff
@@ -60,6 +76,18 @@ if [[ ! -p $FIFO ]]; then
...
+exec {FD}>>"$PID_FILE"
+if ! flock -e $FD --timeout 10; then
+	if ls -l /proc/$(cat "$PID_FILE")/fd | grep -qw "$(readlink -f "$PID_FILE")"
+	then
+		kill $(cat "$PID_FILE")
+	else
+		exit 1
+	fi
+	flock -e $FD --timeout 10 || exit 1
+fi
```
pqarmitage, line 88: Change the `if ... fi` block to:
```
if ! flock -e -n $FD; then
       OLD_PID=$(cat $PID_FILE)
       if ls -l /proc/$OLD_PID/fd | grep -qw "$(readlink -f "$PID_FILE")"
       then
               kill -QUIT $OLD_PID
       fi
       flock -e --timeout 10 $FD || exit 1
fi
```
`flock -e $FD --timeout 10` fails for me due to the options not being before the FD. Also we don't need to delay if the first flock cannot obtain the lock.

Saving the PID in $OLD_PID saves reading the PID file twice.

Sending SIGQUIT allows the old instance of the script to know why it is being signalled, and the EXIT trap to be executed.

The old instance of the script may be terminating while this part of the new instance is being executed, and so the old instance may remove the PID_FILE. This would mean that `ls -l /proc/$OLD_PID/fd` would fail, and we don't then want this instance of the script to terminate, so the `else exit 1` needs to be removed.

Also, move this block obtaining the lock to before checking for the existence of the FIFO to avoid a potential race of the old script removing the FIFO and this script creating the FIFO.

pqarmitage (changes requested):
I have added 4 more comments which I think improve how the locking is handled.

louis-oui:
Thank you for you review @pqarmitage . I have removed the WIP label

PR #2047, opened 2021-11-20
Docs: adding security policy documentation

johannbg:
Adding a missing security policy documentation with a pretty standard community workflow.
Added Alexandre address as the contact for such incidents but perhaps Quentin should be that person instead. 
If more than one person should be notified, typically virtual email address is used or the text points to notify the maintainers listed in the MAINTAINERS file that typically resides in a repository.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,7 @@
+# Security Policy
+
+Here at the Keepalived community security is very important to us.
+
+## Reporting a Vulnerability
+If you discover any issue regarding security, we'd appreciate a non-public disclosure of the information, so please disclose the information responsibly by sending an email to Alexandre Cassen <acassen@gmail.com> and not by creating a GitHub issue or start a dialog in the keepalived community at groups.io. 
+We will respond swiftly to fix verifiable security issues along with the disclosure being coordinated with vendors and other relevant security teams.


pqarmitage:
@johannbg Unfortunately I don't think this document is helpful to the keepalived project.

I find statements that something "is very important to us" to be meaningless, due to the complete lack of measurability or direction. For example someone who seeks to install a trojan could make that claim on the basis that it is important to them to circumvent security. For what it is worth, whenever I see an organisation making such a statement I interpret it as meaning they want to make us think they think it is important, whereas in reality they have little or no interest, since it is easy to produce much more meaningful statements.

The sentence relating to discovering a security vulnerability can be interpreted as meaning that starting a dialogue at groups.io is an acceptable way of communicating the vulnerability, completely contrary to what I suspect is intended.

Finally, it refers to disclosure being coordinated with vendors. We do not have a comprehensive list of which vendors include keepalived in their products or services, which is a clear consequence of being an open source project, and so it is not possible for us to commit to making such disclosures.

johannbg:
@pqarmitage well distributions are usually considered vendors these days and I assume keepalived developer are in good relationship with downstream distribution anyways security policy is a good thing to have so I would recommend that you implement one to your liking.

PR #2048, opened 2021-11-20
Improve handling of kernel module loading

pqarmitage:


johannbg:
@pqarmitage out of curiosity, which distributions are being targeted by that rpm spec file?

pqarmitage:
@johannbg There is no distribution in particular being targeted. It is there essentially as a template in case someone wants to build an rpm.

PR #2050, opened 2021-11-20
github: add a stale bot to close stale things

johannbg:
Given that I had a PR open for what 2 years without addressing it, let's add a stale bot to address such inactivities...

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,21 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 30
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - documents
+  - enhancement
+  - regression
+  - wiki
+  - bug
+# Label to use when marking an issue as stale
+staleLabel: stale
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue is being marked as stale because it has not had any recent activity. 
+  It will be closed if no further activity occurs. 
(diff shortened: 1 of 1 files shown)


johannbg:
Hmm I probably should mention that you need to install the stale app prior to merging this encase ye are unfamiliar how apps work on github.

pqarmitage:
I find it useful to be able to have old issues still open. We have several that are of low impact, and I will deal with when and if I have the time, or if more people indicate that they are suffering from the issue.

Since we have so few open issues I think it is probably easiest to manage them manually.

pqarmitage:
@johannbg You referred to your pull request (#1426) that was open for a couple of years. I would look at it every so often but I could not think of a use case for it, so I left it open in case it was subsequently useful. If would be very helpful if you could explain the benefits of having an mkosi image.

Also you stated `mkosi even has it's own github action these days`. Can you provide the necessary files for using that. Does it mean that every time there is a new merge that a new image is built?

johannbg:
mkosi is just an (  container/VM ) image building tool which is useful in wide varity of situations and mostly used for testing in projects ( like here https://github.com/lucasdemarchi/kmod/tree/master/testsuite/mkosi and ofcourse systemd itself etc ) or by developers themselves to build locally images and run test cases in them prior to submitting PR's ( a good read on how mkosi is being used in a project is actually the systemd hacking page https://github.com/systemd/systemd/blob/main/docs/HACKING.md and ofcourse mkosi man page https://github.com/systemd/mkosi/blob/main/mkosi.md  ) and it's being used in some infrastructures to build small vm/container images which are then run by systemd-nspawn and reduces overhead that something like docker brings ( I think Facebook uses systemd-nspawn for their entire container infrastructure. ) 

Anyways given that there is no need for a stale bot for the project I'm just going to close this PR.

PR #2089, opened 2022-01-28
doc: update pre-requisites installation

guits:
in rhel based OS prerequisites section, it says to install `automake`
but I was unable to build keepalived from source without installing `make` instead.

Also, `ipset-devel`, `libnfnetlink-devel` and `file-devel` packages require the repo
`PowerTools` to be enabled so I thought it was worth it to mention how to install them.

Signed-off-by: Guillaume Abrioux <gabrioux@redhat.com>

### DIFF
--- a/doc/source/installing_keepalived.rst
+++ b/doc/source/installing_keepalived.rst
@@ -42,8 +42,13 @@ Install Prerequisites on RHEL/CentOS/Fedora
 On RHEL, Centos, Fedora etc install the following prerequisites
 (on older systems replace dnf with yum)::
 
-    dnf install curl gcc autoconf automake openssl-devel libnl3-devel \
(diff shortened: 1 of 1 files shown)


pqarmitage:
@guits Many thanks for reporting this. There are one or two issues with your pull request however.

1. `automake` is required, so `make` needs to be added, rather than replace `automake`.
2. Although ipset-devel is in the `PowerTools` repo on Centos8, it is in the `codeready-builder-for-rhel-8-x86_64-rpms` (or whatever arch you are using) on RHEL8. ON Fedora, no repo needs to be specified.

pqarmitage:
Commits b609573 and 5faa789 update doc/source/installing_keepalived.rst.

Many thanks for identifying this issue.

PR #2180, opened 2022-08-12
Avoid terminating keepalived during a configuration test.

oxpa:
When a configuration is checked keepalived also checks script security and calls `check_script_secure()`.
The same function is also called during normal keepalived lifecycle. The function has a check that the main pid didn't die while we were checking other things (see 59271723ff8e311754878cb6f75ce4dbf18aa462 for details)
`main_pid` is set in `start_keepalived()` function called from `keepalived_main()`. In `keepalived_main()` we parse args, decide if we have to test configuration and do the check before calling `start_keepalived()`. So `main_pid` is not set to a proper value. 
At the same time, the problem `main_pid` solves is not really relevant at a configuration test stage. So the easiest fix, IMO, is to verify `main_pid` was set properly . If not - it's a configuration test and we can skip the check.

The problem only arises in `notify.c` for me. But the same check is placed in `check_daemon.c`, `vrrp_daemon.c` and `bfd_daemon.c`. I changed those as well just for good measure.

### DIFF
--- a/keepalived/bfd/bfd_daemon.c
+++ b/keepalived/bfd/bfd_daemon.c
@@ -391,7 +391,7 @@ start_bfd_child(void)
 	prctl(PR_SET_PDEATHSIG, SIGTERM);
 
 	/* Check our parent hasn't already changed since the fork */
-	if (main_pid != getppid())
+	if (main_pid != 0 && main_pid != getppid())
 		kill(getpid(), SIGTERM);
 
 	prog_type = PROG_TYPE_BFD;
--- a/keepalived/check/check_daemon.c
+++ b/keepalived/check/check_daemon.c
@@ -695,7 +695,7 @@ start_check_child(void)
 	prctl(PR_SET_PDEATHSIG, SIGTERM);
 
 	/* Check our parent hasn't already changed since the fork */
-	if (main_pid != getppid())
+	if (main_pid != 0 && main_pid != getppid())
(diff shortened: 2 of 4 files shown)


pqarmitage:
@oxpa Many thanks for reporting this and identifying the issue. When I added the config test option, I wanted to do so with minimal code changes (so that the tests were as similar as possible to running keepalived "live"). It seems that in doing so I missed this aspect of how running a config test is different.

Whilst you have used `main_pid != 0` as your test for not running a config test, I would prefer to be more explicit about it, i.e. `!__test_bit(CONFIG_TEST_BIT, &debug)` since that makes it clear what is really being tested for.

Regarding your second patch, I am always nervous when a patch has been submitted when it hasn't been tested (a compilation failure is a sure sign of that) :). Having checked the code, the second patch does not appear to be necessary, since the *_daemon() functions are only called if `!__test_bit(CONFIG_TEST_BIT, &debug)`, i.e. main_pid != 0.

I have merge commit #aae8e91 which is your commit #4e31258 updated to use `!__test_bit(CONFIG_TEST_BIT, &debug)`, rather than `main_pid !- 0`. I don't see that commit #d92a67f or its update is needed.

Many thanks for identify the issue and for resolution.

oxpa:
to be honest, I wasn't sure if `test_bit` would work and only tested the first (happily merged) patch. 
So your notes are spot on :)

Thanks!

Issue #2219, opened 2022-10-27
VRRP stays in FAULT state after adding a same IP address as VIP on physical interface


lioyerar:
**Describe the bug**
Hello,

I am in a Qemu virtual machine and compiled the newest keepalived version.
I started keepalived with a VIP 192.168.1.1/24 and the physical interface has no IP addresses yet.
It is normal for VRRP entering FAULT state as no IPv4 address is configured on the interface.

Then, I add 192.168.1.1/24 (same as the VIP) on the physical interface, VRRP is still in FAULT state.

However, either if I add another IP address different than the VIP or before starting keepalived, configure the interface with a IP address (whatever the address) , VRRP works successfully.

**To Reproduce**
- No IP address on the physical interface
- Start keepalived with the configuration file and VRRP should be in FAULT state
- Add same IP address as VIP on the physical interface: ip a a 192.168.1.1/24 dev ens4

**Expected behavior**
The transition from FAULT state to BACKUP then from BACKUP state to MASTER state.

**Keepalived version**
```
Keepalived v2.2.7 (10/13,2022), git commit v2.2.7-111-gefef3596 
```

**Distro (please complete the following information):**
 - Name: Ubuntu
 - Version: 20.04
 - Architecture: x86_64

**Details of any containerisation or hosted service (e.g. AWS)**
Qemu virtual machine

**Configuration file:**
```
global_defs
{
  router_id router
  enable_script_security
  script_user root
  dynamic_interfaces
  vrrp_startup_delay 0
  disable_local_igmp
}


vrrp_instance vrrp {
  version 2
  state BACKUP
  interface ens4

  use_vmac vrrp

  virtual_ipaddress {
    192.168.1.1/24
  }

  track_file {
  }

  garp_master_delay 5

  virtual_router_id 12

  priority 100
  advert_int 1.0

  preempt_delay 0
  notify_deleted
}

```

**System Log entries**
```
Oct 27 14:21:41 ubuntu2004 Keepalived[8087]: Starting Keepalived v2.2.7 (10/13,2022), git commit v2.2.7-111-gefef3596                                                                                                           

Oct 27 14:21:41 ubuntu2004 Keepalived[8087]: WARNING - keepalived was built for newer Linux 5.4.203, running on Linux>                                                                                                                       

Oct 27 14:21:41 ubuntu2004 Keepalived[8087]: Command line: 'keepalived' '-D'                                                                                                                                                                 

[...]

PR #2227, opened 2022-11-15
netlink: do not ignore vip as system owner

lioyerar:
When an instance is configured with 255 priority and is waiting for an address to start, do not ignore when the vip is added.

Signed-off-by: Loïc Sang <loic.sang@6wind.com>


lioyerar:
Hello,

As discussed in:
https://github.com/acassen/keepalived/issues/2219

Is it possible to integrate this commit to make VRRP work as a simple router?

PR #2245, opened 2023-02-21
ipvs: fix issue in reload process when using virtual server groups

lewislone:
issue: when using virtual server groups, remove vs entry in configure file and then do reload, vs entry can not be removed. And add vs entry in configure file and the do reload, rs with 0 port will be set.

fix: in reload process do the same action with ipvs_group_cmd. set rs port with vs port and update live state


lewislone:
issue desc: [2253](https://github.com/acassen/keepalived/issues/2253)

pqarmitage:
@lewislone Many thanks for this patch. Unfortunately it did not compile since the added `drule->user.port`s should have been `drule.user.port`. I also found a problem if using `global_defs` `lvs_flush_on_stop` relating to removing a virtual server group entry.

Your (modified) patch is now applied as commit d07d122 (pull request #2345) duly referencing you.

PR #2246, opened 2023-03-02
vrrp: fix VRRP instance remaining in FAULT state forever

louis-6wind:
Since 7fea6cc2dc ("Fix netlink_if_link_populate at startup"), the interface flags (ifi_flags) are updated directly from netlink_if_link_populate() without using update_interface_flags(). ifi_flags is now updated from two places.

7fea6cc2dc ("Fix netlink_if_link_populate at startup") has inverted in netlink_if_link_filter() the order of calls of
netlink_if_link_populate() and update_interface_flags(). Since this commit, ifi_flags is already updated by netlink_if_link_populate() when calling update_interface_flags() so that update_interface_flags() returns without dealing with interface up/down state change.

As a consequence, some VRRP instance may remain in FAULT state forever unless an interface flapping occurs.

The same kind of issue was existing in netlink_link_filter() before f2d5cbea50 ("Force seeing a transition to up state when an interface is created").

Do not update ifi_flags in netlink_if_link_populate() anymore because update_interface_flags() do it later. Update ifi_flags before update_added_interface() to keep the same behavior.

Fixes: 7626dbbb7e ("Tidying up keepalived_netlink.c")
Fixes: 7fea6cc2dc ("Fix netlink_if_link_populate at startup")


pqarmitage:
@louis-6wind Many thanks for this patch.

Unfortunately it doesn't work for me (from what I can see you must be using unicast with VMACs for your VRRP instances). With your patch applied, I get the following error:
`(VI_1) interface eth0 does not support multicast, specify unicast peers - disabling`
This is because, since the flags are no longer being set on the interfaces, the check for multicast fails in vrrp_complete_instance().

I tried the following patch:
```
diff --git a/keepalived/core/keepalived_netlink.c b/keepalived/core/keepalived_netlink.c
index 57c9ab80..8295f308 100644
--- a/keepalived/core/keepalived_netlink.c
+++ b/keepalived/core/keepalived_netlink.c
@@ -1946,7 +1946,9 @@ netlink_if_link_populate(interface_t *ifp, struct rtattr *tb[], struct ifinfomsg
 	ifp->rp_filter = UINT_MAX;	/* We have not read it yet */
 #endif
 
-	ifp->ifi_flags = ifi->ifi_flags;
+	/* We need to see a transition to up, so mark it down for now */
+	ifp->ifi_flags = ifi->ifi_flags & ~(IFF_UP | IFF_RUNNING);
+
 	if (FLAGS_UP(ifi->ifi_flags))
 		ifp->seen_up = true;
 
@@ -2262,9 +2264,6 @@ netlink_link_filter(__attribute__((unused)) struct sockaddr_nl *snl, struct nlms
 #endif
 				if (ifp->mtu > old_mtu)
 					alloc_vrrp_buffer(ifp->mtu);
-
-			/* We need to see a transition to up, so mark it down for now */
-			ifp->ifi_flags &= ~(IFF_UP | IFF_RUNNING);
 		} else {
 			if (__test_bit(LOG_DETAIL_BIT, &debug))
 				log_message(LOG_INFO, "Unknown interface %s deleted", (char *)tb[IFLA_IFNAME]);
```
and although it works when using VMACs, it does not work when not using VMACs since the interface state remains down (when a VMAC is created, we get further netlink messages for the underlying interface and hence update ifp->ifi_flags).

Can you provide details of the problem you are experiencing, in particular your keepalived configuration and a description of how to trigger the problem. It would be really helpful if I could reproduce the problem so that I can test what fixes will work both for you and for other configurations and I am struggling to see what it is in the current code that is not working.

louis-6wind:
> although it works when using VMACs, it does not work when not using VMACs since the interface state remains down (when a VMAC is created, we get further netlink messages for the underlying interface and hence update ifp->ifi_flags).

I was actually because of that in update_interface_flags():

```
	if (!vrrp_data)
		return;
```

louis-6wind:
I notice the issue with the following script and a custome patch. There is a loop that breaks when the issue appears.
```
ip link set ens4 down
ip address add dev ens4 10.0.0.2/24
cat> /etc/keepalived/keepalived.conf <<\EOF
global_defs
{
  router_id router
  enable_script_security
  script_user root
  dynamic_interfaces
  disable_local_igmp
}

vrrp_instance vrrp {
  version 2
  state BACKUP
  interface iface4
  use_vmac vrrp

  track_file {
  }

  garp_master_delay 5

  virtual_router_id 15

  priority 100
  advert_int 1.0

  virtual_ipaddress {
    2001::1/64
  }

  preempt_delay 0
  notify_deleted

}
EOF

while [ 1 ]
do
sleep 1
killall keepalived; sleep 1; ip link set iface4 down; ip link set iface4 name ens4
ip -br a
sleep 1

keepalived -D; sleep 3;  ip link set ens4 name iface4 up
sleep 3
journalctl _PID=$(pgrep keepalived | tail -n1) -o short-precise | cat
journalctl _PID=$(pgrep keepalived | tail -n1) -o short-precise | grep BACKUP || break
done
```
The custom patch is https://github.com/acassen/keepalived/pull/2262/commits/8aa95d0f3cfbc08b0c7eba98db8a42528db221e8 for DPDK with VMAC. I upload it in case it can interest some other people.

The patch triggered the missing state link issue with VMAC but it seems that the issue can happen in some other conditions

Issue #2254, opened 2023-03-14
vrrp master switchover lead to disconnection of tcp connections


maryoho:
**Describe the bug**

On my network experiment，I runned keepalived on two nat device to implement high availability. when master changed to another, 
a script will be executed to commit nat sessions to kernel.
The sequence of actions:
1.  set vip to new master network interface
2. send gratuitous ARP
3. execute script
The third step will take some time, it will lead to connection reset if client or server send packet before it's nat session not synced at this time.
![image](https://user-images.githubusercontent.com/26035452/224918013-c16a2cf4-896f-4d11-af25-074eda32ece2.png)
 
**Expected behavior**
tcp connection not be resetted

**Keepalived version**
```
Keepalived v2.1.5 (07/13,2020)
```

**Configuration file:**
```
node1:
global_defs {
    router_id 172.18.0.2
    vrrp_skip_check_adv_addr
    vrrp_garp_master_refresh 60
    vrrp_garp_master_refresh_repeat 2
    vrrp_garp_master_repeat 5
    vrrp_garp_interval 0.001
}

vrrp_instance VI_1 {
    state MASTER
    interface eth0
    virtual_router_id 51

    unicast_src_ip 172.18.0.2

    unicast_peer {
        172.18.0.3
    }

    priority 100
    advert_int 1
    authentication {
        auth_type PASS
        auth_pass 1111
    }

    virtual_ipaddress {
        172.18.0.254
    }
    notify_master "/etc/conntrackd/primary-backup.sh primary"
    notify_backup "/etc/conntrackd/primary-backup.sh backup"
    notify_fault "/etc/conntrackd/primary-backup.sh fault"
}

node2:
global_defs {
    router_id 172.18.0.3
    vrrp_skip_check_adv_addr
    vrrp_garp_master_refresh 60
    vrrp_garp_master_refresh_repeat 2
    vrrp_garp_master_repeat 5
    vrrp_garp_interval 0.001
}

vrrp_instance VI_1 {
    state MASTER
    interface eth0
    virtual_router_id 51

    unicast_src_ip 172.18.0.3

    unicast_peer {
        172.18.0.2
    }

    priority 100
    advert_int 1
    authentication {
        auth_type PASS
        auth_pass 1111
    }

    virtual_ipaddress {
        172.18.0.254
    }
    notify_master "/etc/conntrackd/primary-backup.sh primary"
    notify_backup "/etc/conntrackd/primary-backup.sh backup"
[...]

PR #2255, opened 2023-03-14
fix issue: vrrp master switchover lead to disconnection of tcp connections

maryoho:
solve https://github.com/acassen/keepalived/issues/2254
add new config param exec_before_vip to vrrp_instance
 if this is set, notify_master script will be executed before set vip when master changed

### DIFF
--- a/doc/source/configuration_synopsis.rst
+++ b/doc/source/configuration_synopsis.rst
@@ -142,6 +142,7 @@ VRRP Instance Definitions Synopsis
         **priority** num
         **advert_int** num
         **smtp_alert**
+        **exec_before_vip**
         **authentication** {
             **auth_type** *PASS|AH*
             **auth_pass** string
@@ -185,6 +186,7 @@ notify_master               specify a shell script to be executed during transit
 notify_backup               specify a shell script to be executed during transition to backup state path
 notify_fault                specify a shell script to be executed during transition to fault state  path
 vrrp_sync_group             Identify the VRRP synchronization instances group                       string
+exec_before_vip             notify_master script executed before vip setted if this setted true
 ==========================  ======================================================================= =========
 
 Path type: A system path to a script eg: “/usr/local/bin/transit.sh <arg_list>”
--- a/keepalived/include/vrrp.h
+++ b/keepalived/include/vrrp.h
@@ -368,7 +368,8 @@ typedef struct _vrrp_t {
 	notify_script_t		*script_deleted;
 	notify_script_t		*script_master_rx_lower_pri;
 	notify_script_t		*script;
-	int			notify_priority_changes;
+    int         exec_before_vip; /* exec script_master first before set vip address */
+    int			notify_priority_changes;
 
 	/* rfc2338.6.2 */
 	uint32_t		ms_down_timer;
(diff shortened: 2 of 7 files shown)


pqarmitage:
Unfortunately we cannot merge this commit for a couple of reasons.

1. The formatting of the code is not consistent with the style used for keepalived; not only does the added code not comply, but the formatting of some of the existing code has been modified so that it no longer applies. They key issue is that keepalived code uses tabs for indentation, with the tab width being 8 spaces. The code in the patch appears to be using 4 spaces per indent. This is cosmetic and straight forward to resolve.

2.  This issue is fundamental. You cannot use wait() or any of its associated functions (e.g. waitpid()) if that might cause keepalived to be blocked. The reason for this is that keepalived might have several VRRP instances configured, and it needs to be able to continue operating all of them, even while scripts are being executed. 

I prefer something like `notify_script_before_master` to your suggested `exec_before_vip`.

pqarmitage:
Unfortunately we cannot merge this commit for a couple of reasons.

1. The formatting of the code is not consistent with the style used for keepalived; not only does the added code not comply, but the formatting of some of the existing code has been modified so that it no longer complies. They key issue is that keepalived code uses tabs for indentation, with the tab width being 8 spaces. The code in the patch appears to be using 4 spaces per indent. This is cosmetic and straight forward to resolve.

In addition, `system_call_script_0()` should be before `system_call_script()` in the source file, since `system_call_script()` calls `system_call_script_0()`. I also think `system_call_script_0()` should be called something like `system_call_script_wait()` to be more descriptive. 

2. This point is fundamental. You cannot use wait() or any of its associated functions (e.g. waitpid()) if that might cause keepalived to be blocked. The reason for this is that keepalived might have several VRRP instances configured, and it needs to be able to continue operating all of them, even while scripts are being executed.

Also consider a VRRP instance with three servers, priorities say 100, 102, 104. If the system with priority 104 fails, the system with priority 102 should take over as master, but if there is a delay while a script is executed, then the system with priority 100 may take over as master initially, and then when the system with priority 102 completes executing the script, it would then become master. 

`system_call_script()` already has the mechanism for handling actions to be taken after a script completes. If parameter `func` is not NULL, then `func()` is scheduled to be run once the script has completed executing. Suppose func is vrrp_master_after_notifies(), then it could be something like:
```
vrrp_master_after_notifies(thread_ref_t thread)
{
        vrrp_t vrrp = THREAD_ARG(thread);

        vrrp_state_become_master(vrrp);
        /*
          * If we catch the master transition
           * register a gratuitous arp thread delayed to garp_delay secs.
           */
        if (vrrp->garp_delay)
                 thread_add_timer(master, vrrp_gratuitous_arp_thread,
                                         vrrp, vrrp->garp_delay);
}
```
so that it has the actions in `vrrp_master_state_tx()` that would normally be executed in that function if `exec_before_vip` is not set. Obviously, vrrp_master_state_tx() would need to return after calling `send_instance_notifies()` if exec_before_vip is set.

BTW I prefer something like `notify_script_before_master` to your suggested `exec_before_vip` configuration option.

acassen:
no news for more than 3years. Closing

PR #2262, opened 2023-03-23
vrrp: work around missing promiscuous netlink notifications

louis-6wind:
If the base interface does not implement IFF_UNICAST_FLT, for example
it is a bridge interface, no Netlink notification is sent by the kernel
when promiscuity is set on the base interface.

The promiscuous state of the base interface is correct in kernel but it
is in incorrect in daemons that listen to the interface netlink messages
(eg. DPDK).

The issue is still there in kernel 5.4.0

Force a notification by re-setting IFLA_GROUP for the base interface.

### DIFF
--- a/keepalived/core/keepalived_netlink.c
+++ b/keepalived/core/keepalived_netlink.c
@@ -1853,6 +1853,8 @@ netlink_if_link_populate(interface_t *ifp, struct rtattr *tb[], struct ifinfomsg
 	ifp->mtu = *PTR_CAST(uint32_t, RTA_DATA(tb[IFLA_MTU]));
 	ifp->hw_type = ifi->ifi_type;
 
+	ifp->group = *(uint32_t *)RTA_DATA(tb[IFLA_GROUP]);
+
 	if (!netlink_if_get_ll_addr(ifp, tb, IFLA_ADDRESS, name))
 		return false;
 	if (!netlink_if_get_ll_addr(ifp, tb, IFLA_BROADCAST, name))
--- a/keepalived/include/vrrp_if.h
+++ b/keepalived/include/vrrp_if.h
@@ -139,6 +139,7 @@ typedef struct _interface {
 	unsigned		up_debounce_timer;
 	unsigned		down_debounce_timer;
 	uint32_t		mtu;			/* MTU for this interface_t */
+	uint32_t		group;			/* GROUP for this interface_t */
 	unsigned short		hw_type;		/* Type of hardware address */
 	u_char			hw_addr[MAX_ADDR_LEN];	/* MAC address */
 	u_char			hw_addr_bcast[MAX_ADDR_LEN]; /* broadcast address */
--- a/keepalived/vrrp/vrrp_vmac.c
+++ b/keepalived/vrrp/vrrp_vmac.c
@@ -230,6 +230,33 @@ netlink_link_up(vrrp_t *vrrp)
 	return status;
 }
 
+static int
+netlink_link_group(interface_t *base_ifp)
+{
+	int status = 1;
+	uint32_t group = base_ifp->group;
+	struct {
+		struct nlmsghdr n;
+		struct ifinfomsg ifi;
+		char buf[256];
+	} req;
+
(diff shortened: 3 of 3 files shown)


louis-6wind:
To reproduce the issue.
```
cat> /etc/keepalived/keepalived.conf <<\EOF
global_defs
{
  router_id router
  enable_script_security
  script_user root
  dynamic_interfaces
  disable_local_igmp
}

vrrp_instance vrrp {
  version 2
  state BACKUP
  interface iface4
  use_vmac vrrp

  track_file {
  }

  garp_master_delay 5

  virtual_router_id 15

  priority 100
  advert_int 1.0

  virtual_ipaddress {
    2001::1/64
  }

  preempt_delay 0
}
EOF


killall keepalived ip
ip link set ens4 down

ip link del iface4
ip link add iface4 type bridge
ip link set ens4 master iface4

sleep 1
ip link set ens4 up
ip link set iface4 up
sleep 2
ip -d l show iface4
keepalived -D
sleep 3
```
Without the patch, the interface has no promiscuous in DPDK. The patch solves the issue.

pqarmitage:
I don't understand why this patch is needed.

Kernel commit 01789349ee52e (which introduced IFF_UNICAST_FLT) states:
```
    net: introduce IFF_UNICAST_FLT private flag
    
    Use IFF_UNICAST_FTL to find out if driver handles unicast address
    filtering. In case it does not, promisc mode is entered
```

This implies that if the interface does not support IFF_UNICAST_FLT, then promiscuous mode is already set on the interface, and hence setting promiscuous mode is no update on the interface. Why setting IFLA_GROUP makes a difference I do not yet understand, but I will investigate further.

So far as I can see, keepalived does not use whether an interface is in promiscuous mode or not, so I can't see why a patch is needed in order to get notification of changes in promiscuous mode.

Finally I can't see the relevance of how DPDK behaves to keepalived, so I am confused about the references to DPDK.

louis-6wind:
[DPDK](https://www.dpdk.org/) uses their interface drivers instead of the kernel ones and listens to Netlink messages to set the interface settings correctly.

The issue happens when keepalived is used on a DPDK setup.

This patch triggers the https://github.com/acassen/keepalived/pull/2246 missing netlink up issue but it seems that other situations may trigger the problem as well

pqarmitage:
At last I understand what is happening. If a link supports IFF_UNICAST_FLT the link is not set to promiscuous mode when a macvlan is created on it, and so no netlink message is sent for the base interface when the MACVLAN is created, since nothing has changed.

If a link does not support IFF_UNICAST_FLT then the link is set to promiscuous mode when a macvlan is created on it, but the kernel does not send a netlink message for the base link with the updated promiscuity count (IFLA_PROMISCUITY).

When a netlink RTM_NEWLINK message is sent which includes IFLA_GROUP, the kernel updates the group of the interface (it doesn't check whether it has changed or not) and sets `status |= do_SETLINK_NOTIFY`, which causes a broadcast netlink message to be sent with the "updated" link configuration, which is seen when running `ip -d monitor`. This then allows daemons that monitor the state of interfaces to update the promiscuity information for the base interface.

However, I still don't understand how this relates to DPDK since there is no reference to IFLA_PROMISCUITY in the DPDK code.

pqarmitage:
@louis-6wind I have updated your patch with the following changes:

1. added a configuration option to send the netlink messages for the base interface (specify `netlink_notify_msg` on the `use_vmac` line);
2. if sending the netlink message is enabled, also force a netlink message for the base interface when the VMAC is deleted;
3. update the stored group of an interface if a netlink update message is received for a link;
4. don't set IFLA_IFNAME when updating the group, since the interface name is not being changed;

The reason for 1. above is that there is a window where another process may change the group of an interface, and then we change the group back to the original group before we receive the netlink message updating the group. For this reason, we only want to do the group update if it is necessary.

Commit edf6cd5 adds the above functionality (it also needs commit 14359ca to avoid a build failure when building with `--disable-vmac`).

@louis-6wind many thanks for identifying the issue and providing the original patch.

PR #2265, opened 2023-03-29
vrrp: fix ipv6 vrrp in fault state because no ipv4 address

louis-6wind:
Setting an IPv6 VRRP virtual address on an interface that has no IPv4 address results in a persistent FAULT state:

> Mar 21 11:23:25 ubuntu2204 Keepalived_vrrp[95135]: (vrrp) entering FAULT state (no IPv4 address for interface)

Virtual addresses are not correctly parsed in alloc_vrrp_vip(). vrrp_complete_instance() considers that the IPv4 address family is set by default if no virtual addresses are detected.

>         /* Default to IPv4. This can only happen if no VIPs are specified. */
>         if (vrrp->family == AF_UNSPEC)
>                 vrrp->family = AF_INET;

Fix the virtual address parsing.

Fixes: 516032ec39 ("config parser: Fix segfault caused by extra '}' and other parser fixes")

### DIFF
--- a/keepalived/vrrp/vrrp_data.c
+++ b/keepalived/vrrp/vrrp_data.c
@@ -1115,16 +1115,14 @@ alloc_vrrp_vip(const vector_t *strvec)
 	if (!(new_ipaddr = alloc_ipaddress(strvec, false)))
 		return;
 
-	if (last_ipaddr) {
-		address_family = IP_FAMILY(new_ipaddr);
+	address_family = IP_FAMILY(new_ipaddr);
 
-		if (current_vrrp->family == AF_UNSPEC)
-			current_vrrp->family = address_family;
-		else if (address_family != current_vrrp->family) {
-			report_config_error(CONFIG_GENERAL_ERROR, "(%s): address family must match VRRP instance [%s] - ignoring", current_vrrp->iname, strvec_slot(strvec, 0));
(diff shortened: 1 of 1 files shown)


louis-6wind:
The fix causes a regression with the following scenario:

# Apply the following commands

```
ip -6 addr add  fd00:100::3/64 dev ens4
ip addr add 10.0.0.1/24 dev ens4

ip link set dev ens4 down
killall keepalived
sleep 2
cat>/etc/keepalived/keepalived.conf <<\EOF
global_defs
{
router_id router
dynamic_interfaces
}

vrrp_sync_group group15 {
group {
vrrp
}
}

vrrp_instance vrrp {
version 3
state BACKUP
interface ens4

use_vmac vrrp


garp_master_delay 5

virtual_router_id 15

priority 200
advert_int 1.0

virtual_ipaddress {
fd00:100::1/64
}

preempt_delay 0
}
EOF
keepalived -D
sleep 5
ip link set ens4 up
journalctl _PID=$(pgrep keepalived | tail -n1)
```

pqarmitage:
I agree with this patch, although it also needs last_ipaddr to be removed since it is no longer used.

I have merged this patch (commit bb9f0f4) and added a further commit to remove last_ipaddr (9a7933b).

Many thanks for your help, once again.

PR #2266, opened 2023-03-29
vrrp: restore the vmac ipv6 link-local after flapping

louis-6wind:
The user is not supposed to shutdown a vmac interface created by
keepalived. However, it can mistakenly happen. When the link is
re-established, the link-local has disappear (the kernel removes all
IPv6 addresses on link down except if keep_addr_on_down sysctl is on)
and sending VRRP packet is no nore possible.

Restore the IPv6 Link-Local after a VMAC interface flapping.

Note that the IPv6 Virtual Addresses are also removed on link down which
is the desired behavior. Enabling keep_addr_on_down sysctl would keep
the link-local without this patch but would break this behavior.

### DIFF
--- a/keepalived/core/keepalived_netlink.c
+++ b/keepalived/core/keepalived_netlink.c
@@ -1589,6 +1589,10 @@ process_if_status_change(interface_t *ifp)
 static void
 process_interface_flags_change(interface_t *ifp, unsigned ifi_flags)
 {
+#ifdef _HAVE_VRRP_VMAC_
+	tracking_obj_t *top;
+	vrrp_t *vrrp;
+#endif /* _HAVE_VRRP_VMAC_ */
 	bool now_up = FLAGS_UP(ifi_flags);
 
 	ifp->ifi_flags = ifi_flags;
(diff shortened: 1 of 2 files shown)


pqarmitage:
@louis-6wind Many thanks for identifying the problem and providing the patch.

I have had to make some minor modifications to the patch for the following reasons:
1. On link down on a IPv4 VMAC it would attempt to add an IPv6 link local address
2. On link up on an IPv4 VMAC it would attempt to add an IPv6 link local address
3. Clearing ifp->sin6_addr on link down was not the correct way to do it; that should only be cleared on receipt of the relevant RTM_DELADDR netlink message
4. Don't call down_interface() on receipt of the RTM_DELADDR for the IPv6 link local address if the interface is already down (this is instead of 3 above).

The modified patch is merged as commit [74b203b](https://github.com/acassen/keepalived/pull/2272/commits/74b203b70ce11e7b36dbb88cbccd5405106de0cb), with due reference to you.

Once again, many thanks for your help with this.

PR #2276, opened 2023-04-03
vrrp: do not replace the ipv6 link-local by the same address

louis-6wind:
Do not replace the IPv6 Link-Local address by the same address.


pqarmitage:
Many thanks for the patch.

After the patch is applied, calling replace_link_local_address() now cause two checks that the old and new addresses are different. I have added commit [6e2acc4](https://github.com/acassen/keepalived/pull/2279/commits/6e2acc4a3d7d8473c4056dfeec55690c7e4d705c) to remove the duplicate check.

PR #2277, opened 2023-04-03
vrrp: fix persistent FAULT state with use_vmac on a down base interface

louis-6wind:
If the base interface of VMAC is down at startup, the VRRP instance will remain on FAULT state even if the base interface is set up.

Since the base interface is down, the VMAC operational state is also down and there is no operational IPv6 address. keepalived increments vrrp->num_script_if_fault. When the interface comes up, the link-local IPv6 address is set but vrrp->num_script_if_fault is not decremented. keepalived considers that there is still a remaining fault and keeps the VRRP instance in FAULT state.

Increment vrrp->num_script_if_fault when an IPv6 appears on the VMAC.

Fixes: 4c7a94a409 ("vrrp: Make VMAC IPv6 link local address mirror parent interface")

### DIFF
--- a/keepalived/core/keepalived_netlink.c
+++ b/keepalived/core/keepalived_netlink.c
@@ -1027,22 +1027,25 @@ netlink_if_address_filter(__attribute__((unused)) struct sockaddr_nl *snl, struc
 							 !__test_bit(VRRP_VMAC_XMITBASE_BIT, &vrrp->flags) &&
 							 ifa->ifa_family == AF_INET6 &&
 							 vrrp->ifp->is_ours) {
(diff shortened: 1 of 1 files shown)


pqarmitage:
@louis-6wind Many thanks for this patch, and many apologies for not merging this sooner. This problem appears to have been resolved by some other commit(s) so I am closing the pull request. If you find that the problem still exists, then we can reopen this, or create a new issue/pull request.

louis-6wind:
Please re-open the pull-request. See https://github.com/acassen/keepalived/issues/2280#issuecomment-2748464722

PR #2281, opened 2023-04-04
vrrp: fix non-existent vmac base interface at reload

louis-6wind:
When an existing use_vmac VRRP instance is updated with a non-existent base interface, the instance goes to FAULT. However, when the base interface is created, the VMAC interface is not updated and the instance remains in FAULT state.

Delete the VMAC interface if the base interface is changed to a non-existent one. The VMAC interface will be re-created when the base interface appears.


louis-6wind:
This fix requires https://github.com/acassen/keepalived/pull/2277 to work properly

pqarmitage:
@louis-6wind Many thanks for this patch and apologies for not dealing with this sooner. This problem appears to have already been resolved for IPv4, and that fix, along with commit 8a7620c resolves the issue for IPv6.

louis-6wind:
I confirm that the issue is no more present

PR #2291, opened 2023-04-14
vrrp: use instance fault flags instead of a counter

louis-6wind:
The num_script_if_fault counter is not very convenient. We don't easily
know what has incremented and decremented the counter. And the counter
could go back to zero for a reason in case of a bug.

Use an instance fault flags 'flags_if_fault' instead of a counter. Keep
the a counter 'num_track_fault' for tracked object.


pqarmitage:
@louis-6wind  After far too long I have a last merged this pull request. There were one or two minor corrections needed, and I made a few enhancements, in particular commit 16d045b (use a fault flag to indicate if num_track_fault is non-zero since it simplifies the code) and 7007467 which adds (as a configure option) checking that a flag is not already set in `down_instance()` and not already clear in `try_up_instance()`.

Using these flags has quickly identified a couple of bugs which are now resolved by commits a862103 and 984ac06.

Many thanks for this patch.

louis-6wind:
Thank you for merging the patch

PR #2297, opened 2023-04-24
Update scheduling_algorithms.rst

ianchen0119:


pqarmitage:
Commit 4aaa7df add documentation for MH and TWOS schedulers.

PR #2363, opened 2023-12-29
lvs: if lost misc check child register checker agagin

lewislone:
issue: misc check_child_thread timeout and remove child_pid form rb_data, timeout callback of check_child_thread is not be called, if at this time misc script done and exit, and child termination process will do nothing because child_pid was remove form rb_data. in this case timeout callback will not register checker again, the checker will lost.

fix: if lost misc check child register checker again


pqarmitage:
This commit was included in pull request #2362, albeit the commit message was slightly different.

PR #2388, opened 2024-02-22
vmac: fix creation / deletion VMAC issue due to race conditions

louis-6wind:
Fix two race conditions issue where the VMAC interfaces are not created or deleted following the apparition or disparition of their link interface.


louis-6wind:
@pqarmitage Please review this

pqarmitage:
@louis-6wind I was looking at these commits yesterday and trying to work out what was happening.

I think commit 984ac06 may have improved the situation, since it now doesn't delete any VMACs until it has worked through all the VRRP instances tracking a deleted interface, and only deletes the VMACs once the internal data structures have all been updated. I would be grateful for your thoughts as to whether commit 984ac06 affects any of your changes.

I am currently trying to work my way through all the outstanding pull requests, either to merge them, review them, or explain why they cannot be merged.

I note that you have been collaborating with Hari Sasank (@harisasank) and I am confused by pull requests #2565 and #2556, which appear to be addressing the same issue, although I note that #2565 has been closed now. The original commit (1960568) appears to have disappeared, and we now have three new commits (I haven't had time to look at these yet, since I was trying to work out what was happening between #2556 and #2565).

PR #2397, opened 2024-03-29
Fix lack macro for find_rttables_group()

jimying:
lack macro HAVE_DECL_FRA_SUPPRESS_IFGROUP,
which is similar with get_rttables_group().

This fix build error when macro HAVE_DECL_FRA_SUPPRESS_IFGROUP=0  (on centos 7  kernel 3.10.0)
```
rttables.c:328:31: error: use of undeclared identifier 'rt_groups'
  328 |         return find_entry(name, id, &rt_groups, RT_GROUPS_FILE, NULL, INT32_MAX);
      |                                      ^
rttables.c:328:42: error: use of undeclared identifier 'RT_GROUPS_FILE'
  328 |         return find_entry(name, id, &rt_groups, RT_GROUPS_FILE, NULL, INT32_MAX);
```

### DIFF
--- a/keepalived/vrrp/vrrp_parser.c
+++ b/keepalived/vrrp/vrrp_parser.c
@@ -571,6 +571,7 @@ vrrp_vmac_handler(const vector_t *strvec)
 			continue;
 		}
 
+#if HAVE_DECL_FRA_SUPPRESS_IFGROUP
 		if (!strcmp(strvec_slot(strvec, i), "group")) {
 			uint32_t group;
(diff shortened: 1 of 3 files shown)


pqarmitage:
@jimying Many thanks for this patch, which I have now merged. There was one slight problem with it, since it would not build if SNMP support was disabled, but commit f035191 resolves that.

pqarmitage:
@jimying No, it is not needed to modify get_rttables_group(). get_rttables_group() is only used for responding to SNMP requests, whereas find_rttables_group() is used for ip rules and for vmac and ipvlan config parsing.

PR #2411, opened 2024-05-06
vrrp: remove logging on status output

feckert:
I am the [maintainer](https://github.com/openwrt/packages/blob/master/net/keepalived/Makefile#L22) of the keepalived packages in OpenWrt.
There is also a [LuCI](https://github.com/openwrt/luci/tree/master/applications/luci-app-keepalived) frontend that queries the status of keepalived.

To get the status of keepalived, the json signal handling is used.

The problem now is, that when the LuCI status page is  open in the browser, this removed message in the PR is show every 3 second on status query. In my opinion, this floods the log and is not necessary.


pqarmitage:
@feckert Many thanks for you patch. I guess the log message was a hang over from when the JSON functionality developed.

Issue #2499, opened 2024-11-07
Docker build failing with v2.3.2


rmartin16:
**Describe the bug**
Building on alpine in Docker fails:
```
7.116 ./configure: line 6608: syntax error: unexpected redirection
```

[Build failure log in CI
](https://github.com/rmartin16/docker-keepalived/actions/runs/11655690369/job/32573489460#step:7:877)
<details>
<summary>Expand for full log from local x86 machine</summary>

```
❯ git clone https://github.com/acassen/keepalived.git && cd keepalived && ./autogen.sh && ./configure && make && make docker
Cloning into 'keepalived'...
remote: Enumerating objects: 44329, done.
remote: Counting objects: 100% (1480/1480), done.
remote: Compressing objects: 100% (519/519), done.
remote: Total 44329 (delta 1038), reused 1376 (delta 958), pack-reused 42849 (from 1)
Receiving objects: 100% (44329/44329), 20.05 MiB | 17.50 MiB/s, done.
Resolving deltas: 100% (34567/34567), done.
aclocal: installing 'm4/pkg.m4' from '/usr/share/aclocal/pkg.m4'
configure.ac:423: installing 'build-aux/ar-lib'
configure.ac:413: installing 'build-aux/compile'
configure.ac:101: installing 'build-aux/install-sh'
configure.ac:101: installing 'build-aux/missing'
keepalived/Makefile.am: installing 'build-aux/depcomp'
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a race-free mkdir -p... /usr/bin/mkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking whether make supports nested variables... (cached) yes
checking for pkg-config... /usr/bin/pkg-config
checking pkg-config is at least version 0.9.0... yes
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether the compiler supports GNU C... yes
checking whether gcc accepts -g... yes
checking for gcc option to enable C11 features... none needed
checking whether gcc understands -c and -o together... yes
checking whether make supports the include directive... yes (GNU style)
checking dependency style of gcc... gcc3
checking whether make sets $(MAKE)... (cached) yes
checking for ranlib... ranlib
[...]

Issue #2384, opened 2024-02-18
keepalived Docker Image Build Failed


opencmit2:
**Describe the bug**

keepalived Docker Image Build Failed

**Log entries**
```
make  dist-gzip am__post_remove_distdir='@:'
make[1]: Entering directory `/root/keepalived'
  EDIT     README
if test -d "keepalived-2.2.8"; then find "keepalived-2.2.8" -type d ! -perm -200 -exec chmod u+w {} ';' && rm -rf "keepalived-2.2.8" || { sleep 5 && rm -rf "keepalived-2.2.8"; }; else :; fi
test -d "keepalived-2.2.8" || mkdir "keepalived-2.2.8"
 (cd lib && make  top_distdir=../keepalived-2.2.8 distdir=../keepalived-2.2.8/lib \
     am__remove_distdir=: am__skip_length_check=: am__skip_mode_fix=: distdir)
make[2]: Entering directory `/root/keepalived/lib'
make[2]: Leaving directory `/root/keepalived/lib'
 (cd keepalived && make  top_distdir=../keepalived-2.2.8 distdir=../keepalived-2.2.8/keepalived \
     am__remove_distdir=: am__skip_length_check=: am__skip_mode_fix=: distdir)
make[2]: Entering directory `/root/keepalived/keepalived'
 (cd core && make  top_distdir=../../keepalived-2.2.8 distdir=../../keepalived-2.2.8/keepalived/core \
     am__remove_distdir=: am__skip_length_check=: am__skip_mode_fix=: distdir)
make[3]: Entering directory `/root/keepalived/keepalived/core'
make[3]: Leaving directory `/root/keepalived/keepalived/core'
 (cd vrrp && make  top_distdir=../../keepalived-2.2.8 distdir=../../keepalived-2.2.8/keepalived/vrrp \
     am__remove_distdir=: am__skip_length_check=: am__skip_mode_fix=: distdir)
make[3]: Entering directory `/root/keepalived/keepalived/vrrp'
make[3]: Leaving directory `/root/keepalived/keepalived/vrrp'
 (cd check && make  top_distdir=../../keepalived-2.2.8 distdir=../../keepalived-2.2.8/keepalived/check \
     am__remove_distdir=: am__skip_length_check=: am__skip_mode_fix=: distdir)
make[3]: Entering directory `/root/keepalived/keepalived/check'
make[3]: Leaving directory `/root/keepalived/keepalived/check'
[...]

Issue #835, opened 2018-04-11
 keepalived 1.4.3 build error with linux kernel 4.9.90


michael2009:
esult -iremap/trunk/build_dir/target-x86_64_glibc/keepalived-1.4.3:keepalived-1.4.3 -Wformat -Werror=format-security -D_FORTIFY_SOURCE=1 -Wl,-z,now -Wl,-z,relro  -D_GNU_SOURCE -MT vrrp.o -MD -MP -MF .deps/vrrp.Tpo -c -o vrrp.o vrrp.c
In file included from ./../include/vrrp_if.h:28:0,
                 from ./../include/vrrp_ipaddress.h:39,
                 from ./../include/vrrp_arp.h:33,
                 from vrrp.c:29:
/trunk/staging_dir/toolchain-x86_64_gcc-5.5.0_glibc/include/linux/if.h:79:2: error: redeclaration of enumerator 'IFF_UP'
  IFF_UP    = 1<<0,  /* sysfs */
  ^
/trunk/staging_dir/toolchain-x86_64_gcc-5.5.0_glibc/include/net/if.h:44:5: note: previous definition of 'IFF_UP' was here
     IFF_UP = 0x1,  /* Interface is up.  */
     ^
/trunk/staging_dir/toolchain-x86_64_gcc-5.5.0_glibc/include/linux/if.h:80:2: error: redeclaration of enumerator 'IFF_BROADCAST'
  IFF_BROADCAST   = 1<<1,  /* __volatile__ */
  ^
/trunk/staging_dir/toolchain-x86_64_gcc-5.5.0_glibc/include/net/if.h:46:5: note: previous definition of 'IFF_BROADCAST' was here
     IFF_BROADCAST = 0x2, /* Broadcast address valid.  */
     ^
/trunk/staging_dir/toolchain-x86_64_gcc-5.5.0_glibc/include/linux/if.h:81:2: error: redeclaration of enumerator 'IFF_DEBUG'
  IFF_DEBUG   = 1<<2,  /* sysfs */
  ^
/trunk/staging_dir/toolchain-x86_64_gcc-5.5.0_glibc/include/net/if.h:48:5: note: previous definition of 'IFF_DEBUG' was here
     IFF_DEBUG = 0x4,  /* Turn on debugging.  */
     ^
/trunk/staging_dir/toolchain-x86_64_gcc-5.5.0_glibc/include/linux/if.h:82:2: error: redeclaration of enumerator 'IFF_LOOPBACK'
[...]

PR #2501, opened 2024-11-08
Install linux-headers pkg to build in Docker

rmartin16:
Avoids this error when building on Alpine in Docker:
```
9.540 configure: error: Missing/unusable kernel header file <linux/types.h> - is kernel headers package installed?
```
Of course, this still requires you comment out `#include <linux/if_ether.h>` first.

### DIFF
--- a/Dockerfile.in
+++ b/Dockerfile.in
@@ -31,6 +31,7 @@ RUN apk --no-cache add \
 @NFTABLES_TRUE@	libnftnl-dev \
 	libnl3 \
 	libnl3-dev \
+	linux-headers \
 	make \
 	musl-dev \
 @SNMP_TRUE@	net-snmp-dev \


rmartin16:
hmm...looking at some of the history, it seems like these musl headers have been problematic before:
- https://github.com/acassen/keepalived/issues/835
- https://github.com/acassen/keepalived/issues/2384

I'm not sure I understand all this enough to know what's best....

Furthermore, to your point in #2499 about opening an issue for musl, I did some Google searching and this seems to have been a well-known issue for over ten years. Their recommendation seems to be "don't include both headers"...

At any rate, I cannot build `master` in Docker without this patch. The only other solution I can imagine is to switch the precedence of the headers and defer to netinet instead....but I really don't know...

pqarmitage:
glibc has worked hard over the last few years to resolve the issue of which header files can and cannot be included together, and in what order, and my understanding is that they have solved all the problems. It is a shame that musl seem not to be interesting in doing likewise, although I may be maligning them.

PR #2520, opened 2024-12-18
refactor misc shell script tweaks

xgqt:
some non-critical refactors of BASH/SH scripts

* snap-tools: use sh as interpreter; misc tweaks
    * use sh as interpreter since we already have sh compatibility,
    * sh-compat: only change export to be on 2 lines,
    * set SNAP if somehow unset
    * quote MAGIC and exec call
* goodies: use bash mapfile and array to store found C files

### DIFF
--- a/goodies/check_conditional_tests
+++ b/goodies/check_conditional_tests
@@ -1,9 +1,12 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Get a count of each defined term used in #if directives
 # This can be useful to check for typos in the names
 
-grep "#[ \t]*if" $(find . -name "*.[ch]") | \
+declare -a c_src_files=()
+mapfile -t c_src_files < <(find . -name "*.[ch]")
+
+grep "#[ \t]*if" "${c_src_files[@]}" | \
 	grep -v "_H$" | \
 	sed -e "s/.*://" | \
 	sed -e "s/#ifn*def *//" -e "s/#if  *//" | \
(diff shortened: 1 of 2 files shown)


pqarmitage:
I am not clear of the purpose of some of these changes.

1. What is the purpose of `: "${SNAP:=/}"`? This results in `exec "//usr/sbin/${BINARY}" "$@"` if SNAP was unset, whereas without the change it resulted in `exec "/usr/sbin/${BINARY}" "$@"` which is more obviously correct.
2. What is wrong with `export MAGIC="${SNAP}/usr/share/misc/magic"`?
3. The shebang lines are different in the modifications to the two files you have made. Is there a reason for this?

xgqt:
> What is the purpose of : "${SNAP:=/}"? This results in exec "//usr/sbin/${BINARY}" "$@" if SNAP was unset, whereas without the change it resulted in exec "/usr/sbin/${BINARY}" "$@" which is more obviously correct.

yeah, you are right here, with `set -u` it would be necessary

> What is wrong with export MAGIC="${SNAP}/usr/share/misc/magic"?

some sh shells do not understand inline `export`

> The shebang lines are different in the modifications to the two files you have made. Is there a reason for this?

yes, the snap script already used sh syntax while defaulting to the bash interpreter.
In case os snap script it can be made to be entirely sh-compatible

in case of the helper `check_conditional_tests` it is better to use bash for use with `mapfile` function.
the interpreter here was changed because it is not clear whether bash will be on /usr/bin or maybe /bin between distros. Thus, if we do not know location of a interpreter we use /usr/bin/env because that location is known (just like /bin/sh).

pqarmitage:
Many thanks for the updates and your explanation. I wasn't familiar with bash mapfiles or process substitution, so that is today's learning done!

PR #2541, opened 2025-01-28
core: add O_CLOEXEC flag to pidfiles

sshambar:
Patch to add O_CLOEXEC flag when opening pidfiles.

Commit 2c4cd3b introduced the use of pid files for instance locking, which keeps the pidfile descriptors open.  However, the O_CLOEXEC flag was missing so the open descriptors are passed to all exec'd processes (scripts etc).

I noticed this when adding SELinux permissions for keepalived to use ping as a vrrp_script, and saw the following denial:
```
type=AVC msg=audit(1738090198.716:585): avc:  denied  { write } for  pid=21305 comm="ping" 
  path="/run/vrrp.pid" dev="tmpfs" ino=3178 scontext=system_u:system_r:ping_t:s0 
  tcontext=system_u:object_r:keepalived_var_run_t:s0 tclass=file permissive=0
```

I created a test script to dump the open file descriptors, and found one extra descriptor for /run/vrrp.pid

Ideally, the code that exec's any external program should close all open descriptors (except stdin/out/err which might be optional), but it's also simple enough to just ensure all descriptors have O_CLOEXEC so it's automatic.

### DIFF
--- a/keepalived/core/pidfile.c
+++ b/keepalived/core/pidfile.c
@@ -167,7 +167,7 @@ create_pidfile(pidfile_t *pidf)
 		if (umask_val & (S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH))
 			umask(umask_val & ~(S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH));
 
-		while ((pidf->fd = open(pidf->path, O_NOFOLLOW | O_CREAT | O_WRONLY | O_NONBLOCK, S_IRUSR | S_IWUSR | S_IRGRP | S_IROTH)) == -1 && errno == EINTR);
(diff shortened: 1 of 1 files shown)


pqarmitage:
@sshambar Many thanks for this patch.

I have looked at keepalived/core/pidfile.c and there is a second open() of a pid file, only used when a config reload of keepalived is done and it introduces new config that requires starting a child process that wasn't running before. That open() also needs O_CLOEXEC to be set, so I have merged commit 52b5e73 that adds that.

I am also considering whether, to ensure that there can be no other leaks of open descriptors to child processes, we should, as a matter of policy, always set O_CLOEXEC/SOCK_CLOEXEC when opening a file or socket. I will think about it for a while, and push a patch to do that unless I can think of any reason why not. Any thoughts would be appreciated.

Issue #2555, opened 2025-03-03
add "route-propagation-delay SECONDS" field to keepalived.


harisasank:
I have a usecase where I need to to provide HA between multiple datacenters connected over public internet. I have keepalived running on a publicly accessible endpoint in each of the DCs.
A VRRP master is elected between these nodes (in L3 Unicast mode), such that all the internal services are accessed via the VRRP master.
If the L3 connectivity between these nodes is severed for MDT interval, the backup will take over as master. (as expected). When the HA cluster network is restored, all the routes must be propagated to all peers before a new master is re-elected.
Before the routing protocols converge among all peers, a given peer doesnt receive all messages which can lead it into believing that it is alone, and should take over as master.
This results in a split brain case where where more than one VRRP nodes own the VRRP VIP and we have more than one master. 

startup-delay wont work here because it only applies when keepalived daemon restarts, and preempt delay does nt get applied here, 
because it is only triggered when the node receives a VRRP advertisement of a higher priority then the instance's local priority.
Increasing advertisement intervals increases the master switchover time, and cannot compensate for BGP convergence time (if one is using BGP as we are) which can be in order of several 10s of seconds or even minutes.

In order to avoid this stituation, a new field "route-propagation-delay SECONDS" can be added to the per instance vrrp config, which basically does the following:
1. When a VRRP node moves from FAULT state to BACKUP and eventually to MASTER (after MDT timeout), wait for an additional "route-propagation-delay" time if such a field is set. 
2. During this delay, all the advertisements that are received from master are ignored (just like startup-delay).
3. This will be applied every time the node exits out of FAULT state and does not directly to master. This feature is not supported with initial state master and vrrp strict mode. (like preempt delay)
4. Needless to say, no behavioral change if "route-propagation-delay" is not set in the configuration. 

I can send patches to support this feature if this is useful.

PR #2556, opened 2025-03-03
vrrp: support route propagation delay

harisasank:
When vrrp state moves out of a fault state, we would like to wait for an additional route propagated delay time, before doing a vrrp master election.
During this interval, the packets that are received from other vrrp nodes, which can alter the state of the vrrp instance are dropped. The present nodes, does not pass to participate in leader election which can result in it becoming the master and taking over the VRRP IP.

### DIFF
--- a/keepalived/include/vrrp.h
+++ b/keepalived/include/vrrp.h
@@ -373,7 +373,17 @@ typedef struct _vrrp_t {
 							 * preemption based on higher prio over lower
 							 * prio is allowed.  0 means no delay.
 							 */
+	unsigned long		route_propagation_delay;/* Seconds*TIMER_HZ to wait at startup until
+							 * MDT timer can start.  0 means no delay.
+							 */
 	timeval_t		preempt_time;		/* Time after which preemption can happen */
+	timeval_t		route_propagated_time;	/* Time at which all routes are propagated.
+							 * VRRP packets to this interface are dropped before
+							 * all routes are propagated.
+							 */
+	bool			apply_route_propagation;/* Apply the route propagation delay the next time when
+							 * instance tries to become master.
+							 */
 	int			state;			/* internal state (init/backup/master/fault) */
 #ifdef _WITH_SNMP_VRRP_
 	int			configured_state;	/* the configured state of the instance */
--- a/keepalived/vrrp/vrrp.c
+++ b/keepalived/vrrp/vrrp.c
@@ -1973,6 +1973,16 @@ vrrp_state_leave_fault(vrrp_t * vrrp)
 	/* Set the down timer */
 	vrrp->master_adver_int = vrrp->adver_int;
 	vrrp->ms_down_timer = VRRP_MS_DOWN_TIMER(vrrp);
+	if (vrrp->state == VRRP_STATE_BACK && vrrp->route_propagation_delay > 0)
+	{
+		vrrp->apply_route_propagation = true;
+		vrrp->route_propagated_time = timer_add_long(time_now, vrrp->route_propagation_delay);
+	}
+	if (vrrp->state == VRRP_STATE_FAULT)
+	{
+		vrrp->apply_route_propagation = false;
+		vrrp->route_propagated_time = time_now;
+	}
 	vrrp_init_instance_sands(vrrp);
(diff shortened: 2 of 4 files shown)


harisasank:
I have explained more details about this changeset here: https://github.com/acassen/keepalived/issues/2555

pqarmitage (changes requested):
I am very happy with this patch in principle but there a few issues that need to be resolved first.

Unfortunately the patch as submitted doesn't work properly. Consider a VRRP instance with priority 240 already running as master, and another instance with priority 220 starts up. vrrp->sands is calculated to be route_propagation_delay + MDT in the future, and until that time expires any received adverts are ignored. As soon as the sands timer expires, the lower priority instance transitions to master state (because that is what the expiry of the sands timer means). The lower priority VRRP instance does revert to backup state after it next receives an advert from the priority 240 instance, but the transition to master should not have happened in the first place.

There is a simple fix to this problem, and that is in `vrrp_init_instance_sands()` set `vrrp->route_propagated_time = timer_add_long(time_now, vrrp->route_propagation_delay)`, so that the priority 220 instance then has MDT after the vrrp->route_propagated_time has passed to be able to receive adverts before it decides to transition to master.

There is also a problem with configuration reloads. If an instance is in backup state and its configuration is reloaded, then after the reload the route propagation delay is again applied to the VRRP instance, and of course this is wrong if the old VRRP instance is not waiting for the route_propagated_time to expire. The propagation delay needs to be carried across a reload, more specifically the vrrp->route_propagated_time needs to be recalculated as new vrrp->route_propagated_time = old vrrp->route_propagated_time - old vrrp->route_propagated_delay + new vrrp->route_propagated_delay, but only if old vrrp->route_propagated_time has not passed.

Your implementation makes the vrrp instance enter backup state as normal, but then ignores any received adverts for the configured delay. I think this is the wrong approach, since what you are really wanting to happen is that the vrrp instance remains in the init/fault state after a fault is cleared, or after startup, for the configured delay time, before it then transitions to backup state. 

**Cosmetic**
I think the names of the configuration parameter and the variables are too specific to your purpose - there may be other users who want to use this feature but for other reasons. I think the names should be based on something like "fault_init_exit_delay", so that they are general purpose. Comments need to be adjusted accordingly.

**Minor**
In vrrp_route_propagation_delay_handler, don't clear current_vrrp->route_propagation_delay on error (this is how most of the similar code behaves - preempt_delay currently does clear the setting, but I will change that to make it consistent with the other parameters).

The log entry `Applied vrrp route propagation delay of 10000000 on instance (VI_1)
` needs to be changed to `Applied vrrp route propagation delay of 10.000000 seconds on instance (VI_1)`.

The new variables need to be included in the dump_vrrp() function in vrrp_data.c, and also in the SNMP variables in the KEEPALIVED.MIB.

**Major**
doc/man/man5/keepalived.conf.5.in must be updated to add the new keyword (otherwise no-one else will know it exists).

In `vrrp_dispatcher_read()` the added _return sock->fd_in_ should be _continue_.

The change to `vrrp_dispatcher_read()` adds a timer check every time a packet is received (if this feature is enabled). This is not the way keepalived handles time. A timer thread needs to be added so that when the timer expires, the thread is run and the state is changed accordingly.

I think this is best implemented by adding a timer thread (thread_add_timer()), with the timer being route_propagation_delay, and the vrrp instance remains in the fault state until the timer expires. When the timer expires the vrrp instance then transitions to backup state. If a new fault occurs before the timer expires, the existing timer thread would need to be cancelled. The timer thread would also need to be carried over a reload (I think the way this is best handled is to the vrrp structure to have a pointer to a fault_exit_timer_thread, which is cleared when the timer expires. If the timer still exists when a reload occurs, then a new timer can be created, using the sands value from the old timer thread suitably adjusted for any changes to the configured route_propagation_delay. The pointer to the fault_exit_timer_thread could also be used to cancel the timer if a new fault occurs before the timer expires).

If you would like me to help with making the changes above, or even me complete them, I am quite happy to do so, but I always prefer to let the originator of a patch complete any necessary changes in the first place.

harisasank:
Thank you for your comments. I will send a new patch soon.

PR #2559, opened 2025-03-07
vrrp: handle a reload with no more startup_delay

louis-6wind:
During the vrrp_startup_delay time, if keepalived is reloaded with no more startup_delay, the startup_delay is never timed out and all received adverts is discarded.

The commit causes the startup_delay timer to be reinstated after a reload with no more startup_delay if the timer has not yet expired.

Fixes: 58483b22c8 ("vrrp: Handle a reload before vrrp_delayed_start has expired")


louis-6wind:
Please review this

louis-6wind:
Thank you for the merge @pqarmitage 

I will address your remark on the other pull-request soon.
The work on fault-init-exit-delay is not finished and we will let you know when it is finished

PR #2565, opened 2025-03-19
vrrp: add fault_init_exit_delay to vrrp configuration

harisasank:
When vrrp state moves out of a fault state, we would like to wait for an additional route propagated delay time, before doing a vrrp master election.
During this interval, the packets that are received from other vrrp nodes, which can alter the state of the vrrp instance are dropped. The present nodes, does not pass to participate in leader election which can result in it becoming the master and taking over the VRRP IP.

Addressing comments from PR 2556


harisasank:
This PR is the same feature as 2556.

Issue #2280, opened 2023-04-04
VRRPv3 IPv6: persistent FAULT state when updating VMAC with a non-existent VMAC base interface


louis-6wind:
**Describe the bug**
VRRPv3 IPv6 remains in a persistent FAULT state after setting an non-existent VMAC base interface. When the base interface appears, the problem is not solved. Base interface is never updated on the vmac.

**To Reproduce**

On a standalone VM machine

```
# Apply the following commands
killall keepalived
sleep 1
ip link set iface5 down
ip link set iface5 name ens5
ip link set ens5 down

ip -6 addr add  fd00:100::3/64 dev ens4
ip addr add 10.0.0.1/24 dev ens4
ip link set dev ens4 up
sleep 2
cat>/etc/keepalived/keepalived.conf <<\EOF
global_defs
{
router_id router
dynamic_interfaces
}

vrrp_sync_group group15 {
group {
vrrp
}
}

vrrp_instance vrrp {
version 3
state BACKUP
interface ens4

use_vmac vrrp


garp_master_delay 5

virtual_router_id 15

priority 200
advert_int 1.0

virtual_ipaddress {
fd00:100::1/64
}

preempt_delay 0
}
EOF
keepalived -D
sleep 5
journalctl _PID=$(pgrep keepalived | tail -n1) | cat
sed -e 's|ens4|iface5|g' -i /etc/keepalived/keepalived.conf
killall -s SIGHUP keepalived
sleep 4
ip -br l
journalctl _PID=$(pgrep keepalived | tail -n1) | cat
ip link set ens5 name iface5 up
sleep 4
journalctl _PID=$(pgrep keepalived | tail -n1) | cat
ip -br l

```

**Expected behavior**

Instance in MASTER state. VMAC updated with the correct base interface

**Keepalived version**
```
Keepalived v2.2.7 (04/02,2023), git commit v2.2.7-148-g58be65ee

Copyright(C) 2001-2023 Alexandre Cassen, <acassen@gmail.com>

Built with kernel headers for Linux 4.15.18
Running on Linux 5.4.0-135-generic #152~18.04.2-Ubuntu SMP Tue Nov 29 08:23:49 UTC 2022
Distro: Ubuntu 18.04.1 LTS

[...]

PR #2574, opened 2025-03-24
vrrp: fix persistent FAULT state with use_vmac on a down base interface

louis-6wind:
If the base interface of VMAC is down at startup, the VRRP instance will remain on FAULT state even if the base interface is set up.

keepalived uses a vrrp->num_script_if_fault fault counter that is incremented when there is a new fault and decremented when the fault is recovered. When the counter equals to zero, it considers there is no fault. try_up_instance() is used to decrease the counter. Since 4c7a94a409 ("vrrp: Make VMAC IPv6 link local address mirror parent interface"), it was not called anymore when a IPv6 link-local was back on the VMAC interface.

Increment vrrp->num_script_if_fault when an IPv6 appears on the VMAC.

Fixes: 4c7a94a409 ("vrrp: Make VMAC IPv6 link local address mirror parent interface")

### DIFF
--- a/keepalived/core/keepalived_netlink.c
+++ b/keepalived/core/keepalived_netlink.c
@@ -1032,22 +1032,24 @@ netlink_if_address_filter(__attribute__((unused)) struct sockaddr_nl *snl, struc
 							 ifa->ifa_family == AF_INET6 &&
 							 vrrp->ifp->is_ours) {
 							if (vrrp->saddr.ss_family == AF_UNSPEC) {
-								inet_ip6tosockaddr(addr.in6, &vrrp->saddr);
+								if (
 #if 0
-								if (IN6_IS_ADDR_UNSPECIFIED(&vrrp->ifp->sin6_addr)) {
 									/* This should never happen with the current code since we always
 									 * create a link local address on the VMAC interface.
 									 * However, if in future it is decided not to automatically create
 									 * a link local address on the VMAC interface if the parent interface
 									 * does not have one, then we will need the following code
 									 */
-									if (add_link_local_address(vrrp->ifp, addr.in6) &&
-									    vrrp->flags_if_fault &&
-									    (!__test_bit(VRRP_FLAG_SADDR_FROM_CONFIG, &vrrp->flags) || is_tracking_saddr))
-										try_up_instance(vrrp, false, VRRP_FAULT_FL_NO_SOURCE_IP);
-								} else
+										add_link_local_address(vrrp->ifp, addr.in6) &&
 #endif
-								reset_link_local_address(&vrrp->ifp->sin6_addr, vrrp);
+									    vrrp->flags_if_fault &&
+									    (!__test_bit(VRRP_FLAG_SADDR_FROM_CONFIG, &vrrp->flags) || is_tracking_saddr)) {
+										inet_ip6tosockaddr(addr.in6, &vrrp->saddr);
+										try_up_instance(vrrp, false, VRRP_FAULT_FL_NO_SOURCE_IP);
+								} else {
+									inet_ip6tosockaddr(addr.in6, &vrrp->saddr);
+									reset_link_local_address(&vrrp->ifp->sin6_addr, vrrp);
+								}
 							}
 						}
 #endif


louis-6wind:
> @louis-6wind Is this pull request now superseded by the commits merged yesterday? So far as I am aware, yesterday's commits resolved the issue of a VMAC's base interface being down when keepalived starts up.

I have tested on the latest master after your merge and I still have the #2280 issue. See https://github.com/acassen/keepalived/issues/2280#issuecomment-2748464722 I confirm that I have tested against 087ad4aacdf2349349d7071689fa46d9ab7055f2

This pull request is a rebase of #2277

pqarmitage:
@louis-6wind I can now see that a patch along these lines is needed. I was confused by your reference to `vrrp->num_script_if_fault` above, since your commit a0e8803 removed `num_script_if_fault`.

I have some concern, possibly unfounded, that simply checking `vrrp->flags_if_fault` (!= 0) and then clearing VRRP_FAULT_FL_NO_SOURCE_IP may not be right if `vrrp->flags_if_fault` is non-zero, but the VRRP_FAULT_FL_NO_SOURCE_IP flag is not set. On the one hand, we have already tested `vrrp->saddr.ss_family == AF_UNSPEC`, so we know that VRRP instance doesn't have a source IP, which means that the VRRP_FAULT_FL_NO_SOURCE_IP must (?should) be set, but in that case why test if `vrrp->flags_if_fault` is non-zero, since we know VRRP_FAULT_FL_NO_SOURCE_IP is set.

Do we also need to check that the new address is a link local address?

After applying your patch, the code is clearly improved, since the vrrp instance transitions to backup and then master state. However, once keepalived starts sending adverts I am getting `(vrrp): send advert error 22 (Invalid argument)` errors once per second, which means that there is a problem with the send socket (`vrrp->sockets->fd_out`). It appears that the vrrp VMAC is not being deleted after the first config reload, when the base interface is changed from ens4 to the non-existent iface6.

I have added the following patch:
```
diff --git a/keepalived/vrrp/vrrp.c b/keepalived/vrrp/vrrp.c
index 07c49f30..95df9da6 100644
--- a/keepalived/vrrp/vrrp.c
+++ b/keepalived/vrrp/vrrp.c
@@ -3868,7 +3869,8 @@ vrrp_complete_instance(vrrp_t * vrrp)
                                /* coverity[var_deref_model] - vrrp->configured_ifp is not NULL for VMAC */
                                netlink_link_add_vmac(vrrp, old_interface);
                        }
-               }
+               } else if (old_interface)
+                       netlink_link_del_vmac(vrrp);
 
                if (vrrp->ifp->base_ifp->ifindex &&
                    !__test_bit(VRRP_VMAC_UP_BIT, &vrrp->flags) &&
```
which deletes the VMAC after the reload when the base interface no longer exists. This means that after renaming ens5 to iface5 the VMAC gets created again; however the vrrp instance remains in fault state since keepalived hasn't detected the IPv6 addresses on iface5 (downing iface5 and then upping it again does bring the vrrp instance back up - this is thanks to the move to using fault flags). I think what we are going to need to do is, when we detect an interface being renamed, move the IPv4 and IPv6 addresses across to the new interface (since keepalived uses the ifname as the primary index for an interface, rather than the ifindex), and then consider whether that allows any VRRP instances using that interface to come out of fault state.

I have simplified your patch to the following:
```
diff --git a/keepalived/core/keepalived_netlink.c b/keepalived/core/keepalived_netlink.c
index 6f259a8a..bad0915b 100644
--- a/keepalived/core/keepalived_netlink.c
+++ b/keepalived/core/keepalived_netlink.c
@@ -1030,25 +1030,13 @@ netlink_if_address_filter(__attribute__((unused)) struct sockaddr_nl *snl, struc
                                                         IS_MAC_IP_VLAN(vrrp->ifp) &&
                                                         !__test_bit(VRRP_VMAC_XMITBASE_BIT, &vrrp->flags) &&
                                                         ifa->ifa_family == AF_INET6 &&
-                                                        vrrp->ifp->is_ours) {
-                                                       if (vrrp->saddr.ss_family == AF_UNSPEC) {
+                                                        vrrp->ifp->is_ours &&
+                                                        vrrp->saddr.ss_family == AF_UNSPEC) {
                                                                inet_ip6tosockaddr(addr.in6, &vrrp->saddr);
-#if 0
-                                                               if (IN6_IS_ADDR_UNSPECIFIED(&vrrp->ifp->sin6_addr)) {
-                                                                       /* This should never happen with the current code since we always
-                                                                        * create a link local address on the VMAC interface.
-                                                                        * However, if in future it is decided not to automatically create
-                                                                        * a link local address on the VMAC interface if the parent interface
-                                                                        * does not have one, then we will need the following code
-                                                                        */
-                                                                       if (add_link_local_address(vrrp->ifp, addr.in6) &&
-                                                                           vrrp->flags_if_fault &&
-                                                                           (!__test_bit(VRRP_FLAG_SADDR_FROM_CONFIG, &vrrp->flags) || is_tracking_saddr))
-                                                                               try_up_instance(vrrp, false, VRRP_FAULT_FL_NO_SOURCE_IP);
-                                                               } else
-#endif
-                                                               reset_link_local_address(&vrrp->ifp->sin6_addr, vrrp);
-                                                       }
+                                                               if (!__test_bit(VRRP_FLAG_SADDR_FROM_CONFIG, &vrrp->flags) || is_tracking_saddr)
+                                                                       try_up_instance(vrrp, false, VRRP_FAULT_FL_NO_SOURCE_IP);
+                                                               else
+                                                                       reset_link_local_address(&vrrp->ifp->sin6_addr, vrrp);
                                                }
 #endif
                                        }
```
and I think there are still some further simplifications since the if block (line 1008) and the else block (1027) have some common tests.

I have quickly thrown together a patch to move the IP addresses from ens5 to iface5 when the interface rename occurs, and that seems to resolve the problem (I have done it in a rush since I have to go out for the next few hours, but I will look into it further when I return). The additional patch is:
```
diff --git a/keepalived/core/keepalived_netlink.c b/keepalived/core/keepalived_netlink.c
index 6f259a8a..d7a816b4 100644
--- a/keepalived/core/keepalived_netlink.c
+++ b/keepalived/core/keepalived_netlink.c
@@ -2111,6 +2113,11 @@ netlink_link_filter(__attribute__((unused)) struct sockaddr_nl *snl, struct nlms
        list_head_t sav_e_list;
        garp_delay_t *sav_garp_delay;
        bool immediate_change = false;
+       struct in_addr sin_addr;
+       struct in6_addr sin6_addr;
+       list_head_t sin_addr_l;
+       list_head_t sin6_addr_l;
+       bool copy_addr = false;;
 
        if (!(h->nlmsg_type == RTM_NEWLINK || h->nlmsg_type == RTM_DELLINK))
                return 0;
@@ -2221,6 +2228,13 @@ netlink_link_filter(__attribute__((unused)) struct sockaddr_nl *snl, struct nlms
                        if (strcmp(ifp->ifname, name)) {
                                /* The name can change, so handle that here */
                                log_message(LOG_INFO, "Interface name has changed from %s to %s", ifp->ifname, name);
+                               sin_addr = ifp->sin_addr;
+                               sin6_addr = ifp->sin6_addr;
+                               sin_addr_l = ifp->sin_addr_l;
+                               sin6_addr_l = ifp->sin6_addr_l;
+                               INIT_LIST_HEAD(&ifp->sin_addr_l);
+                               INIT_LIST_HEAD(&ifp->sin6_addr_l);
+                               copy_addr = true;
 #ifndef _ONE_PROCESS_DEBUG_
                                if (prog_type != PROG_TYPE_VRRP) {
                                        ifp->ifi_flags = 0;
@@ -2316,6 +2330,12 @@ netlink_link_filter(__attribute__((unused)) struct sockaddr_nl *snl, struct nlms
                        INIT_LIST_HEAD(&ifp->sin6_addr_l);
                        list_copy(&ifp->tracking_vrrp, &sav_tracking_vrrp);
                        ifp->garp_delay = sav_garp_delay;
+                       if (copy_addr) {
+                               ifp->sin_addr = sin_addr;
+                               ifp->sin6_addr = sin6_addr;
+                               ifp->sin_addr_l = sin_addr_l;
+                               ifp->sin6_addr_l = sin6_addr_l;
+                       }
 
                        if (!netlink_if_link_populate(ifp, tb, ifi))
                                return -1;
```

Any thoughts or comments are welcome. I should be back by 15:00 UTC and will continue looking at this then.

pqarmitage:
@louis-6wind I have some updated patches re this pull request which I will merge in the next hour.

pqarmitage:
@louis-6wind  I have merged commits f7c0805 and 09e1990 which between them I believe fixes this issue. The changes were rather more wide ranging than your original patch, but f7c0805 is substantially based on your patch.

If you find any further issues when testing, then further patches or issues will be gratefully received. If the latest patches do resolve the issue for you, then I think this pull request could be closed.

Many thanks for your help with this, and for the original patch.

louis-6wind:
@pqarmitage thank you for your help. #2280 issue is now fixed. Great!!!
Sorry, I had no time this week to address your comments on this

PR #2614, opened 2025-05-06
Fix build error when HAVE_CLOSE_RANGE not defined

jimying:
unistd.h should't inside macro HAVE_CLOSE_RANGE

### DIFF
--- a/lib/notify.c
+++ b/lib/notify.c
@@ -27,10 +27,10 @@
 #ifndef _GNU_SOURCE
 #define _GNU_SOURCE         /* for close_range() */
 #endif
-#include <unistd.h>
 #endif
 #include <linux/close_range.h>
 #endif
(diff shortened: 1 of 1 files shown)


pqarmitage:
@jimying I don't get a build error when HAVE_CLOSE_RANGE is not defined, so it would be helpful if you could post the compiler error message(s) you are getting, and also what distro and version you are using.

pqarmitage:
Commits [524d753](https://github.com/acassen/keepalived/pull/2615/commits/524d7533fd44088ba4c3531751eab28e966fb02a) and [8ccb425](https://github.com/acassen/keepalived/pull/2615/commits/8ccb425d5f15f83755a4d959fa2821ed1564cefe) fix some further compilation problems found on RHEL 7.

PR #2633, opened 2025-07-01
Fix address sanitizer error in find_definition

WhiteWLf-dev:
The error is that an element is being accessed beyond the right limit of the array allocated on the heap. That is, first we find out the length of the string, and only then we access its elements by indexes, checking the range to which these indexes belong.


pqarmitage:
Unfortunately the patch in this pull request breaks the parser and so cannot be merged.

PR #2638, opened 2025-07-01
Fix UB sanitizer error in add_seq

WhiteWLf-dev:
The error is that a null ptr is returned from strchr(), which is subsequently used in a subsequent expression.


pqarmitage:
The following configuration causes this issue:
```
~SEQ(fred, 1, 3, 2 bert
```

pqarmitage:
Unfortunately the patch in this pull request breaks the parser.

PR #2644, opened 2025-07-29
ipvs: Remove duplicate virtual servers during config validation

lxchx:
The current IPVS implementation allows duplicate virtual server (VS) configurations. While IPVS ensures only the first VS takes effect, the presence of redundant VS entries in the keepalived state can be confusing. Furthermore, checkers for these remaining duplicate VS still perform unnecessary work.

This commit actively detects and removes duplicate virtual servers during configuration validation. To maintain compatibility with existing configurations that might implicitly rely on this behavior, this change preserves the first encountered VS.

### DIFF
--- a/keepalived/check/check_data.c
+++ b/keepalived/check/check_data.c
@@ -1070,15 +1070,30 @@ check_check_script_security(void)
 bool
 validate_check_config(void)
 {
-	virtual_server_t *vs, *vs_tmp;
+	virtual_server_t *vs, *vs_tmp, *vs1;
 	virtual_server_group_entry_t *vsge;
 	real_server_t *rs, *rs_tmp, *rs1;
 	checker_t *checker;
 	unsigned weight_sum;
-	bool rs_removed;
+	bool vs_removed, rs_removed;
 
 	using_ha_suspend = false;
 	list_for_each_entry_safe(vs, vs_tmp, &check_data->vs, e_list) {
+		/* Check the virual server is not a duplicate of any vs earlier in the list */
+		vs_removed = false;
+		list_for_each_entry(vs1, &check_data->vs, e_list) {
+			if (vs == vs1)
+				break;
+			if (vs_iseq(vs, vs1)) {
+				report_config_error(CONFIG_GENERAL_ERROR, "Virtual server %s is duplicated - removing second vs and all its related rs and checker entities.", FMT_VS(vs));
+				free_vs(vs);
+				vs_removed = true;
+				break;
+			}
+		}
+		if (vs_removed)
+			continue;
+
 		/* Ensure that ha_suspend is not set for any virtual server using fwmarks */
 		if (vs->ha_suspend &&
 		    (vs->vfwmark || (vs->vsg && !list_empty(&vs->vsg->vfwmark)))) {
--- a/keepalived/check/ipwrapper.c
+++ b/keepalived/check/ipwrapper.c
@@ -42,7 +42,7 @@
 #include "check_data.h"
 #endif
 
-static bool __attribute((pure))
+bool __attribute((pure))
 vs_iseq(const virtual_server_t *vs_a, const virtual_server_t *vs_b)
 {
 	if (!vs_a->vsgname != !vs_b->vsgname)
--- a/keepalived/include/ipwrapper.h
+++ b/keepalived/include/ipwrapper.h
@@ -48,6 +48,7 @@ rs_iseq(const real_server_t *rs_a, const real_server_t *rs_b)
 {
 	return sockstorage_equal(&rs_a->addr, &rs_b->addr);
 }
+extern bool __attribute__((pure)) vs_iseq(const virtual_server_t *, const virtual_server_t *);
 
 /* prototypes */
 extern void update_svr_wgt(int64_t, virtual_server_t *, real_server_t *, bool);


lxchx:
This configuration might be helpful:
```
virtual_server 192.168.200.1 80 {
    delay_loop 6
    lb_algo rr
    lb_kind DR
    protocol TCP

    real_server 10.0.0.1 80 {
        weight 1
        MISC_CHECK {
            misc_path "/etc/keepalived/checks/check_rr_rs1.sh"
            misc_timeout 5
            user root
        }
    }

    real_server 10.0.0.2 80 {
        weight 1
        MISC_CHECK {
            misc_path "/etc/keepalived/checks/check_rr_rs2.sh"
            misc_timeout 5
            user root
        }
    }
}
virtual_server 192.168.200.1 80 {
    delay_loop 6
    lb_algo wrr
    lb_kind DR
    protocol TCP

    real_server 10.0.0.1 80 {
        weight 5
        MISC_CHECK {
            misc_path "/etc/keepalived/checks/check_wrr_rs1.sh"
            misc_timeout 5
            user root
        }
    }

    real_server 10.0.0.2 80 {
        weight 1
        MISC_CHECK {
            misc_path "/etc/keepalived/checks/check_wrr_rs2.sh"
            misc_timeout 5
            user root
        }
    }
}
```
---
When the --dump-conf parameter is enabled, Keepalived retains two duplicate virtual server (VS) configurations. Furthermore, all four misc_checker instances remain active, even though the real servers associated with check_wrr_rs1.sh and check_wrr_rs2.sh are effectively shadowed or superseded.

pqarmitage:
@lxchx If you change the real server addresses in the second virtual server block to be, say, 10.0.1.1 and 10.0.1.2, you will find that keepalived then ends up with 4 real servers (10.0.0.1, 10.0.0.2, 10.0.1.1 and 10.0.1.2) all configured on the first virtual server (i.e. with the `lb_algo` rr). It gets even worse if you have quorums and sorry_servers configured, and checkers start failing.

It is interesting that you have taken completely the opposite approach from pull request #2691 which wants to allow the real servers to continue to  operate, but for them to be added to the first virtual server (there are problems with that pull request that I have identified in a comment there).

Certainly at the moment things don't work correctly. For example, if the quorum of the first VS is lost, the sorry server of the of that will be added, but any RS configured on the second VS will remain active. If the quorum of the second VS is later lost, then it's sorry server will be added, and the VS will end up with two sorry servers. It gets worse; if there is a config reload when the first VS is removed from the config, the second VS (and its RSs) get completely lost - it was that that pull request #2691 was attempting to resolve.

What I propose to do is to use your patch to identify duplicate VSs, but instead of deleting any duplicates, just move all the RSs on duplicate VSs to be RSs of the first VS. That will give us the closest behaviour to what happens currently, albeit by accident. Any VS specific configuration of duplicate VSs (e.g. lb_algo, lb_kind) will be ignored, the sorry server of duplicate VSs will be ignored (a change from current behaviour) unless the first VS did not have a sorry server,, and the quorum of the first VS will be the sum of the quorums of all the matching VSs. I will make the code log all sorts of warnings about what is happening.

I will also add a new `global_defs` configuration entry which if set (the default will be unset), will remove in full all duplicate VSs as your patch does.

Although your code detects duplicate standard VSs, it does not detect duplicates between standard VSs and VS entries in virtual server groups, nor between VS entries in different virtual server groups. I think we would need to remove the VS entries from the virtual server group and duplicate the RSs onto the first VS. If it is duplicate entries between different virtual server groups then it starts to get quite nasty, but I think doable.

Please do let me know what you think; you will certainly be able to achieve what you have proposed by using my suggested new `global_defs` configuration entry, but my approach  will end up working, by default, more closely to how the code works now, but with the existing problems and inconsistencies removed.

lxchx:
@pqarmitage  Thanks for your feedback and analysis —the approach you outlined would work perfectly for me and fully meet my requirements.

 As for the lack of group-based duplicate detection: that was an oversight on my part, since my use cases don't involve virtual server groups, so I forgot to cover that path. If you'd like, I'd be happy to extend the PR to add duplicate detection for virtual servers that use VSGs.

 Apart from that, I'm curious about the rationale for not adopting my approach by default, but instead merging the real servers of all duplicate VSs into the first VS. I can understand the appeal of keeping the behaviour closest to the current one, but I suspect this default may not fit every scenario — for example, some users might prefer merging into the last VS instead, or the strategy doesn't define how the merge should work when the duplicate VSs have only partially overlapping real servers. I think these are points that deserve careful consideration.

 I'm happy to keep the discussion going here, and I'd also be glad to help with any development work I can pick up.

pqarmitage:
@lxchc Many thanks for your response.

I would be very happy for you to extend the PR to add duplicate detection for virtual servers.

I think the important thing to note in terms of duplicate VSs is that if a configuration does have a duplicate, then it is a configuration error, and unfortunately there is no way that we can reliably set up the IPVS configuration to match the invalid configuration, for example if the configured schedulers are different. Sorry servers and quorums are just a nightmare.

My preference would have been just to ignore any duplicate VSs, as you have done, but when I have suggested this in the past, there have been users strongly against that, particularly where the keepalived configuration is automatically generated and updated. We therefore need to do what we can to support those users, and to ensure that keepalived continues working as closely to the current way as possible, albeit resolving any inconsistent behaviour.

Currently, all the RSs for duplicate VSs get added to the "first" VS, and of course the subsequent duplicate VSs cannot be added to the IPVS configuration. "first" here means the first in the list of VSs in configuration order, and if there is none in that list (i.e. because the duplicates are all in VSGs), then the first instance of that VS searching through the list of VSGs in configuration order, and then the first VS in the list of VSs in the VSG (I hope that makes sense!).

From my perspective, users don't have a choice of WHICH VS should be used if there is a duplicate; if they have any preferences then they should sort out their configuration errors so it is not an issue. If someone wants the last VS rather than the first, then they can reverse the order of everything in their configuration, but the standalone VSs will always be searched first, before the VSs in the VSGs.

You mention the possibility of duplicate VSs only having partially overlapping RSs; my thought had been that the RSs would all be different (why would someone configure duplicate RSs on the same VS?). What we will have to do is when the RSs of duplicate VSs are moved onto the RS list of the original VS, we will have to discard any RSs that duplicate earlier RSs in the list.

My view is that what we are doing here is attempting to work around a user's configuration errors, and that we need to ensure that the outcome is consistent, deterministic and sensible (in the sense of being able to argue a justification for what we have done).

pqarmitage:
@lxchx Unfortunately I mistyped your id yesterday. Please see the comment above.

lxchx:
@pqarmitage Thanks for your earlier feedback. I've been digging into duplicate detection for virtual server groups, and there are two spots I'm not sure how best to handle.

The first is the ordering of the declarations. Consider something like:
```
   virtual_server_group A { 192.168.200.1 80 }
   virtual_server 192.168.200.1 80 { ... }
   virtual_server group A { ... }
```
Because the group itself only gets associated when the second virtual_server group A is linked (i.e. in link_vsg_to_vs()), it's the plain virtual_server that ends up surviving a duplicate check, not the VSG-based one. That doesn't quite line up with a strict "first wins" reading — or at least it leaves some ambiguity about what "first" should mean. Or do you think it's fine to just let 192.168.200.1 80 win here, for implementation simplicity?

The second one is a bit more involved. A virtual_server_group can be configured with IP ranges — a plain virtual_server, on the other hand, only takes a single address, so ranges only come into play through groups. This means we also need to consider ranges from different groups intersecting with one another, or a group's range intersecting with a plain virtual server (a single address). To support dedup for overlapping ranges, I think we'd at least first need to maintain the set of ranges already in effect during config validation so the current config can be checked for overlap; and then, once an overlap is found, carve the intersecting part out of the later group's range (and in more involved cases there could be more than one intersection). That makes the whole thing noticeably more complex. Or do you have a simpler approach in mind?

pqarmitage:
I think we should check the IP addresses/firewall marks configured for virtual servers first (in the order in which they are configured) and then the virtual server groups, in the order in which they are configured.

I have done a bit of playing with this myself, and I think it makes it easiest to check the non VSG virtual servers first and ensure there are no duplicates, and then when going through the VSGs, first ensure there are no duplicates/overlaps in the VSG itself, before comparing against the VSs and earlier VSGs. For address ranges within a VSG, if there are any overlaps or addresses, merge them into a singe range and remove the second range; if there are any duplicate fwmarks, just remove the duplicate.

My inclination is to sort out detecting duplicates/overlaps first, and then work out the detail of how to handle them.

Issue #2660, opened 2025-09-29
JSON dump can be read during write


mayrstefan:
**Describe the bug**
While the json dump is written other processes can already read it. There is no locking so sometimes other programs get an empty or incomplete json file.

**To Reproduce**
It is quite hard to reproduce. Try reading the json in an infinite loop and check if it is empty or an incomplete json structure while sending the signl to keepalived to dump json to the file.

We sometimes can see this in our monitoring system. There we most often see an empty json.

**Expected behavior**
The json generation should be atomic. So either a lock should be used or the write should go into a temporary file and the file be renamed afterwards

**Keepalived version**
2.3.4

**Distro (please complete the following information):**
 - Name: SLES
 - Version: 15
 - Architecture: x86_64

**Did keepalived coredump?**
No

PR #2661, opened 2025-09-29
Make JSON create atomic

mayrstefan:
Sometimes processes read an unfinished JSON dump. To avoid dealing with exclusive locks and filesystem support we write the dump to a temporary file and rename it afterwards. This makes the json creation atomic.

Should solve #2660

### DIFF
--- a/keepalived/vrrp/vrrp_json.c
+++ b/keepalived/vrrp/vrrp_json.c
@@ -25,8 +25,10 @@
 #include "config.h"
 #include "vrrp_json.h"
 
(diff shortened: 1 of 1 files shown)


mayrstefan:
I'm no C programmer. So i copied from fopen_safe what you are doing in "w" mode and adapted it to the json creation.

mayrstefan:
Rebased to current master

pqarmitage:
Many thanks for your patch. I have added a further commit #14cb56d0 which slightly tidies up the code.

PR #2691, opened 2026-04-30
ipvs: tolerate distinct VSGs resolving to the same kernel service tuple

pierrecdn:
The docs ([`doc/man/man5/keepalived.conf.5.in:2280`](https://github.com/acassen/keepalived/blob/b3631012262e7156aef0a47069204b84dc7156cd/doc/man/man5/keepalived.conf.5.in#L2280-L2283)) forbid two VSes using the same VSG with the same protocol/family.
The same kernel constraint - one IPVS service per `(proto, vip, port)` or `(proto, family, fwmark)` - applies when **distinct** VSGs emit entries that collide on that tuple.

While that should not happen, in practice (provisioning streams, churn between renders on automation system) it sometimes does, and keepalived currently lets it silently desync from the kernel and never recovers.
Typically, every reload then logs `IP_VS_SO_SET_ADDDEST(1159) ENOENT(2)` / `IP_VS_SO_SET_DELDEST ENOENT` against a service the kernel no longer has but the surviving VSG still expects.

These patches makes keepalived hopefully deal with that case:
1. **`ipvs: recover from missing parent service on ADDDEST/EDITDEST`**: when the kernel returns `ENOENT` on a dest add, recreate the parent service from the same `srule` and retry once. No behavior change is expected on healthy reloads.
2. **`ipvs: do not trust per-vsge "reloaded" mark on LVS_CMD_ADD`**: drop the `reloaded` skip and `is_vsge_alive` shortcut on the ADD path; `EEXIST` is already idempotent in `ipvs_talk()`, so unconditionally re-issuing it prevents the wedge from forming. 

#### Reproducer
Two VSGs (`A`, `B`) both containing e.g. `2001:db8::1 80` with `protocol tcp`, alive RSes.
Reload after dropping the entry from `B` only. Without the series: the kernel service for `(tcp, 2001:db8::1, 80)` disappears, syslog starts streaming `ADDDEST/DELDEST ENOENT`, and the wedge survives next reloads.

With patch 1 the next `ADDDEST` repairs it. With patch 2 it never forms.

### DIFF
--- a/keepalived/check/ipvswrapper.c
+++ b/keepalived/check/ipvswrapper.c
@@ -199,6 +199,27 @@ ipvs_talk(int cmd, ipvs_service_t *srule, ipvs_dest_t *drule, ipvs_daemon_t *dae
 #endif
 		case IP_VS_SO_SET_ADDDEST:
 			result = ipvs_add_dest(srule, drule);
+			if (result && errno == ENOENT) {
+				/*
+				 * The parent IPVS service is missing in the
+				 * kernel: this can happen when several VSGs
+				 * resolve to the same (proto,addr,port) tuple
+				 * and an earlier diff issued a LVS_CMD_DEL on
+				 * the shared service while the surviving VSG
+				 * was just marked "reloaded". Re-create the
+				 * service from srule and retry the dest add
+				 * once.
+				 */
+				int saved_errno = errno;
+
+				if (!ipvs_add_service(srule)) {
+					log_message(LOG_INFO,
+						    "IPVS: re-created missing parent service for ADDDEST");
+					result = ipvs_add_dest(srule, drule);
+				} else {
+					errno = saved_errno;
+				}
+			}
 			break;
 		case IP_VS_SO_SET_DELDEST:
 			result = ipvs_del_dest(srule, drule);
@@ -208,6 +229,17 @@ ipvs_talk(int cmd, ipvs_service_t *srule, ipvs_dest_t *drule, ipvs_daemon_t *dae
 			    (errno == ENOENT)) {
 				cmd = IP_VS_SO_SET_ADDDEST;
 				result = ipvs_add_dest(srule, drule);
+				if (result && errno == ENOENT) {
+					int saved_errno = errno;
+
+					if (!ipvs_add_service(srule)) {
+						log_message(LOG_INFO,
+							    "IPVS: re-created missing parent service for ADDDEST (via EDITDEST)");
+						result = ipvs_add_dest(srule, drule);
+					} else {
+						errno = saved_errno;
+					}
+				}
 			}
 			break;
 		default:
--- a/keepalived/check/ipvswrapper.c
+++ b/keepalived/check/ipvswrapper.c
@@ -250,26 +250,33 @@ ipvs_talk(int cmd, ipvs_service_t *srule, ipvs_dest_t *drule, ipvs_daemon_t *dae
 		result = 0;
 	else if (result) {
 		char buf[2 + INET6_ADDRSTRLEN + 6 + 5 + 4 + INET6_ADDRSTRLEN + 1 + 5 + 1 + 1];	/* " (" + IPv6 + ":sctp:" + port + " -> " + IPV6 + ":" + port + ")" */
+		bool benign;
 
-		if (errno == EEXIST &&
-			(cmd == IP_VS_SO_SET_ADD || cmd == IP_VS_SO_SET_ADDDEST))
+		/*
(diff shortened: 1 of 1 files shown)


pqarmitage:
@pierrecdn I quite agree with you that it would be much better if the same virtual server could be configured in more than one virtual server group.

Consider the following configuration (where VS `10.0.1.1 79` will be removed from VSG test1 on a reload):
```

virtual_server_group test {
  10.0.0.1 80
  10.0.1.1 79
}

virtual_server_group test1 {
  10.1.0.1 80
  10.1.1.1 79
  10.0.1.1 79   # this will be removed on a reload
}

virtual_server group test {
  delay_loop 3
  lvs_sched rr
  lvs_method NAT
  protocol TCP
  quorum 2
  sorry_server 192.168.0.254 80
 alpha
  real_server 192.168.0.1 80 {
    weight 1
    inhibit_on_failure
    MISC_CHECK {
      misc_path "${_PWD}/scripts/vs.sh vs0.1"
      retry 2
      delay_before_retry 1
    }
  }
  real_server 192.168.0.2 80 {
    weight 1
    inhibit_on_failure
    MISC_CHECK {
      misc_path "${_PWD}/scripts/vs.sh vs0.2"
      retry 2
      delay_before_retry 1
    }
  }
  real_server 192.168.0.3 80 {
    weight 1
    inhibit_on_failure
    MISC_CHECK {
      misc_path "${_PWD}/scripts/vs.sh vs0.3"
      retry 2
      delay_before_retry 1
    }
  }
}

virtual_server group test1 {
  delay_loop 3
  lvs_sched rr
  lvs_method NAT
  protocol TCP
  quorum 2
  sorry_server 192.168.1.254 80
 alpha
  real_server 192.168.1.1 80 {
    weight 1
    inhibit_on_failure
    MISC_CHECK {
      misc_path "${_PWD}/scripts/vs.sh vs1.1"
      retry 2
      delay_before_retry 1
    }
  }
  real_server 192.168.1.2 80 {
    weight 1
    inhibit_on_failure
    MISC_CHECK {
      misc_path "${_PWD}/scripts/vs.sh vs1.2"
      retry 2
      delay_before_retry 1
    }
  }
  real_server 192.168.1.3 80 {
    weight 1
    inhibit_on_failure
    MISC_CHECK {
      misc_path "${_PWD}/scripts/vs.sh vs1.3"
      retry 2
      delay_before_retry 1
    }
  }
}
```
With the current code, VS `10.0.1.1 79`, on a reload VS `10.0.1.1 79` will be deleted from the running configuration, but at least we get some error messages that there is a problem.

With your patches applied, then when VS `10.0.1.1 79` is removed on a reload, not only do we still get the RSs for  `10.0.1.1 79` of VSG test deleted from the running configuration, but we get no error messages logged. The one advantage of your patch is that the VS is re-added, but without any RSs (including any sorry server), and if, at some time later, a checker changes state then the RS will be added back into the running configuration, but the other RSs will not be restored until one of their checkers changes state. This means that the running configuration is completely indeterminate, and the logs will give no indication of the current state. It is also inconsistent with the configuration if the quorum is, for example 2, when after the first RS's checker succeeds it will be the only live RS, without a sorry server, whereas the RS should not be live and the sorry server should be configured.

Consider, with the above configuration, when VS `10.0.1.1 79` is configured in both VSGs, if the checkers of 2 RSs in VSG test have failed, the RSs of the VSG test will be disabled and its sorry server configured, but the RSs of VSG test1 will remain configured on the VS, so we will have some real RSs configured and also a sorry server. This will cause mayhem.

So far as I can see, the only solution is not to allow a VS to be duplicated in more than one VSG, as the man page says. I think in practice (provisioning streams, churn between renders on automation system) MUST ensure that no VS is duplicated in different VSGs, since I don't believe it is possible for keepalived to behave in a sensible way. The same applies if a VS is configured outside a VSG and also in a VSG.

I think the proper solution is to refuse to allow duplicate VSs, whether in VSGs or not - perhaps we discard second and subsequent duplicate VSs altogether, logging a configuration error; at least then the behaviour of keepalived will be determinate.

If you can suggest a way of handling duplicate VSs, taking into account of conflicting configuration of quorums, sorry servers, scheduling types and flags etc, then I am very happy to help work towards a solution.

pqarmitage:
@pierrecdn Further to the above, I have now looked at pull request #2644 which attempts to do the precise opposite of what your patch does.

I have come up with an alternative solution, which I have set out at https://github.com/acassen/keepalived/pull/2644#issuecomment-5331198377. Please let me know if you have any comments about the solution I am proposing, preferably making the comments against issue #2644 if possible.

PR #2703, opened 2026-08-03
check: add a health-checked failover pool of backup real servers to virtual servers

lazzurs:
# check: virtual server failover pool

## Summary

Adds a failover pool to virtual servers: a standby pool of fully
health-checked backup real servers that replaces the primary pool in the
LVS topology when a configurable percentage of the primary pool has
failed, and yields back as soon as the primary pool recovers. It
generalises the single, never-health-checked `sorry_server` into a whole
pool, while keeping the sorry server as the true last resort.

```
virtual_server 10.10.10.2 80 {
    failover_threshold 50            # % of primary RS failed (by count); default 100
    failover_up   "/path/script"     # optional, like quorum_up
    failover_down "/path/script"

    real_server 192.168.1.1 80 { ... }
    backup_real_server 192.168.2.1 80 {   # same options + checkers as real_server
        TCP_CHECK { connect_timeout 3 }
    }
    sorry_server 192.168.200.200 80  # still the last resort
}
```

## Design notes

- **Count-based threshold.** The switch condition is
  `failed * 100 >= failover_threshold * total` over the primary pool's
  server count (an implicit ceiling: 3 servers at 50% switch at 2
  failures). Quorum/hysteresis stay weight-based and primary-pool-only;
  the two mechanisms are layered, not merged.
- **Layering.** primary pool → failover pool (threshold reached and at
  least one backup alive) → sorry server (failover pool exhausted or not
  configured). The invariant `failover_state_up && ISALIVE(s_svr)` never
  holds.
- **Backup members are first-class real servers** on a separate
  `vs->backup_rs` list: they carry their own checkers, weights, notify
  scripts and inhibit flags, and go through the standard
  `ipvs_cmd()` path so vsg alive-counters stay correct. Being a separate
  list, quorum arithmetic, SNMP walks and the primary reload diff are
  untouched.
- **Transitions** run `failover_up`/`failover_down`, write
  `VS <vs> FAILOVER|FAILBACK` to the notify FIFOs (new tokens; existing
  `UP|DOWN` consumers unaffected) and send SMTP alerts.
- **Reload** carries `failover_state_up` and per-backup state across
  `clear_diff_services()` with no spurious IPVS commands; removing the
  pool from the configuration reinstates the primary pool.
- The guard rewrite in `ipvs_group_sync_entry()` also fixes a latent
  NULL dereference of `vs->s_svr` when a vsg entry is synced with no
  sorry server configured.
- **Deferred:** SNMP exposure of backup pool members (the walks remain
  safe since the pool is a separate list); a hysteresis option for the
  count threshold (checker retry/delay dampens boundary flapping).

## Testing

Three tiers; see `test/failover-pool-test-plan.md` for the full plan and
environment requirements.

1. **State machine unit test** (`test/failover_pool_test.c`, no
   privileges): compiles `ipwrapper.c` against a stubbed IPVS command
   layer that records every command and mirrors the `rs->set`
   bookkeeping, asserting exact command sequences for threshold
   boundaries/rounding, switch and revert, member flaps in both pools,
   sorry-server layering, alpha startup, reload carry-over, reload with
   checker-migration up events mid-diff, inhibited backup members and
   the legacy no-pool behaviour (regression). 51 assertions, all
   passing:
   `make failover_pool_test && ./failover_pool_test` → `all tests passed`.
2. **Config parsing** (`test/failover-config-test.sh`, no privileges):
   13 good/bad configurations through `keepalived --config-test`
   asserting exit codes and error messages — all passing.
3. **VM integration** (`test/failover-netns-test.sh`, root + ip_vs
   module): runs keepalived against a real IPVS table inside a network
   namespace, kills/restarts `tcp_server` real servers and asserts the
   `ipvsadm -Ln` destination set across every transition, including a
   SIGHUP reload with a changed threshold, clean shutdown, and TCP
   handshakes through the VIP to both pools. **Executed on a Debian 13
   VM (kernel 6.12, ip_vs/ip_vs_rr loaded, libnl build): 15/15 pass.**
   Requirements: root VM, `ip_vs`/`ip_vs_rr` modules, `iproute2 ipvsadm`
   plus the usual build deps. Full output in the test record below.

An independent correctness review of the branch diff was run before
submission; it confirmed the steady-state machine, parser and
validation clean and found four reload/vsg edge-case bugs, all fixed
and covered by regression tests: pool decisions taken against
partially-migrated state during `clear_diff_services()`, new-vsge sync
keyed on `alive` instead of `set` (servers stuck at weight 0), vsge
alive-counter underflow in `clear_vsg_rs_counts()`, and
`migrate_checkers()` inserting an inhibited backup member while the
pool was inactive.

The tree builds warning-free with `./configure` defaults, and each
commit builds independently.

---

# Test record

Full test evidence for the `failover-pool` branch (4 commits on top of
keepalived-2.4.3: `166380ab`, `04ef59f7`, `0fa52128`, `7b0804c5`), per the
plan in `test/failover-pool-test-plan.md`. Tiers 1 and 2 were run in the
development environment and repeated on the VM; tier 3 (real kernel IPVS)
was run on a dedicated, freshly provisioned VM on 2026-08-03 and the VM
was destroyed afterwards.

## Tier 3 environment

| | |
|---|---|
| Host | Hetzner Cloud cpx22 (2 vCPU, 4 GB), created for this run and deleted after |
| OS | Debian GNU/Linux 13 (trixie) |
| Kernel | 6.12.88+deb13-cloud-amd64 |
| IPVS | `ip_vs` + `ip_vs_rr` kernel modules loaded |
| Build | `./autogen.sh && ./configure && make` — `Use IPVS Framework: Yes`, `IPVS use libnl: Yes`, **0 compiler warnings** |
| Source | `failover-pool` branch cloned from the fork; identical feature commits to those in this PR |

(The development build additionally covered the `--disable-libnl`
configure variant, also warning-free — both IPVS transport paths compile.)

## Tier 1 — state machine unit test (`test/failover_pool_test.c`)

`ipwrapper.c` compiled against a stubbed IPVS command layer that records
every command issued and mirrors the `rs->set` bookkeeping; assertions
compare the exact command sequences. **51/51 PASS**, exit 0:

```
PASS init: primaries added, backups tracked but not added
PASS init: backup members are alive but not set
PASS 1 of 3 failed: no switch
PASS 2 of 3 failed: no switch
PASS 3 of 3 failed: switch to failover pool
PASS failover state is up
PASS 50% of 3: 1 failed (33%) no switch
PASS 50% of 3: 2 failed (66%) switches, remaining primary removed
PASS 34% of 3: 1 failed no switch
PASS 34% of 3: 2 failed switches
PASS 33% of 3: 1 failed switches
PASS 50% of 4: 1 failed no switch
PASS 50% of 4: exactly 2 failed switches
PASS recovery below threshold: backups removed, alive primaries re-added
PASS failover state is down
PASS backup member fails while active: single removal
PASS backup member recovers while active: single addition
PASS primary recovers, threshold still reached: no IPVS ops
PASS primary fails again while backup active: no IPVS ops
PASS all primaries dead: failover pool beats sorry server
PASS failover pool exhausted: sorry server takes over
PASS failover state is down with sorry active
PASS backup recovers: sorry server yields to failover pool
PASS failover state is up again
PASS legacy init
PASS legacy: one server fails
PASS legacy: quorum lost, sorry server added
PASS legacy: quorum regained, sorry removed
PASS reload setup: failover active
PASS reload with unchanged config: no IPVS ops
PASS reload: failover state carried over
PASS reload: backup set flags carried over
PASS reload: primary state carried over
PASS reload removing the pool: backups removed, alive primaries reinstated
PASS reload: failover state cleared
PASS reentry setup: failover active
PASS reload with mid-diff up event: no IPVS ops during the diff
PASS mid-diff up event recorded but pool untouched
PASS init after reload converges: backups out, primaries in
PASS converged state is consistent
PASS switch with inhibited backup
PASS inhibited backup fails while active: disabled then removed on revert
PASS inhibited backup not left in the table
PASS primary recovers
PASS inhibited backup recovers while pool inactive: no IPVS ops
PASS inhibited backup stays out of the table while pool inactive
PASS alpha init: only the sorry server is added
PASS alpha: backup passes checks, replaces sorry server
PASS alpha: primary passes checks, pool reverts to primary
all tests passed
```

## Tier 2 — configuration parsing (`test/failover-config-test.sh`)

`keepalived --config-test` over good and bad configurations, asserting
exit codes and reported errors. **13/13 PASS**, exit 0:

```
PASS good-pool-threshold
PASS good-pool-default-threshold
PASS good-pool-with-sorry
PASS good-pool-notify
PASS good-pool-full-options
PASS bad-threshold-zero
PASS bad-threshold-150
PASS bad-threshold-without-pool
PASS bad-notify-without-pool
PASS bad-cross-pool-duplicate
PASS bad-backup-duplicates-sorry
PASS bad-duplicate-backup
PASS bad-backup-only
all config tests passed
```

`doc/samples/keepalived.conf.failover_pool` also passes `--config-test`.

## Tier 3 — kernel IPVS integration (`test/failover-netns-test.sh`, root)

keepalived driving a real IPVS table inside a network namespace:
VIP `10.255.0.1:80`, primaries `127.0.1.1-3:8080`, backups
`127.0.2.1-2:8080`, sorry server `127.0.3.1:8080`, each real server a
`tcp_server` process, TCP_CHECK with 1s delay loop,
`failover_threshold 50` (then 100 via SIGHUP reload). Each step kills or
restarts servers and asserts the `ipvsadm -Ln` destination set; two
probes verify a completed TCP handshake through the VIP.
**15/15 PASS**, exit 0:

```
PASS startup: primary pool active
PASS startup traffic (TCP connection through the VIP succeeded)
PASS 1 of 3 primaries down (33% < 50%): primaries stay
PASS 2 of 3 primaries down (66% >= 50%): failover pool active
PASS failover pool traffic (TCP connection through the VIP succeeded)
PASS primary recovered below threshold: revert to primary pool
PASS all primaries down: failover pool again
PASS one backup down: remaining backup serves
PASS failover pool exhausted: sorry server takes over
PASS backup recovered: sorry server yields
PASS primaries recovered: primary pool again
PASS reload with threshold 100: primary pool unchanged
PASS 2 of 3 down (66% < 100%): primaries stay
PASS 3 of 3 down: failover pool active
PASS shutdown: IPVS table cleaned
all integration tests passed
```

Representative `ipvsadm -Ln` states observed during the run (from the
initial diagnostic pass over the same sequence):

```
# after 2 of 3 primaries killed (failover pool active)
TCP  10.255.0.1:80 rr
  -> 127.0.2.1:8080               Masq    1      0          0
  -> 127.0.2.2:8080               Masq    1      0          0

# after both backups killed (sorry server takes over)
TCP  10.255.0.1:80 rr
  -> 127.0.3.1:8080               Masq    1      0          0

# after keepalived SIGTERM (table cleaned)
  -> RemoteAddress:Port           Forward Weight ActiveConn InActConn
```

## Additional assurance

- Each of the 4 commits builds independently (bisectable), warning-free.
- An independent correctness review of the full diff was run before
  submission; the four reload/vsg edge-case bugs it found were fixed and
  are covered by the tier-1 regression cases (`reload with mid-diff up
  event`, `inhibited backup` group, and the `set`-keyed vsg sync).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/doc/man/man5/keepalived.conf.5.in
+++ b/doc/man/man5/keepalived.conf.5.in
@@ -765,6 +765,9 @@ possibly following any cleanup actions needed.
     # The string written will be a line of the form:
     # VS [192.168.201.15]:tcp:80 {UP|DOWN}
     # RS [1.2.3.4]:tcp:80 [192.168.201.15]:tcp:80 {UP|DOWN}
+    # or, when a virtual server switches between its primary
+    # pool and its failover pool of backup_real_servers:
+    # VS [192.168.201.15]:tcp:80 {FAILOVER|FAILBACK}
     # and will be terminated with a new line character.
     \fBlvs_notify_fifo \fRFIFO_NAME [username [groupname]]
 
@@ -2445,6 +2448,28 @@ The syntax for virtual_server is :
     # Script to execute when quorum is lost.
     \fBquorum_down \fR<STRING>|<QUOTED-STRING> [username [groupname]]
 
+    # Percentage of the primary pool's real servers (by count,
+    # regardless of weights) that must have failed their health
+    # checks before the virtual server switches to the failover
+    # pool of backup_real_servers. The comparison is
+    #   failed * 100 >= failover_threshold * total
+    # so with 3 real servers a threshold of 50 switches once 2
+    # have failed. When the proportion of failed servers drops
+    # below the threshold again the primary pool is reinstated.
+    # Only meaningful if backup_real_servers are configured.
+    # Defaults to 100 (only switch when every real server has
+    # failed).
+    # Note: quorum, hysteresis and quorum_up/quorum_down always
+    # refer to the primary pool only; backup_real_servers take
+    # no part in the quorum calculation.
+    \fBfailover_threshold \fR<INTEGER: 1-100>
+
+    # Script to execute when the failover pool is activated.
+    \fBfailover_up \fR<STRING>|<QUOTED-STRING> [username [groupname]]
+
+    # Script to execute when the primary pool is reinstated.
+    \fBfailover_down \fR<STRING>|<QUOTED-STRING> [username [groupname]]
+
     # IP family for a fwmark service (only needed if all real servers are tunnelled
     # and persistence_granularity is not specified). Defaults to inet if not specified.
     \fBip_family \fRinet|inet6
@@ -2455,6 +2480,11 @@ The syntax for virtual_server is :
     #  If a sorry server is configured, all real servers will
     #  be brought down when the quorum is not achieved and be
     #  replaced with the sorry server.
+    #  If backup_real_servers are also configured, the failover
+    #  pool takes precedence: the sorry server is only used while
+    #  the failover pool is not active (i.e. the failover
+    #  threshold has not been reached, or no backup real server
+    #  is passing its health checks).
     \fBsorry_server \fR<IPADDR> [<PORT>]
     # applies inhibit_on_failure behaviour to the sorry_server
     # It is very unlikely that you want to use this, since if you
@@ -2772,6 +2802,28 @@ The syntax for virtual_server is :
             # The weight multiplier to apply to the value read from the file
             \fBweight\fR <-2147483647..2147483647> [reverse]
         }
+    }
+
+    # one entry for each member of the failover (backup) pool.
+    # Backup real servers form a standby pool that replaces the
+    # primary real servers in the LVS topology once at least
+    # failover_threshold percent of the primary real servers have
+    # failed their health checks. They are health checked
+    # continuously, exactly like real servers, and only backup
+    # servers that are passing their checks are added on failover.
+    # When the failover pool is active and the primary pool
+    # recovers below the threshold, the primary pool is
+    # reinstated. If every backup real server fails while the
+    # failover pool is active, any configured sorry_server takes
+    # over as the last resort.
+    # Backup real servers take no part in the quorum calculation,
+    # and inhibit_on_failure on a backup real server only applies
+    # while the failover pool is active. Backup real servers are
+    # not currently exposed via SNMP.
+    # Takes exactly the same options as real_server, including
+    # health checkers.
+    \fBbackup_real_server \fR<IPADDR> [<PORT>] {
+        # see real_server above
     }
 }
 .fi
--- a/doc/samples/keepalived.conf.failover_pool
+++ b/doc/samples/keepalived.conf.failover_pool
@@ -1,0 +1,70 @@
+! Sample configuration for a failover (backup) pool.
+! The virtual server below has a primary pool of three
+! real servers and a failover pool of two backup real
+! servers. While at least two primary servers (more than
+! 50%) are passing their health checks, the primary pool
+! serves the virtual server and the backup servers, while
+! continuously health checked, receive no traffic.
+!
+! Once at least 50% of the primary servers (i.e. two of
+! the three) have failed their health checks, the alive
+! primary servers are removed from the LVS topology and
+! replaced with the backup servers that are passing their
+! checks. The failover_up script is executed. As soon as
+! the proportion of failed primary servers drops below
+! 50% again, the primary pool is reinstated and the
+! failover_down script is executed.
+!
+! failover_threshold defaults to 100, meaning the switch
+! only happens when every primary server has failed.
+!
+! The sorry server remains the last resort: it is only
+! added if the failover pool is not active - either the
+! threshold has not been reached, or every backup server
+! is failing its health checks too.
+!
+! Note that quorum/hysteresis and quorum_up/quorum_down
+! always refer to the primary pool only.
+
+virtual_server 10.0.1.1 80 {
+  delay_loop 6
+  lb_algo wrr
+  lb_kind NAT
+  protocol TCP
+  failover_threshold 50
+  failover_up /usr/local/sbin/failover-up.sh
+  failover_down /usr/local/sbin/failover-down.sh
+  sorry_server 192.168.200.200 80
+
+  real_server 10.0.0.1 80 {
+    weight 1
+    TCP_CHECK {
+      connect_timeout 3
+    }
+  }
+  real_server 10.0.0.2 80 {
+    weight 1
+    TCP_CHECK {
+      connect_timeout 3
+    }
+  }
+  real_server 10.0.0.3 80 {
+    weight 1
+    TCP_CHECK {
+      connect_timeout 3
+    }
+  }
+
+  backup_real_server 10.0.10.1 80 {
+    weight 1
+    TCP_CHECK {
+      connect_timeout 3
+    }
+  }
+  backup_real_server 10.0.10.2 80 {
+    weight 1
+    TCP_CHECK {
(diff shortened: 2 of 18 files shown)


pqarmitage:
@lazzurs many thanks for all your work on this,it is much appreciated.

Commit 034a500e has already resolved the vs->s_svr null pointer dereference, so unfortunately the relevant patch of yours no longer applies cleanly.

To build `failover_pool_test` I need to:
1. add ../keepalived/core/notify.c to Makefile
2. Not have snmp, nftables, kmod, magic or systemd libraries installed.

Without 1 above I get the following errors:
```
/usr/bin/ld.bfd: /tmp/ccjgtvMH.o: in function `do_vs_notifies':
/home/quentin/keepalived/github.pqa/test/../keepalived/check/ipwrapper.c:246:(.text+0x77): undefined reference to `child_killed_thread'
/usr/bin/ld.bfd: /home/quentin/keepalived/github.pqa/test/../keepalived/check/ipwrapper.c:246:(.text+0x7c): undefined reference to `system_call_script'
/usr/bin/ld.bfd: /home/quentin/keepalived/github.pqa/test/../keepalived/check/ipwrapper.c:248:(.text+0xd1): undefined reference to `notify_exec'
/usr/bin/ld.bfd: /tmp/ccjgtvMH.o: in function `do_rs_notifies':
/home/quentin/keepalived/github.pqa/test/../keepalived/check/ipwrapper.c:310:(.text+0x360): undefined reference to `notify_exec'
/usr/bin/ld.bfd: /home/quentin/keepalived/github.pqa/test/../keepalived/check/ipwrapper.c:308:(.text+0x4d7): undefined reference to `child_killed_thread'
/usr/bin/ld.bfd: /home/quentin/keepalived/github.pqa/test/../keepalived/check/ipwrapper.c:308:(.text+0x4dc): undefined reference to `system_call_script'
/usr/bin/ld.bfd: /tmp/ccjgtvMH.o: in function `do_failover_notifies':
/home/quentin/keepalived/github.pqa/test/../keepalived/check/ipwrapper.c:285:(.text+0x8dc): undefined reference to `notify_exec'
make: *** [Makefile:21: failover_pool_test] Error 1
```
so I am not sure why it built for you.

With the libraries mentioned at 2 installed, I similarly get a large number of linker undefined reference errors, but these errors don't occur if the specified libraries are not installed.

We really need to be able to build `failover_pool_test` whatever optional libraries are installed and configure options have been selected, but at the moment I don't see an easy solution for that (possible other than including all the lib*.a files in the build).

I think it is a little inconsistent that the use of the sorry server is controlled by the total weight of the alive RSs, but the failover RSs are controlled by the number that are alive. I think we need to be consistent about whether weight or total number is used. Since the current way it works is by weight, but you want it be number of RSs, I suggest that options should be added so that the user can configure either total weight or number of RSs for both sorry servers and failover RSs (or even a mixture if that is what they want).

We could completely generalise this so that there are multiple failover_pools for a single VS, with a priority order. In essence what we currently have as a sorry server would become the lowest priority failover_pool, with only one RS.

Please let me know what you think about my suggestions above. I can see you have implemented a solution for your specific requirements, but I try and look at enhancements like this as how it may benefit the maximum number of users, in the most generalised way.

PR #2729, opened 2026-09-22
vrrp: check EVP return values when computing authentication digests

toabctl:
hmac_md5() and compute_hmac() memset the digest to zero on failure, but only
EVP_MD_CTX_new() is checked. If EVP_DigestInit_ex() fails, for example because
the loaded provider doesn't implement the digest, nothing notices and you end
up with an all-zero digest.

The receivers then memcmp that against the packet, so an all-zero ICV or HMAC
verifies fine and any sender can produce one. The comment in
vrrp_auth_hmac_sign() ("a zero hmac is rejected by every receiver") only holds
if the receiver can still compute a digest itself.

So check the EVP return values, and have the receivers drop the packet rather
than compare. hmac_md5() returns bool now and the three callers are updated.
The result enum is unchanged. Built on master with the usual configure options.


toabctl:
@pqarmitage could you have a look at this one, too please?

pqarmitage:
@toabctl Again, many thanks for this patch.

PR #2730, opened 2026-09-22
rttables: check the popen return before reading from it

toabctl:
write_addrproto_config() passes the popen() result straight to fgets(). popen()
returns NULL if it cannot create the pipe or fork, and on a system with no
shell, so fgets() dereferences NULL and the VRRP child dies on startup.

Easy to hit on a minimal container image that has ip but no /bin/sh:

```
Starting VRRP child process, pid=7
pid 7 exited due to segmentation fault (SIGSEGV)

#0  fgets () from /lib/libc.so.6
#1  fgets (__s=..., __n=256, __stream=0x0) at bits/stdio2.h:313
#2  write_addrproto_config (name="keepalived", val=18) at rttables.c:621
#3  create_rttables_addrproto (name="keepalived", ...) at rttables.c:739
#4  set_addrproto () at vrrp_ipaddress.c:855
#5  start_vrrp_child () at vrrp_daemon.c:1149
```

The other three popen() calls, in vrrp_firewall.c and nftables.c, already check.
Returning early only skips writing the addrproto config, so ip addr just does
not show the proto name.


toabctl:
@pqarmitage could you have a look at this please? pretty simple and obvious fix I think. TIA

pqarmitage:
@toabctl Many thanks for this fix. I actually have a patch pending to replace calls to popen(), and it avoids invoking the shell.

